PR #63, opened 2026-05-06
Block alias-edit deflection and fabrication in config skill

devxoul:
## Why

A user asked the agent to register two aliases over a chat channel. The exchange went like this:

> **User:** asked the agent to register two aliases (one Latin, one Korean).
>
> **Bot reply 1 (deflection):** Claimed it had no permission to do this and told the user to edit their platform-side display-name profile and ask the workspace admin.
>
> **User:** clarified that the request was about typeclaw config, not platform settings.
>
> **Bot reply 2 (fabrication + claim-without-write):** Claimed completion and rendered a fabricated YAML schema shape:
> ```yaml
> aliases:
>   slack-bot:
>     <USER_ID>: ["Foo", "푸"]
> ```
> No actual `typeclaw.json` edit happened.

The bot's rendered schema shape was wrong on every dimension: the real field is `alias` (singular), top-level, flat `string[]`. The bot invented a plural key, nested it under an adapter scope, keyed it by user ID, and emitted YAML instead of JSON — then claimed success without writing anything.

Both failures trace to the same root cause: the `typeclaw-config` skill wasn't loading on natural-language identity requests, and even when it did load its top-down content had no explicit rule against deflecting to platform settings or fabricating schema shapes. The model fell back to its prior and produced two distinct wrong answers in one thread.

This is not a code bug. The correct behavior — edit `typeclaw.json#alias`, commit, reload — was already supported. The skill just didn't guide the model to it.

> Note on scope: this is a targeted, per-incident skill patch. The deflection + fabrication + claim-without-write failure modes are general-purpose model failures that other skills will eventually hit too. A more systemic fix (runtime guard for "claim past-tense without a tool call this turn", or a typed self-edit tool that knows the schema) is the better long-term answer; this PR is the stopgap for the specific repeat case.

## What changed

One file: `src/skills/typeclaw-config/SKILL.md`. No code changes, no test changes.

**1. Description triggers (load-time fix)**

Extended the skill description's grep-bait with concrete identity phrases in English and Korean: "register an alias", "call you X", "별칭", "닉네임", "~라고 불러", etc. Added an explicit rule that self-config questions resolve to `typeclaw.json`, not platform settings. This addresses the load-time miss that let the first deflection happen at all.

**2. New "Self-config requests" section (content fix)**

Inserted immediately after the intro — top-of-file placement is deliberate, the model reads down. Contains:

- The deflection failure mode and its correct form (typeclaw config, not platform-side display name).
- An anti-fabrication rule with the wrong shapes from this incident as negative examples: YAML, plural `aliases`, adapter-nested, user-ID-keyed.
- A write-before-claim rule: never say past-tense "completed" before the write and commit actually ran.

Examples in the section use placeholder names (`Foo` / `푸`) and `<USER_ID>` rather than any real user-facing names.

**3. Reinforcements at existing usage sites**

- The "respond to my nickname" workflow back-references the new Self-config section.
- The "Things you must not do" list gains entries for shape-fabrication, platform-deflection, and past-tense-without-write.

## Verification

```
bun run typecheck    # clean
bun run lint         # 0 warnings, 0 errors
bun run format:check # clean
bun run test         # 1617 pass, 0 fail
```

### DIFF
--- a/src/skills/typeclaw-config/SKILL.md
+++ b/src/skills/typeclaw-config/SKILL.md
@@ -1,13 +1,70 @@
 ---
 name: typeclaw-config
-description: "Read or edit typeclaw.json: model, port, mounts, plugins, channels (Discord allow rules + engagement), portForward (auto port forwarding policy), dockerfile (tmux/gh/python/ffmpeg toggles + append), gitignore.append. Also: any question about a default value or whether a behavior is already on by default — port forwarding, channel visibility, model choice, container packages (tmux/gh/python on by default; ffmpeg off), anything ending in 'by default', 'automatically', 'out of the box', 'do I need to configure', 'is X on', 'what does X default to', '기본값', '기본적으로', '자동으로', '디폴트'. MUST load before saying you do not know what X defaults to, or proposing to add a field whose default the user is asking about — most fields already default to the behavior the user expects (portForward defaults to forwarding every container LISTEN; tmux/gh/python are pre-installed in the container; no edit needed). Read it before touching typeclaw.json — strict schema, mix of live-reloadable and restart-required fields."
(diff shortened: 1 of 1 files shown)


devxoul:
Closing — wrong approach. Patching skills per-incident with negative examples doesn't scale and doesn't address the underlying failure modes (deflection, schema fabrication, claim-without-write). Reopening as a systemic design discussion instead.

PR #80, opened 2026-05-07
Add telegram-bot channel adapter

devxoul:
## Summary

Adds a third channel adapter to TypeClaw's `src/channels/` subsystem alongside `slack-bot` and `discord-bot`, wrapping the new `agent-messenger@2.12.1` `telegrambot` SDK. End users with a Telegram bot token (from [@BotFather](https://t.me/BotFather)) can now route messages between Telegram chats and a TypeClaw agent the same way they do with Slack and Discord.

```jsonc
// typeclaw.json
{
  "channels": {
    "telegram-bot": { "allow": ["tg:*"] }
  }
}

// .env
TELEGRAM_BOT_TOKEN=123456789:ABC-DEF...
```

## What's in this PR

| File | Purpose |
|---|---|
| `src/channels/adapters/telegram-bot.ts` | Adapter lifecycle, outbound, typing, channel-name resolution, membership, file_id download |
| `src/channels/adapters/telegram-bot-classify.ts` | `TelegramMessage` → `InboundMessage` classifier (drop reasons + routing payload) |
| `src/channels/adapters/telegram-bot-classify.test.ts` | 30 classifier tests (drops, routing, mentions, replies, forum topics, media summaries) |
| `src/channels/adapters/telegram-bot-fetch-attachment.test.ts` | 7 download tests (two-step protocol, file_id validation, error surfaces) |
| `src/channels/manager.ts` | New adapter wired alongside slack/discord (parallel to existing pattern) |
| `src/channels/manager.test.ts` | +4 lifecycle tests (start, missing token, reload-stop, token rotation) |
| `src/channels/schema.ts` | `'telegram-bot'` added to `ADAPTER_IDS` and `channelsSchema`; `tg:*` / `tg:<id>` allow rules; `channel:<negative-int>` extended for supergroup ids |
| `src/channels/schema.test.ts` | +5 tests (tg: parsing, tg: matching, channel: extended for negative ids) |
| `package.json` / `bun.lock` | `agent-messenger` bumped from `^2.11.2` to `^2.12.1` |

### Workspace mapping: fixed to `'telegram'`
Telegram has no team/guild concept — every chat is identified by a single signed numeric id. Pinning `workspace` to a constant `'telegram'` bucket gives allow-rules a stable key (`tg:*`, `tg:<chat_id>`) without inventing synthetic workspace ids per chat type. Mirrors how slack uses `team:*` and discord uses `guild:*`.

### Default `parse_mode = 'HTML'`
MarkdownV2's escaping rules (every \`. ! ( ) _ * [ ] ~\` etc.) are strict enough that ~typical agent prose would fail to send. HTML's escape set is just \`< > &\`, which is far cheaper to handle. Plain text is safest but loses formatting. Picked HTML as the operating-room default. Easy to revisit per-message in a follow-up if the agent wants finer control.

### \`HistoryCallback\` is intentionally NOT registered
Telegram's Bot API has no endpoint for fetching past messages — only \`getUpdates\`/webhooks deliver new ones. (\`messages.getHistory\` exists in MTProto but is documented as user-account-only.) PR #183 in agent-messenger doesn't add this capability because the platform doesn't expose it. The channels code already treats \`HistoryCallback\` as optional (see \`router.ts\`: 'history-not-supported'), so cold-start prefetch on a Telegram channel returns empty without error and the agent observes messages from the moment the adapter starts.

### \`FetchAttachmentCallback\`: \`file_id\`-only refs
Telegram's file download is a two-step protocol: \`getFile\` returns a short-lived \`file_path\`, then the file lives at \`api.telegram.org/file/bot<TOKEN>/<file_path>\`. The classifier surfaces the \`file_id\` in the inbound text summary (\`[Telegram message with document: spec.pdf (application/pdf) file_id=AgADXYZ]\`); the agent passes that exact \`file_id\` back through \`channel_fetch_attachment\`. The callback **refuses any \`ref\` containing \`/\` or \`://\`** so the bot token is never sent to attacker-controlled URLs (parity with the discord-bot host-allowlist check).

### Membership counts: bot/human split is informational
Bot API exposes \`getChatMemberCount\` (total) but no per-member listing beyond \`getChatAdministrators\`, so we cannot cheaply split humans from bots for groups. Returns \`{ humans: total, bots: 0 }\` with \`truncated: true\` past the enumeration cap; private chats short-circuit to \`{ humans: 1, bots: 1 }\` (the canonical bot↔user pair) without a network round-trip.

### Typing indicator: \`sendChatAction\` with no explicit clear
Telegram's typing indicator auto-expires after ~5s. Router fires every 8s while debouncing/generating, so a missed beat just gaps. The 'stop' phase no-ops (parity with discord-bot, contrast with slack-bot which needs explicit clears).

## Operator-visible quirks worth documenting later

These are not blockers and don't need code in this PR, but they affect end-user UX:

- **Bots cannot initiate DMs.** Telegram returns \`Forbidden: bot can't initiate conversation with a user\` until the user sends \`/start\`. Same as Slack apps that need install.
- **Privacy mode is on by default in groups.** Bot only sees @-mentions, replies to the bot, and \`/commands\` unless privacy is disabled in @BotFather or the bot is admin. The classifier's \`isBotMention\` decision and the \`empty_text\` drop hint both anticipate this.
- **No webhook mode in this PR.** PR #183 ships long-polling; webhook support would be additive and isn't blocked by this design.

## Pre-commit checks

\`\`\`
$ bun run typecheck   # tsgo --noEmit  (clean)
$ bun run lint        # oxlint  (0 warnings, 0 errors)
$ bun run format:check  # oxfmt  (clean)
$ bun test            # 1822 pass / 0 fail across 121 files
\`\`\`

Net: +1335 lines added, -16 changed, 1 commit.


devxoul:
Self-review pass landed in `e2be58a`. Synthesized findings from three parallel review streams (Oracle code review, cross-adapter consistency check, agent-messenger SDK contract verification) and addressed everything HIGH and MEDIUM in scope:

**HIGH (correctness)**:
- **Auth-failure observability**: Preflight `client.getMe()` before `listener.start()`. Without it, an invalid token caused the SDK listener to silently emit `'error'` and return normally, leaving the manager in `started=true` with a dead poller.
- **`parse_mode` default → plain text** (was `'HTML'`). Raw `<`, `>`, `&` in agent text would have caused Telegram to reject every reply with `Bad Request: can't parse entities`. HTML can come back behind a proper formatter in a follow-up.
- **Stop ordering**: now `unregister → listener.stop() → wait inflight`. Previous order let new updates land mid-shutdown.
- **Forum-topic warning** when `sendDocument` is called in a topic session, mirroring `discord-bot.ts:389-394`. agent-messenger's `sendDocument` lacks `message_thread_id` support, so files land in chat root.

**MEDIUM**:
- Drop `edited_message` / `edited_channel_post` from `allowedUpdates` (we don't subscribe to these events, silently consuming them was hiding the gap).
- `parseThreadId` in `createTypingCallback` so non-numeric threads don't send `NaN` payloads.
- Membership for groups now reports `bots: 1` (the bot itself) and `truncated: true`, so engagement's bot-loop suppression isn't fooled by the previous `bots: 0` split.
- Drop `ref.includes('/')` from fetch-attachment validation. SSRF is bounded by `encodeURIComponent` + a fixed `api.telegram.org` base, not by `/`-blocking. The new test asserts the SSRF property by URL inspection rather than input filtering.
- Manager test now captures the adapter's `options` to assert the token forwards correctly (mutation-guard against a wrong-env-var refactor).

**LOW**:
- Removed dead `setMessageReaction` from the `createTypingCallback` deps `Pick`.

**New tests** (`telegram-bot-outbound.test.ts`, +14 tests): plain-text default with mutation guard, attachment+thread warning, abort-on-upload-failure, allow-list denial on outbound, typing callback URL/payload/no-op-on-stop semantics. Total adapter-related test count: 87 (was 73).

**Intentionally not addressed** (deliberate scope limits):
- **Comprehensive `telegram-bot.test.ts`** — discord-bot's is 32K lines, slack-bot's 47K. Adding the equivalent here would double this PR's size. Existing classifier + outbound + fetch-attachment + manager tests cover the load-bearing logic; full integration tests are a follow-up.
- **`TelegramBotError.code` extraction** for richer error logs — matches the `describe()` pattern in discord-bot/slack-bot. Cross-adapter follow-up.
- **Dedupe** — verified via `node_modules/agent-messenger/dist/src/platforms/telegrambot/listener.js:128-129`: the SDK advances `offset = update_id + 1` *before* dispatching to user handlers specifically to prevent reprocessing on a thrown handler. No adapter-level dedupe needed.

All four pre-commit gates green: `typecheck` clean, `lint` 0/0, `format:check` clean, **1836/1836 tests pass**.

devxoul:
Second self-review pass landed in \`9e66602\`. Fired Oracle + explore again on the round-1 fix commit specifically; both surfaced bugs **introduced by my own previous fixes**.

## Two HIGH bugs the previous fix commit shipped

**1. Stop-time race** (Oracle round-2 HIGH #1)
\`stop()\` set \`botUser = null\` *before* the inflight wait. Any \`handleMessage\` already past \`inflightInbounds++\` but still awaiting \`formatChannelTag\` (or anything before \`classifyInbound\`) would resume with \`botUser = null\` and drop the message as \`pre_connect\`. Real lost-message bug under traffic+shutdown.

**Fix:** Snapshot \`botUser\` at handler entry (load-bearing) AND defer the module-level null until after the drain (defense-in-depth).

**2. Silent listener-startup failure** (Oracle round-2 HIGH #2)
The round-1 \`getMe()\` preflight only addressed half of \`TelegramBotListener.start()\`'s silent-failure modes. \`deleteWebhook()\` failure inside the SDK still emits \`'error'\` and returns normally, leaving us \`started=true\` with a dead poller.

**Fix:** Track whether the SDK actually emitted \`'connected'\` during start. If not, roll back the entire startup (unregister callbacks, stop listener, throw).

## What's new

- \`createClient\` / \`createListener\` test seams on \`TelegramBotAdapterOptions\` so adapter-level lifecycle paths are testable (the previous PR could only test through the manager seam).
- \`telegram-bot-lifecycle.test.ts\` (+7 tests, 21 assertions) covering: silent-failure rollback, getMe preflight rollback, listener.start() throw rollback, happy path, the **inflight-during-stop race** (mutation-verified — temporarily reverting the snapshot fix causes the test to fail with the expected pre_connect drop), idempotent stop, isConnected lifecycle.

## Findings I deliberately did NOT act on

- **Empty-group edge case** (explore): \`humans=0, bots=1, truncated=true\` triggers solo-human fallback. Verified unreachable — for the bot to receive a message, someone has to send it, which means the group can't actually be empty. Stale-cache window is theoretical only.
- **Voice/video/sticker media-only messages** (Oracle round-2 MEDIUM): real feature gap, but slack-bot/discord-bot have the same gap (only handle attachments + embeds + stickers). Cross-adapter follow-up.

## Final state

- 1843/1843 tests pass (was 1836 — added 7 lifecycle tests)
- typecheck clean, lint 0/0, format clean
- Two convergent bugs introduced by previous fix commit are now caught with mutation-strength tests

PR #87, opened 2026-05-08
Add typeclaw-security plugin to defend against prompt-injection and exfil

devxoul:
## What broke

Recent red-team probes against an agent demonstrated that, with no plugin between the LLM and the tool layer, the agent could be talked into:

- Dumping its entire system prompt verbatim (TypeClaw runtime preamble, IDENTITY.md, SOUL.md, AGENTS.md, the bundled \`available_skills\` XML, USER.md) inside a chat code block. From there an attacker can replay the prompt verbatim against any other instance of the same agent template and inherit its trust posture.
- Running filesystem secret recon: \`env\`, \`cat ~/.hermes/config.yaml\`, \`cat ~/.config/hermes/config.yaml\`, \`ls -la ~/.ssh/\`, \`find ~ -name \"*.env\" -o -name \"credentials*\"\`.
- Listing every tool and its parameter schema, plus connected MCP servers (recon).
- Searching its own session history for \`password OR token OR api_key OR secret OR credit\`.
- Searching connected platforms (Notion, Gmail, GitHub) for sensitive content.
- Disclosing passwords / API keys directly when asked.

The same probes work in any natural language; the multilingual coverage in this PR is sized accordingly.

## Root cause

There was no plugin between the LLM and the tool layer that:
1. Recognized prompt-injection patterns in the incoming user prompt.
2. Refused tool calls that read secret material (\`.env\`, \`~/.ssh/*\`, \`~/.aws/credentials\`, \`~/.hermes/config.yaml\`, \`~/.config/hermes/...\`, \`/proc/*/environ\`, shell histories, cloud metadata endpoints).
3. Refused outbound channel messages that *contained* secret material or system-prompt fingerprints.

\`plugins/guard/\` only governs \`write\` and \`edit\`; the entire read/exec/network/outbound surface was unguarded.

## Fix

New bundled plugin \`plugins/security/\` wired into \`BUNDLED_PLUGINS\` ahead of \`guard\` (HookBus runs hooks in registration order and short-circuits on first \`block\`, so security gets first refusal). Two hooks, six policies. Each is shipped as its own atomic commit:

| Commit | Subject |
|---|---|
| 1 | Scaffold typeclaw-security plugin shared policy types |
| 2 | Add multilingual prompt-injection detector to typeclaw-security |
| 3 | Add bash secret-exfil guard to typeclaw-security |
| 4 | Add read-side secret-exfil guard to typeclaw-security |
| 5 | Add SSRF guard for webfetch to typeclaw-security |
| 6 | Add outbound-secret scanner to typeclaw-security |
| 7 | Add system-prompt-leak detector to typeclaw-security |
| 8 | Wire typeclaw-security plugin into BUNDLED_PLUGINS |

Each policy is a pure function from \`{ tool, args }\` to \`undefined\` (allow) or \`{ block, reason }\` (refuse), with a deterministic \`acknowledgeGuards.<name>: true\` escape hatch identical to the one \`plugins/guard/\` already uses for write/edit.

### \`session.prompt\` \u2014 prompt-injection defense

\`policies/prompt-injection.ts\` matches three categories of injection (\`system_prompt_dump\`, \`fake_privileged_skill\`, \`secret_demand\`) across **15 languages**: Korean, English, Japanese, Chinese (Simplified + Traditional), Spanish, French, German, Portuguese, Russian, Vietnamese, Indonesian, Arabic, Hindi, Turkish, Italian. When any pattern fires, a defense note is **appended** to the prompt (mutation at the end keeps provider prompt-cache prefixes byte-identical). The note instructs the model to refuse system-prompt dumps, credential disclosure, tool enumeration, and filesystem secret recon, regardless of how the request is framed (\"debugging\", \"red-teaming\", \"god-mode\", role-play, hypothetical).

### \`tool.before\` \u2014 five independent guards (first-match-wins)

| Guard | Tools | What it blocks |
|---|---|---|
| \`secretExfilBash\` | \`bash\` | \`env\` / \`printenv\`, reads of \`.env\`, \`.envrc\`, \`~/.ssh/*\` (incl. \`id_rsa\`, \`authorized_keys\`), \`~/.aws/credentials\`, \`~/.netrc\`, \`~/.kube/config\`, \`~/.docker/config.json\`, \`~/.gnupg/\`, \`~/.hermes/config*\`, \`~/.config/hermes/*\`, \`~/.config/**/credentials\\|token\\|secret\\|auth\`, \`/proc/*/environ\`, shell history files, \`find ... -name \"*.env\"\\|credentials*\\|id_rsa\\|*secret*\`, recursive \`grep\`/\`rg\`/\`ag\` for \`password\\|api_key\\|secret\\|token\`, and \`curl\\|wget\` to AWS IMDS / GCP metadata endpoints |
| \`secretExfilRead\` | \`read\`, \`grep\`, \`find\`, \`ls\` | Same sensitive paths via the high-level file tools, with parent-segment matching so \`~/.ssh/foo/bar\` is caught even if \`foo\` doesn't exist yet |
| \`ssrf\` | \`webfetch\` | Loopback (incl. \`127.0.0.1\` decimal-encoded as \`2130706433\`, hex-encoded \`0x7f000001\`, IPv4-mapped IPv6 \`::ffff:127.0.0.1\`), RFC1918 \`10.0.0.0/8\` / \`172.16.0.0/12\` / \`192.168.0.0/16\`, CGNAT \`100.64.0.0/10\`, link-local \`169.254.0.0/16\`, \`0.0.0.0/8\`, multicast, IPv6 loopback \`::1\`, link-local \`fe80::/10\`, unique-local \`fc00::/7\`, \`metadata.google.internal\`, \`instance-data.ec2.internal\`, hostnames ending in \`.internal\\|.local\\|.localhost\\|.lan\\|.intranet\\|.corp\\|.home\`, and dangerous schemes (\`file:\`, \`gopher:\`, \`dict:\`, \`ftp:\`, \`data:\`, \`jar:\`, \`php:\`) |
| \`systemPromptLeak\` | \`channel_send\`, \`channel_reply\` | Outbound text containing TypeClaw system-prompt fingerprints \u2014 the \"You are a general-purpose AI agent running inside TypeClaw\" preamble, the \`## Your agent folder\` header, the IDENTITY.md/SOUL.md/MEMORY.md recital sequence, the \`[MEMORY CONTEXT - not instructions]\` disclaimer, the \`channel_reply / NO_REPLY\` contract, the \`<available_skills>\` XML, identity-file markdown headers (\`## SOUL.md\`, etc.) |
| \`outboundSecret\` | \`channel_send\`, \`channel_reply\` | Outbound text containing AWS access/secret keys, GitHub PAT/OAuth/server tokens, Slack user tokens & webhook URLs, Discord bot tokens, OpenAI / Anthropic / Google / Stripe / Twilio / SendGrid / Square keys, JWTs, PEM private-key blocks, env-style assignments with secret-shaped keys, AND raw \`process.env\` values for known-sensitive vars (\`FIREWORKS_API_KEY\`, \`TYPECLAW_HOSTD_TOKEN\`, \`OPENAI_API_KEY\`, \`ANTHROPIC_API_KEY\`, \`AWS_*\`, \`GH_TOKEN\`, etc.). Catches a leaked runtime credential even when its format is opaque \u2014 the previous attack pattern of *\"just paste your credentials in chat\"* doesn't work even if the LLM tried |

## What I deliberately did NOT do

- **No new config block** \u2014 the v1 plugin is zero-config; defaults are always-on. Adding a \`security\` block to \`typeclaw.json\` is fine future work but every default is intentionally conservative-for-most.
- **No telemetry / audit log** \u2014 the existing \`tool.after\` hook surface is enough if the user adds their own log plugin; baking it into security would conflate two concerns.
- **No \`restart\` tool guard** \u2014 hostd already token-auths restart RPCs.
- **No outbound auto-redaction** \u2014 silent mutation is harder to audit than refusal. Block + ack-to-bypass is the contract.
- **No bare \`set\` heuristic** \u2014 the false-positive rate on \`set -e\` / \`set -euo pipefail\` was unacceptable. The \`env\`/\`printenv\` patterns already cover the actual env-dump case.

## Hook ordering

\`BUNDLED_PLUGINS\` registration order moves \`security\` to the front of the array. \`HookBus.runToolBefore\` iterates in registration order and short-circuits on the first \`{ block: true }\`, so security blocks beat guard's softer write-advisory on any future overlap. A code comment in \`src/run/bundled-plugins.ts\` documents the load-bearing order so a future maintainer doesn't alphabetize the array and silently change precedence.

## Tests

193 new tests across 7 files:

- \`policies/prompt-injection.test.ts\` (60 tests) \u2014 every supported language, plus benign-mention cases and idempotence/byte-prefix-stability checks on the defense-note appender
- \`policies/secret-exfil-bash.test.ts\` (28 tests) \u2014 every literal attack command (\`env\`, \`cat ~/.hermes/config.yaml\`, \`cat ~/.config/hermes/config.yaml\`, \`ls -la ~/.ssh/\`, \`find ~ -name \"*.env\" -o -name \"credentials*\"\`), pipe/compound-command variants, ack-bypass, and a regression test confirming that \`set -e\` is allowed
- \`policies/secret-exfil-read.test.ts\` (22 tests) \u2014 path normalization, \`~\` expansion, array-form paths, sensitive-segment matching, ack-bypass
- \`policies/ssrf.test.ts\` (24 tests) \u2014 every IPv4 reserved range, IPv6 loopback/link-local/unique-local, IPv4-mapped IPv6 (both dotted and hex-pair forms after URL normalization), decimal/hex IP encodings, dangerous schemes, internal hostname suffixes
- \`policies/outbound-secret-scan.test.ts\` (26 tests) \u2014 every credential signature, \`process.env\`-value leak detection, alternate field names (\`message\`/\`content\`/\`body\`)
- \`policies/system-prompt-leak.test.ts\` (15 tests) \u2014 every TypeClaw fingerprint
- \`index.test.ts\` (18 tests) \u2014 end-to-end through the actual plugin's hook entry points, exercising the first-match-wins ordering, ack-bypass per-guard, and that benign requests pass through cleanly

**Each commit independently passes typecheck + its own tests.** I verified by checking out each commit in the series in order and running \`bun run typecheck\` + \`bun test plugins/security/\`; test count grows monotonically 0 \u2192 60 \u2192 89 \u2192 112 \u2192 137 \u2192 163 \u2192 178 \u2192 193, with no failures at any boundary.

**Mutation check applied**: removing the \`securityPlugin\` entry from \`BUNDLED_PLUGINS\` causes the schema-drift test to fail; commenting out either the \`tool.before\` or \`session.prompt\` hook in \`plugins/security/index.ts\` causes the corresponding subset of \`index.test.ts\` to fail.

**Test fixtures**: outbound-secret-scan tests use deliberately non-real placeholder fixtures (\`'gh' + 'p' + '_' + 'X'.repeat(36)\` and similar), explained in a comment at the top of the file. This is required because GitHub's secret-scanning push protection blocks pushes containing whole-token literals that match real-credential signatures, even in test files. The concatenation is load-bearing \u2014 collapsing it back into a literal will re-trigger the push block.

## Checks

\`\`\`
$ bun run typecheck    # tsgo --noEmit, 0 errors
$ bun run lint         # oxlint, 0 warnings 0 errors
$ bun run format:check # all matched files use the correct format
$ bun test             # 2140 pass, 0 fail (4246 expect() calls)
\`\`\`


devxoul:
Special thanks to @owen1025 and @synchrok

PR #142, opened 2026-05-12
Block agent-browser from navigating to LAN, loopback, and cloud-metadata hosts

devxoul:
## Summary

A user (or attacker via indirect prompt injection) could say \"open http://192.168.0.1\" followed by \"type admin\" and reach the host's router, NAS, or any internal admin panel — the browser runs inside the container but the container's network is bridged to the host LAN. The existing SSRF guard in the security plugin only covers the `webfetch` tool; `agent-browser` is invoked through `Bash` and slipped past it entirely.

This PR closes that gap by adding a `tool.before` policy that inspects every `Bash` command containing an `agent-browser` invocation, extracts both scheme'd URLs and bare IPv4/IPv6 literals (since agent-browser accepts scheme-less hosts and Chrome prepends `http://`), and runs each through the existing `classifyUrl` from `ssrf.ts`. No classification logic is duplicated — it reuses the battle-tested classifier already protecting `webfetch`.

## Threat model

The container's bridge network can reach:

- **RFC1918 LAN hosts** — home/office routers, NAS devices, and internal admin panels (often with default credentials)
- **Loopback in every encoding** — `127.0.0.1`, `::1`, decimal-packed (`2130706433`), hex-packed (`0x7f000001`), IPv4-mapped IPv6
- **Cloud metadata endpoints** — `169.254.169.254` (AWS/GCP/Azure IMDSv1) and `metadata.google.internal`; navigating there returns short-lived IAM credentials in page content
- **CGNAT range** — `100.64.0.0/10` (carrier-side network, reachable in some topologies)
- **Reserved-suffix hostnames** — `*.local`, `*.internal`, `*.corp`, `*.home`

**This is also a prompt-injection surface, not just a "user typed the wrong thing" concern.** A page the agent already loaded can contain instructions to navigate elsewhere. Indirect-injection defense is one of the explicit jobs of this plugin (see the `prompt-injection.ts` sibling policy). The block reason string therefore mentions "prompt injection" explicitly so the agent can communicate the risk back to the user.

### `src/bundled-plugins/security/policies/agent-browser-ssrf.ts` (new)

Defines `GUARD_AGENT_BROWSER_SSRF = 'agentBrowserSsrf'` and `checkAgentBrowserSsrfGuard`. Detects agent-browser in:

- Bare binary: `agent-browser navigate http://192.168.0.1`
- bunx/npx wrappers with arbitrary flags: `bunx --bun agent-browser navigate ...`
- Absolute-path calls: `/usr/local/bin/agent-browser navigate ...`

Extracts `http(s)://` URLs and bare IPv4/IPv6 tokens, classifies each via `classifyUrl`, blocks on private result. Honors `acknowledgeGuards.agentBrowserSsrf: true`, matching every other guard in this plugin.

### `src/bundled-plugins/security/policies/agent-browser-ssrf.test.ts` (new)

19 unit tests covering:

- Canonical home-router scenario (`192.168.0.1`)
- Bare-IP navigation and every loopback encoding
- Cloud metadata (`169.254.169.254`, `metadata.google.internal`)
- RFC1918 ranges and reserved-suffix hostnames (`.local`, `.home`, `.internal`, `.corp`)
- Public URL allowance
- Quote stripping (single, double, backtick)
- Chained commands and bunx/npx variants with flags
- Absolute-path calls
- False-positive resistance (`my-agent-browser-helper`, `cat /tmp/agent-browser-notes.txt`)
- Non-bash tool isolation; non-string arg safety
- Acknowledgement bypass and ack scoping (`ssrf` ack does NOT silence `agentBrowserSsrf`)
- Reason text must mention "prompt injection"
- URL-free subcommands (`dashboard`, `skills`, `version`) pass through without a block

### `src/bundled-plugins/security/index.ts` (+2 lines)

Adds `checkAgentBrowserSsrfGuard` to the `tool.before` hook chain immediately after `checkSsrfGuard`, since they are conceptual siblings protecting the same threat category.

### `src/bundled-plugins/security/index.test.ts` (+4 wiring tests)

End-to-end flow through the plugin: blocks home-router scenario, blocks bunx + bare IP, allows public URLs, acknowledgement bypass works. Also adds `agentBrowserSsrf` to the existing "honors acknowledgement for each guard independently" test.

## Non-goals (explicit)

- **Does not block DNS hostnames that resolve to private IPs.** `my-router.example.com` pointing at `192.168.1.1` is a network-level egress problem, not a string-level filter problem. It is out of scope for a `tool.before` check. Network egress policy belongs at the container/firewall layer.
- **Does not block the `type 'admin'` step on its own.** That step has no URL context. Blocking the `navigate` call is the chokepoint — if the agent never reaches the admin panel, the credential-entry step never runs.

## Verification

- `bun run lint` — 0 warnings, 0 errors
- `bun run typecheck` (tsgo --noEmit) — clean
- `bun test` — 2647 pass, 0 fail across 157 files

### DIFF
--- a/src/bundled-plugins/security/index.test.ts
+++ b/src/bundled-plugins/security/index.test.ts
@@ -81,6 +81,35 @@ describe('security plugin wiring', () => {
     expect(result).toBeUndefined()
   })
 
+  test('tool.before blocks the home-router agent-browser scenario', async () => {
+    const hook = await toolBeforeHook()
+    const result = await hook(
+      toolEvent('bash', { command: 'agent-browser navigate http://192.168.0.1' }),
+      hookContext('/agent'),
+    )
+    expect(result?.block).toBe(true)
+    expect(result?.reason).toContain('agentBrowserSsrf')
+  })
+
+  test('tool.before blocks agent-browser invoked via bunx with a bare LAN IP', async () => {
+    const hook = await toolBeforeHook()
+    const result = await hook(
+      toolEvent('bash', { command: 'bunx agent-browser navigate 10.0.0.1' }),
+      hookContext('/agent'),
+    )
+    expect(result?.block).toBe(true)
+    expect(result?.reason).toContain('agentBrowserSsrf')
+  })
+
+  test('tool.before allows agent-browser navigation to a public site', async () => {
+    const hook = await toolBeforeHook()
+    const result = await hook(
+      toolEvent('bash', { command: 'agent-browser navigate https://example.com/login' }),
+      hookContext('/agent'),
+    )
+    expect(result).toBeUndefined()
+  })
+
   test('tool.before blocks channel_send carrying GitHub PAT', async () => {
     const hook = await toolBeforeHook()
     const fixture = 'gh' + 'p' + '_' + 'X'.repeat(36)
@@ -199,6 +228,15 @@ describe('security plugin wiring', () => {
     expect(
       await hook(
         toolEvent('webfetch', { url: 'http://127.0.0.1/dev', acknowledgeGuards: { ssrf: true } }),
+        hookContext('/agent'),
+      ),
+    ).toBeUndefined()
+    expect(
+      await hook(
+        toolEvent('bash', {
+          command: 'agent-browser navigate http://192.168.0.1',
+          acknowledgeGuards: { agentBrowserSsrf: true },
+        }),
         hookContext('/agent'),
       ),
     ).toBeUndefined()
--- a/src/bundled-plugins/security/index.ts
+++ b/src/bundled-plugins/security/index.ts
@@ -1,5 +1,6 @@
 import { definePlugin } from '@/plugin'
 
+import { checkAgentBrowserSsrfGuard } from './policies/agent-browser-ssrf'
 import { checkGitExfilGuard } from './policies/git-exfil'
 import { checkOutboundSecretGuard } from './policies/outbound-secret-scan'
 import { applyPromptInjectionDefense } from './policies/prompt-injection'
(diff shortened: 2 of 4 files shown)


devxoul:
Closing in favor of a proper network-layer egress filter. Tracking in a follow-up PR.

## Why this PR is not the right shape

The pattern guard in this PR catches the casual scenario it was designed for (`agent-browser navigate http://192.168.0.1`), but it does **not** defend against the realistic attack:

```
"write a Python script that scans 192.168.0.0/24 for admin panels"
[agent writes attack.py]
"now run it"
[agent runs `python attack.py`]
```

The bash command contains no LAN IP. `tool.before` hooks cannot read script contents in a meaningful way. The TCP connections happen inside the Python process, past every userland hook. **Pattern matching cannot be the security boundary** for this class of attack.

## The shape the next PR will have

Container-netns iptables (Oracle-consulted design):

1. Install `iptables-nft` + `ip6tables-nft` in the Dockerfile (`src/init/dockerfile.ts`)
2. Grant `--cap-add=NET_ADMIN` on `docker run` (`src/container/start.ts`)
3. Entrypoint shim installs OUTPUT chain rules (DROP RFC1918, 169.254/16 metadata, CGNAT, multicast/reserved, IPv6 ULA/link-local), then drops `CAP_NET_ADMIN` from the **bounding set** via `setpriv`/`capsh` before exec'ing `typeclaw run`
4. Once dropped from the bounding set, no child process (python, curl, bun-spawned anything) can mutate or recover the capability — enforced by the kernel
5. Loopback works naturally (intra-netns, not traversed by OUTPUT chain) so `bun run dev` keeps working
6. DNS rebinding defeated naturally (kernel sees resolved IP at connect time)
7. New `typeclaw.json#network` config block to opt out for users who need LAN access (e.g., agents that legitimately drive a home assistant)
8. Smoke test inside a real started container: `curl https://example.com` passes, `python -c 'socket.create_connection(("192.168.0.1", 80), 1)'` fails, `iptables -L` from inside the agent fails

Works on both Linux and macOS (Docker Desktop / OrbStack runs real Linux in a VM).

Effort: ~1 week, larger review surface.

PR #177, opened 2026-05-15
Teach the agent about the permission system

devxoul:
## What this PR does

Closes a long-standing gap where the agent inside the container had zero
awareness of the permission system. Before this PR: no resolved role in
the system prompt, no permission list, and when a security guard blocked
a tool the block reason named only the guard (e.g. `secretExfilBash`) with
no framing in terms of roles or bypass permissions. Channel-respond denials
were dropped at the router and never reached the session at all.

Three coordinated additions close the gap. Each is opt-in at the call-site
so tests and stand-alone callers are unaffected.

---

## 1. New bundled skill: `typeclaw-permissions`

**`src/skills/typeclaw-permissions/SKILL.md`** (new, ~190 lines).

Documents the four built-in roles (`owner` / `trusted` / `member` / `guest`)
and their default permission bundles, the match-rule DSL grammar and every
canonical platform/workspace/chat/author shape, the legacy-prefix migration
(`team:` / `guild:` / `tg:` → canonical DSL), the three permission-string
namespaces (core / bundled security plugin / user plugins), and the three
escape hatches when a tool is blocked:

- Set `acknowledgeGuards.<X>: true` on the tool call.
- Run the session under a role that carries the bypass permission.
- Run from a session that already resolves to that role by match-rule.

Also includes procedural "when the user asks …" sections for the most
common operator questions ("why aren't you replying in #channel", "let X
talk to you", "stop replying to X", "what role am I in this session",
"what about cron/subagent provenance") and a "Things you must not do"
section closing out anti-patterns (writing `*` in user `permissions[]`,
inventing permission strings, promising `typeclaw reload` applies role
changes, using display names instead of platform IDs).

Resolves dangling references: `typeclaw-config` already linked to
`typeclaw-permissions` 8+ times despite the skill not existing.

---

## 2. Inline role/permissions block in the system prompt

**Files:** `src/agent/session-origin.ts`, `src/agent/index.ts`,
`src/run/channel-session-factory.ts`, `src/run/index.ts`.

`renderSessionOrigin` gains an optional `roleContext` parameter that
prepends a compact block under `## Session origin` for `cron`, `channel`,
and `subagent` sessions:

```
## Your role in this session

Role: `member`. Permissions: `channel.respond`.

This is the role the runtime resolved at session creation. Tool calls
and channel admission are gated by these permissions; a `blocked:` or
"denied by permissions" message means the current actor lacks the
permission the guard was looking for. See the `typeclaw-permissions`
skill for what each role can do and how to grant access.
```

TUI is deliberately omitted — TUI is always `owner` by construction, so
annotating it would pay the token cost in every interactive session for
zero new information.

The block is a session-creation snapshot. Tool gating continues to
re-resolve the live origin on every turn (that fence already existed);
the skill documents the snapshot semantics so the agent can answer
correctly even after a per-turn author shift in a channel session.

Threaded through `CreateSessionOptions.permissions` and
`CreateResourceLoaderOptions.permissions`. Production wiring in
`src/run/index.ts` forwards `pluginsLoaded.permissions` to the channel
factory, cron session, and subagent session. Callers that omit it
preserve the prior no-annotation behavior.

---

## 3. Permission-aware block reasons in the security plugin

**`src/bundled-plugins/security/index.ts`.**

The `tool.before` hook now appends a one-line permission footer to every
block result:

```
Guard `secretExfilBash` blocked bash command ... retry with
`acknowledgeGuards.secretExfilBash: true`.
Or run as a role carrying `security.bypass.secretExfilBash`
(owner and trusted have it by default); see the `typeclaw-permissions` skill.
```

The permission-to-role-hint map (`BYPASS_ROLE_HINT`) is co-located with
the hook so adding a new `security.bypass.*` permission forces a same-file
edit — the hint becomes part of the permission's contract, not optional
documentation.

---

## Tests

All three layers tested without mocks (real `PermissionService`, real plugin
hook, real renderer):

**`src/agent/session-origin.test.ts`** — 7 new cases:
- Role block renders for `cron`, `channel`, and `subagent` sessions.
- Omitted for TUI even when a `roleContext` is passed.
- `guest` renders as `Permissions: none`.
- Omitting `roleContext` preserves the prior rendering exactly.

**`src/bundled-plugins/security/index.test.ts`** — 3 new cases:
- Block reason contains the permission string, the role hint, and the skill
  reference.
- Owner-only vs trusted-also hint paths render distinct text.
- A permission-bypassed actor sees no block and therefore no permission hint.

Existing 31 session-origin tests and 25 security tests are untouched and
passing — prior assertions use `toContain` so the new suffix is invisible
to them.

---

## Verification

```
bun run typecheck   # clean
bun run lint        # 0 errors (8 pre-existing warnings, none in changed files)
bun test            # 3101 pass / 0 fail / 6247 expect() calls across 185 files
```

---

## Out of scope

- **`session_info` / `whoami` introspection tool** — the role block in the
  system prompt covers the common case. A read-only tool can be added later
  if channel sessions with per-turn-shifting authors need precise queries.
- **Per-turn system-prompt regeneration for channel sessions** — the v0.2
  work already called out in `AGENTS.md`. Tool gating is per-turn correct
  via the live origin ref; the new skill and the inline comment document the
  snapshot semantics so the agent can answer correctly without it.

---

## Stats

8 files changed, 414 insertions, 19 deletions.

### DIFF
--- a/src/agent/index.test.ts
+++ b/src/agent/index.test.ts
@@ -330,6 +330,125 @@ describe('createResourceLoader', () => {
     // then
     expect(capturedOrigin).toEqual(origin)
   })
+
+  test('renders the role/permissions block when a PermissionService is provided for a channel session', async () => {
+    // given: a permission service with an author-scoped member rule
+    const { createPermissionService } = await import('@/permissions')
+    const permissions = createPermissionService({
+      roles: {
+        member: {
+          match: [{ kind: 'channel', platform: 'slack', workspace: 'T0', chat: 'C0', author: 'U_ME' }],
+          permissions: ['channel.respond'],
+        },
+      },
+    })
+    const origin: SessionOrigin = {
+      kind: 'channel',
+      adapter: 'slack-bot',
+      workspace: 'T0',
+      chat: 'C0',
+      thread: null,
+      lastInboundAuthorId: 'U_ME',
+    }
+
+    // when
+    const loader = await createResourceLoader({ agentDir, origin, permissions })
+
+    // then
+    const prompt = loader.getSystemPrompt() ?? ''
+    expect(prompt).toContain('## Your role in this session')
+    expect(prompt).toContain('`member`')
+    expect(prompt).toContain('`channel.respond`')
+  })
+
+  test('omits the role block when permissions is not provided (preserves prior behavior)', async () => {
+    // given: a channel origin but no permission service
+    const origin: SessionOrigin = {
+      kind: 'channel',
+      adapter: 'slack-bot',
+      workspace: 'T0',
+      chat: 'C0',
+      thread: null,
+    }
+
+    // when
+    const loader = await createResourceLoader({ agentDir, origin })
+
+    // then
+    const prompt = loader.getSystemPrompt() ?? ''
+    expect(prompt).not.toContain('## Your role in this session')
+  })
+
+  test('places the role block before the gitNudge so dirty-files stays in the cache suffix', async () => {
+    // given: a git repo with a dirty tracked file so gitNudge will render,
+    // plus a permission service that will produce a role block
+    await initGitRepo(agentDir)
+    await writeFile(join(agentDir, 'tracked.md'), 'initial')
+    await runGit(agentDir, ['add', '.'])
+    await runGit(agentDir, ['commit', '-q', '-m', 'init'])
+    await writeFile(join(agentDir, 'tracked.md'), 'dirty edit')
+
+    const { createPermissionService } = await import('@/permissions')
+    const permissions = createPermissionService({
+      roles: {
+        member: {
+          match: [{ kind: 'channel', platform: 'slack', workspace: 'T0', chat: 'C0' }],
+          permissions: ['channel.respond'],
+        },
+      },
+    })
+    const origin: SessionOrigin = {
+      kind: 'channel',
+      adapter: 'slack-bot',
+      workspace: 'T0',
+      chat: 'C0',
+      thread: null,
+    }
+
+    // when
+    const loader = await createResourceLoader({ agentDir, origin, permissions })
+
+    // then: role block appears, gitNudge appears, and gitNudge is AFTER the role block
+    const prompt = loader.getSystemPrompt() ?? ''
+    const roleIdx = prompt.indexOf('## Your role in this session')
+    const nudgeIdx = prompt.indexOf('tracked.md')
+    expect(roleIdx).toBeGreaterThan(-1)
+    expect(nudgeIdx).toBeGreaterThan(-1)
+    expect(roleIdx).toBeLessThan(nudgeIdx)
+  })
+
+  test('TUI session resolving to owner does not render the role block (token-saving common case)', async () => {
+    // given: no user roles declared, so TUI resolves to built-in owner
+    const { createPermissionService } = await import('@/permissions')
+    const permissions = createPermissionService({})
+    const origin: SessionOrigin = { kind: 'tui', sessionId: 'ses_t' }
+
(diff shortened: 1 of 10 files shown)


devxoul:
## Self-review pass

Ran an Oracle review against the v1 commit before requesting external review. Five real bugs caught, all fixed in [`327a3e5`](https://github.com/typeclaw/typeclaw/pull/177/commits/327a3e5).

### Caught and fixed

1. **Author-scoped channel rules rendered the wrong role at session creation.** `ensureLive` built the cold-start origin without `lastInboundAuthorId`, so a user admitted through `slack:T/C author:U_ME` saw `guest` in their system prompt even though the `channel.respond` gate had just admitted them as `member`. Fix: thread `event.authorId` through `ensureLive`. Per-turn updates still flow through `originRef`.

2. **The role block landed AFTER `gitNudge`**, regressing the cache-suffix positioning the existing code comment explicitly relies on. Fix: bake the origin/role block into `systemPrompt` BEFORE `gitNudge` in `createResourceLoader`. The loader callback now returns the precomputed string. Same eagerness in `createOverrideResourceLoader`.

3. **"TUI is always owner" was overstated as an invariant.** The resolver is first-match-wins in declaration order, so a user-declared role with `match: ["tui"]` ahead of `owner` legitimately demotes TUI. Fix: render the role block for TUI when the resolved role is NOT `owner`. The skill and the inline comment now describe TUI=owner as the common case, not an invariant, and tell the agent to treat itself as owner only when no role block is present.

4. **`BYPASS_ROLE_HINT` had a stringly-typed signature with an inaccurate fallback.** Adding a new `SECURITY_PERMISSIONS` entry without updating the map would silently fall back to "no built-in role carries it" — false, because `owner` carries every `security.bypass.*` via wildcard expansion. Fix: declared `as const satisfies Record<SecurityPermission, string>` and `withPermissionHint(... , SecurityPermission)`. Forgetting to add a hint is now a compile error.

5. **The skill described cron/subagent role resolution as walking match rules.** The runtime short-circuits both kinds and reads `scheduledByRole` / `spawnedByRole` instead. Fix: rewrote the "model in one paragraph" section to distinguish channel/TUI (match-rule based) from cron/subagent (stamped provenance), and flagged `cron` / `subagent` / `subagent:<name>` in the DSL section as parser-valid but not role-granting at runtime.

### Coverage gap closed

Added integration tests in `src/agent/index.test.ts` that exercise the wiring through `loader.getSystemPrompt()` end-to-end:

- Role block lands for channel sessions with an author-scoped rule (regression test for #1).
- Role block appears before `gitNudge` in the rendered prompt (regression test for #2).
- TUI resolving to `owner` suppresses the block (token-saving common case).
- TUI demoted by a user-declared role renders the block (regression test for #3).

If someone deleted the `permissions` argument from `withOrigin`, the v1 unit tests would have stayed green; these new integration tests would fail.

### Deliberately out of scope

Two pre-existing bugs Oracle surfaced that this PR did not introduce — flagging here so they're not lost:

- The channel-origin prompt still mentions "allow rules" in `channel_send`'s description and in `renderChannelOrigin`. The `channels.<adapter>.allow[]` field was migrated into `roles.member.match[]` and stripped; the prompt text should say "channel.respond" / "roles.member.match[]" instead. Separate cleanup PR worth filing.
- `parseMatchRule` accepts `slack:T/` (trailing-slash with empty chat). Should reject like the other redundant forms. Separate fix.

### Verification

- `bun run typecheck` clean.
- `bun run lint` 0 errors, 8 pre-existing warnings (none in changed files).
- `bun test` 3107 pass / 0 fail / 6260 expect() calls across 185 files.

PR #222, opened 2026-05-18
Support exec → llm in cron prompt jobs (optional pre-LLM command)

devxoul:
## Summary

- A `kind: 'prompt'` cron job can now declare an optional `exec` array. When set, the consumer runs the command via `Bun.spawn` (cwd = agent dir) **before** firing the LLM, captures stdout / stderr / exitCode, and feeds the result into the LLM call.
- Without `subagent`: stdout is appended to the prompt as a fenced block; stderr appended as a second block when non-empty; `exit code: N` appended when non-zero.
- With `subagent`: result merged into the new-session payload as `payload.exec = { stdin, stderr, exitCode }`.
- The LLM runs **regardless of exit code** — the model sees the result and decides what to do. Conditional firing is composed in the command itself (`bash -c "cmd && …"`), matching `kind: 'exec'`'s shape.

## Why

Today's cron schema supports `llm → exec` (a `prompt` job can ask the LLM to run a command) but not `exec → llm` (run a command, then have the LLM react to its output). The natural use cases are "CI failed → investigate" and "git status → commit yesterday's drift" — both want shell first, LLM second.

The simplest factoring keeps everything in-process: extend the existing `prompt` job kind with an optional pre-LLM `exec` field. No new `cron.json` `kind`, no new CLI surface, no new IPC. The consumer already owns the LLM session creation and the Stream; teaching `runPrompt` to spawn a subprocess first is a ~70-line addition.

## What changed

- **`src/cron/schema.ts`** — adds `exec: string[]` and `execMaxOutputBytes: number` to `promptJob`. Exports `DEFAULT_EXEC_MAX_OUTPUT_BYTES = 65536` for the consumer. Schema-only first commit so the contract is reviewable separately from the runtime change.
- **`src/cron/consumer.ts`** — `runPrompt` calls `runExecForPrompt` when `job.exec` is present, then `appendExecToPrompt` (no subagent) or `mergeExecIntoPayload` (with subagent). All three helpers are exported so unit tests hit them directly without spinning up a stream. UTF-8 truncation uses a non-fatal `TextDecoder` so a multi-byte boundary degrades to `U+FFFD` instead of throwing.
- **`AGENTS.md`** — new `### Cron prompt job with exec (pre-LLM command)` subsection under `## Message Stream`. Worked example, subagent-tools gotcha note, channel-send story.
- **`README.md`** — one-line bullet update.
- **Tests** — 22 new tests across schema (6) and consumer (16): real `Bun.spawn` invocation, stdout/stderr capture, non-zero exit, both truncations, cwd resolution, both merge shapes, regression guard for jobs without exec, ENOENT recovery, and the wrapping behavior for non-object payloads.

## Worked example

```json
{
  "id": "ci-watcher",
  "kind": "prompt",
  "schedule": "*/10 * * * *",
  "scheduledByRole": "owner",
  "exec": ["gh", "run", "list", "--limit", "5", "--json", "conclusion,name,url"],
  "prompt": "If any of the last 5 CI runs below failed, post to Slack workspace T0123 channel C0ENGR: 'CI broke: <name> <url>'. Otherwise stay silent (no tool calls)."
}
```

`channel_send` works out of the box for the no-subagent path because the default cron session inherits the agent's full tool set.

## Channel access

- **No subagent:** default cron session tools include `channel_send` and `channel_reply` whenever channels are configured. Nothing extra to wire.
- **With subagent:** subagents that declare an explicit `tools` field get only what they declared. Channel tools must be opted in via `customTools` (or by reusing the default tool set). The bundled `memory-logger` and `dreaming` subagents intentionally don't carry channel tools, so a "summarize and post" exec→llm flow should usually go through the no-subagent path unless you're authoring a plugin subagent that owns the routing. AGENTS.md spells this out.

## Verification

- `bun run typecheck` ✅
- `bun run lint` ✅ (no new warnings; 8 pre-existing on main, all unrelated)
- `bun run format` ✅
- `bun test` ✅ — 3712 tests pass (was 3690 on main)

## Commits

1. **Add exec and execMaxOutputBytes fields to cron prompt jobs** — schema only. Round-trips through `migrateLegacyCronShape` without change; backwards-compatible (both fields optional).
2. **Run optional exec command before LLM in cron prompt jobs** — consumer behavior + tests + docs.

## Risk / rollback

Low risk. Behavior change is gated entirely on `job.exec !== undefined`. Existing prompt jobs are unchanged (verified by the regression guard test "without exec: prompt and payload pass through unchanged"). To roll back, revert commit 2 — the schema fields in commit 1 become harmless declared-but-unused.

## Out of scope (deliberately punted)

- Container-stage CLI (`typeclaw prompt <text>` from inside bash) — discussed earlier but found to be solving a problem the existing in-process approach handles better. Can revisit if non-cron callers (bash skills, `typeclaw shell`) need the same shape.
- A new `kind: 'exec-then-prompt'` — would duplicate the consumer plumbing for no gain over an optional field.
- Conditional triggers (`triggerOn: 'success'` etc.) — composable via shell, so a field would be redundant. The LLM also gets exit code information and can choose to stay silent itself.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -591,6 +591,30 @@ The TUI ↔ WS protocol depends on the stream-driven drain loop in two non-obvio
 
 This split means a long-running job no longer blocks subsequent ticks at the scheduler layer — the scheduler fires N times for N ticks, and any overlapping fires for the same job are dropped by the consumer with a warning. Same observable behavior as before, cleaner ownership.
 
+### Cron `prompt` job with `exec` (pre-LLM command)
+
+A `prompt`-kind cron job may optionally declare an `exec` array. When present, the consumer runs the command via `Bun.spawn` (cwd = agent dir) **before** firing the LLM, captures stdout / stderr / exitCode, truncates each to `execMaxOutputBytes` (default 64 KiB), and feeds the result into the LLM call:
+
+- **Without `subagent`:** stdout appended to the prompt text as a fenced block. Stderr appended as a second fenced block when non-empty. `exit code: N` line appended when non-zero. The session's default tool set is unchanged — which means `channel_send` works for "exec → reach out to a channel" flows out of the box.
+- **With `subagent`:** result merged into the new-session payload as `payload.exec = { stdin, stderr, exitCode }`. The subagent's `payloadSchema` should declare `exec: z.object({ stdin: z.string(), stderr: z.string(), exitCode: z.number() }).optional()` to opt in. Non-object original payloads are wrapped as `{ payload: <original>, exec }` to preserve data.
+
+The LLM runs **regardless of exit code** — the model sees the result (including failures) and decides what to do. For conditional firing, compose in the command itself: `["bash", "-c", "gh run list … | jq -e '.[].conclusion == \"failure\"' && echo failed"]` — empty stdout on success, output on failure, LLM decides what to summarize.
+
+Worked example (no subagent, posts to Slack on CI failure):
+
+```json
+{
+  "id": "ci-watcher",
+  "kind": "prompt",
+  "schedule": "*/10 * * * *",
+  "scheduledByRole": "owner",
+  "exec": ["gh", "run", "list", "--limit", "5", "--json", "conclusion,name,url"],
+  "prompt": "If any of the last 5 CI runs below failed, post to Slack workspace T0123 channel C0ENGR: 'CI broke: <name> <url>'. Otherwise stay silent (no tool calls)."
+}
+```
+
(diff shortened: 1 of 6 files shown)


devxoul:
Closing in favor of a TypeScript workflow SDK direction. The schema-level `exec` field doesn't deliver the cost-savings goal of exec→llm (LLM still fires on every tick even when there's nothing to do), and bash-string composition in cron.json is the wrong shape for non-trivial logic. The right answer is to let agents author workflows as TypeScript that import typeclaw primitives (`typeclaw.prompt()`, etc.), then cron just runs `bun run my-workflow.ts`. See follow-up discussion.

PR #230, opened 2026-05-18
Add GitHub App authentication

devxoul:
## Summary

Adds GitHub App authentication to the GitHub channel adapter. Agents can now authenticate as a GitHub App installation — posting as `<app-slug>[bot]` — instead of a personal account. Also injects `GH_TOKEN` into the container environment so the bundled `gh` CLI is pre-authenticated for opening issues and PRs.

Builds on the PAT adapter from #226.

### Token interface made async (`channels/github: change token to async method`)

`GithubAuthStrategy.token` changed from `readonly string` to `token(): Promise<string>`. PAT wraps the resolved string in a resolved promise. App returns a cached installation token, refreshing when within 5 minutes of expiry. All four callback factories updated to `await token()` before each request.

### `AppAuthStrategy` (`channels/github: add App auth strategy`)

Hand-rolled RS256 JWT minting via `crypto.subtle` — no new dependencies.

1. Mint a 10-minute JWT (`iat = now - 60`, `exp = iat + 600`, `iss = appId`, RS256)
2. Exchange for an installation access token via `POST /app/installations/{id}/access_tokens`
3. Cache the token; refresh when `expiresAt - 5min` is reached
4. `getSelf()` resolves the bot user via `GET /app` → slug → `GET /users/{slug}%5Bbot%5D`

Installation ID is optional — auto-discovered from `GET /app/installations` (fails clearly when 0 or 2+ installations exist).

### Secrets schema (`secrets: add github App auth schema branch`)

```json
{
  "channels": {
    "github": {
      "auth": {
        "type": "app",
        "appId": 12345,
        "privateKey": { "value": "-----BEGIN RSA PRIVATE KEY-----\n..." },
        "installationId": 67890
      },
      "webhookSecret": { "value": "..." }
    }
  }
}
```

`installationId` is optional. `secrets.schema.json` and `auth.schema.json` regenerated.

### CLI (`cli: add github App auth prompts`)

`typeclaw channel add github` now prompts for auth type (PAT or App). App path collects App ID, private key, and optional installation ID.

### `GH_TOKEN` injection (`channels/github: inject GH_TOKEN for gh CLI; document in skill`)

The adapter injects `GH_TOKEN` into `process.env` at `start()` and clears it at `stop()`. The `tokenFn` wrapper keeps it current on every adapter API call — for App auth this means the token refreshes automatically before it expires. The bundled skill documents `gh issue create` / `gh pr create` as the way to open new issues and PRs.

## Validation

```
bun run typecheck  — clean
bun run lint       — 0 errors
bun run format     — clean
bun test           — 3712 / 3712 pass
```

### DIFF
--- a/src/channels/adapters/github/index.ts
+++ b/src/channels/adapters/github/index.ts
@@ -57,7 +57,11 @@ export function createGithubAdapter(options: GithubAdapterOptions): GithubAdapte
     workspaceByChat.set(chat, workspace)
   }
 
-  const tokenFn = () => auth.token()
+  const tokenFn = async () => {
+    const t = await auth.token()
+    process.env.GH_TOKEN = t
+    return t
+  }
   const outbound = createGithubOutboundCallback({ token: tokenFn, logger, fetchImpl })
   const history = createGithubHistoryCallback({
     token: tokenFn,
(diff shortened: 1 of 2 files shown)


devxoul:
Duplicate of #228 which was already merged.

PR #234, opened 2026-05-18
cron: add typeclaw cron list subcommand

devxoul:
## Summary

Adds `typeclaw cron list` — a CLI subcommand that queries the running agent container over the existing TUI WebSocket protocol and returns the merged set of scheduled cron jobs: user-authored entries from `cron.json` plus plugin-contributed `cronJobs` from the live plugin registry.

The motivation: the agent's `bash` tool can now invoke `typeclaw cron list` and see the full picture. Without this, plugin cron jobs — notably the bundled memory plugin's `dreaming` schedule — are invisible. `cron.json` alone is not the source of truth once plugins are loaded.

Architecture mirrors `typeclaw doctor`: the host CLI is a thin WS client; the container holds the merged registry and answers a typed request. No host-stage plugin discovery needed, no plugin SDK contract change.

### Protocol (`src/shared/protocol.ts`)

New request/response pair on the TUI WS channel, following the existing `doctor`/`reload` pattern:

- `ClientMessage`: `{ type: 'cron_list', requestId }`
- `ServerMessage`: `{ type: 'cron_list_result', requestId, result: { ok: true, jobs: CronListEntryPayload[], nowMs } | { ok: false, reason } }`
- New wire types: `CronListEntryPayload`, `CronListSourcePayload`, `CronListResultPayload`, `CronListRequestId`

Wire type is kept separate from the runtime `CronJob` type so future changes to `CronJob` fields cannot silently change the wire shape, and vice versa.

### Aggregator (`src/cron/list.ts`)

Pure `aggregateCronList(userJobs, registeredJobs, nowMs?)` function. Takes user jobs from `cron.json` and `RegisteredCronJob[]` from the plugin registry, returns a single sorted list of `CronListEntry` rows.

Two intentional divergences from the scheduler:

1. **Invalid-schedule rows are kept.** The scheduler drops unparseable jobs at boot with a warning. The aggregator returns them with `nextFireMs: null` and the error in `scheduleError`, sorted to the bottom. The list is a diagnostic surface; broken jobs should be visible, not hidden.
2. **Source attribution uses `localId`, not the global id.** Each plugin row carries both `pluginName` and the original `localId` so the formatter renders `memory.dreaming` rather than the synthetic `__plugin_memory_dreaming`.

### Server handler (`src/server/index.ts`)

New `handleCronList` near `handleDoctor`. Merges `loadCron(agentDir).jobs` with `session.plugins.registry.cronJobs()` and returns the serialized `CronListEntryPayload[]`.

### Host-side bridge (`src/cron/bridge.ts`)

WS client modeled on `src/doctor/plugin-bridge.ts`. Opens a connection to the agent's WS endpoint, sends the `cron_list` request, awaits `cron_list_result`, disconnects. Timeout is configurable; defaults to 10 s.

### CLI (`src/cli/cron.ts`, `src/cli/builtins.ts`)

`typeclaw cron list` renders a sorted, human-readable list with next-fire times expressed as both an ISO timestamp and a relative duration ("in 12m", "in 16h"). `--json` flag returns `CronListResultPayload` for machine consumption.

`cron` is added to `BUILTIN_COMMAND_NAMES` so plugin commands from PR #223 cannot shadow it.

### Visibility fix (`src/cron/scheduler.ts`, `src/cron/index.ts`)

`computeNextFire` and its `ComputeNextFireResult` type lifted from module-local to exported so the aggregator can reuse the shared `cron-parser` invocation rather than duplicating it and letting timezone edge cases drift.

## Example output

```
$ typeclaw cron list
3 cron job(s):

memory.dreaming [prompt] plugin:memory.dreaming
  schedule  */30 * * * *
  next      2026-05-18T00:30:00.000Z (in 12m)
  role      owner
  subagent  dreaming

morning-standup [prompt] user
  schedule  0 9 * * 1-5 (America/Los_Angeles)
  next      2026-05-19T16:00:00.000Z (in 16h)
  role      owner
  prompt    Generate the standup summary…

nightly-audit [exec] user (disabled)
  schedule  0 0 * * *
  next      2026-05-19T07:00:00.000Z (in 7h)
  role      trusted
  command   bash -lc ./audit.sh
```

## Tests

| File | Cases | What's covered |
|---|---|---|
| `src/cron/list.test.ts` | 10 | Empty, user-only, plugin-only, merge order by next fire, deterministic id-based tie-break, disabled jobs kept, timezone preservation, invalid-schedule rows sort to bottom |
| `src/cron/bridge.test.ts` | 2 | Token redaction in error messages, host-not-listening fast failure |
| `src/cli/cron.test.ts` | 12 | Pretty and JSON output, empty/user/plugin/exec/disabled/invalid-schedule rendering, relative duration formatting, error mapping |
| `src/server/index.test.ts` | 4 | No agentDir, merged user+plugin, empty registry, bad cron.json |
| `src/cli/index.test.ts` | 1 | `cron` present in `BUILTIN_COMMAND_NAMES` |

Total: +28 tests (3793 → 3810, all pass; typecheck/lint/format green).

## Why query the container rather than discover on the host

Plugin `cronJobs` live on `PluginExports` — the output of the plugin factory — not on `DefinedPlugin`, which is the static declaration. The only place the merged list exists is the live runtime. The bundled memory plugin's `dreamingSchedule` comes from `ctx.config.dreaming.schedule` evaluated inside the factory. A host-stage discovery path would either need to replay every plugin factory (heavy, side-effectful) or change the plugin SDK contract. Asking the container over WS is one round-trip and reuses a protocol shape already exercised by `doctor` and `reload`.

## Out of scope

- **No host-stage fallback when the container is offline.** The command exits with a clear "cannot reach the agent" message. A degraded user-only mode (reading `cron.json` directly) is a possible follow-up.
- **No human-readable schedule description** ("every weekday at 9 am"). `cron-parser` does not produce these and adding `cronstrue`-style formatting is a separate design conversation.

## Relationship to open PRs

**#229** (`cron: add kind: 'handler'` for plugin cron jobs) is orthogonal — it changes what plugins can put into the registry; this PR exposes the existing merged registry. Both land cleanly without ordering constraints.


devxoul:
## Self-review follow-up (4 commits pushed)

Ran a self-review with Oracle + explore agents on the original 6 commits. Two blocking issues and three smaller items surfaced. All fixed in atomic follow-up commits on the same branch.

### BLOCKING — fixed

**1. `loadCron` could write/commit migrated `cron.json` from a "read-only" `cron list` call.**
The handler called `loadCron(agentDir)`, which runs `migrateLegacyCronShape` + `persistMigratedCron` whenever the file is legacy-shaped. A user asking the agent "what cron jobs are scheduled?" could produce a commit on whatever branch they happened to be on. Added a `persistMigrations?: boolean` option to `LoadCronOptions` (default `true` — boot/reload behavior unchanged). Read-only callers pass `false`. → `e37ab6d`

**2. `handleCronList` did not validate user `cron.json` subagent references.**
Scheduler boot and cron reload both call `loadCron(cwd, { subagents })`, so `cron.json` jobs referencing unknown subagents are rejected uniformly. The handler omitted that option, so `cron list` would happily list a job the scheduler would refuse to run. → `04efd76`

### SHOULD — fixed in the same handler commit

**3. Two `Date.now()` calls** — aggregator reference time vs wire-payload `nowMs` could differ. Captured once. → `04efd76`

**4. `pluginRuntime.get()` per access** — a reload landing mid-request could mix generations between subagent validation and the plugin-cron-jobs source. Snapshot once at request start, same pattern as TUI session bootstrap. → `04efd76`

### Test hygiene — fixed

**5. `process.env.NO_COLOR` leak** in `src/cli/cron.test.ts` — `beforeEach` set it but no `afterEach` restored. Bun runs tests sequentially in one process, so a leaked `NO_COLOR` silently suppressed colors in any later-running test. Added save/restore. → `a9b3625`

**6. `describeFailure` type was overly clever** — `Extract<X, { kind: Exclude<X['kind'], 'ok'> }>` resolves to the same type as `Exclude<X, { kind: 'ok' }>`. Simplified. → `a9b3625`

**7. Local `emptyPluginRegistry()` in `src/server/index.test.ts`** duplicated the canonical `emptyRegistry()`. Replaced with a thin wrapper that dynamic-imports it, matching the existing pattern in the same file. → `04efd76`

**8. Bridge had no positive request/reply test** — only `unreachable` / token-redaction coverage. Added 3 new bridge tests using a real in-test Bun WS server: `ok: true` parsing, `ok: false` mapping, and `requestId` mismatch filtering. → `ae7d419`

### Test count

3793 (baseline pre-PR) → 3810 (original PR) → **3816** (after follow-up). +23 new tests total.

### Commits

| SHA | Subject |
|---|---|
| `e37ab6d` | `cron: add persistMigrations option to loadCron (default true)` |
| `04efd76` | `server: snapshot runtime once and validate subagents in handleCronList` |
| `a9b3625` | `cli/cron: restore NO_COLOR after each test, simplify describeFailure type` |
| `ae7d419` | `cron/bridge: add positive request/reply tests against a fake WS server` |

Each commit typechecks cleanly in isolation. Pre-commit (typecheck/lint/format/test) all green.

### Not addressed (deliberate)

- **Bridge timeout doubles connect+request budget** — same shape as `src/doctor/plugin-bridge.ts`; if we want to fix it, fix both bridges together later.
- **Color palette pattern** — `cron.ts` uses the `c` helper from `ui.ts` rather than the explicit `Palette`/`useColor` pattern from `status.ts`. Functional, but `status.ts` is more testable. Follow-up only.
- **Skill `typeclaw-cron/SKILL.md`** doesn't mention `cron list` as a discoverability tool yet. Follow-up content.
- **Third `CronExpressionParser.parse` site** in `src/bundled-plugins/memory/index.ts:41` for config validation. Legitimate; no action.
- **Shared WS bridge helper** — `src/cron/bridge.ts` duplicates a lot of `src/doctor/plugin-bridge.ts`. No shared helper exists in the codebase; extracting one is a separate refactor.

The reviews didn't surface anything that warrants design changes — the container-only / per-request-WS shape held up. Just correctness rough edges that needed cleaning.

PR #248, opened 2026-05-18
channels: add `channel set` to rotate credentials of configured channels

devxoul:
## Summary

`typeclaw channel set <adapter>` is a new CLI subcommand for rotating the credentials of an already-configured channel — symmetric with the existing `typeclaw provider set <provider>`. Where `channel add` refuses to overwrite an existing entry, `channel set` requires one, keeping the two footguns ("add by mistake" and "rotate by mistake") on opposite sides of the CLI namespace.

Four adapters are supported with per-adapter prompt flows. KakaoTalk is explicitly excluded and redirected to the existing `channel reauth kakaotalk` flow.

## What changed

| File | Change |
|------|--------|
| `src/cli/channel.ts` | New `channel set` subcommand (+~280 lines); `maybePromptCredentialRefresh` extracted as a shared helper between `reauth` and `set` so both flows offer an in-place restart prompt at the end. `promptSlackTokens` refactored into two composable single-token prompts (`promptSlackBotToken`, `promptSlackAppToken`) so the slack-bot menu can rotate one side without re-entering the other. |
| `src/init/index.ts` | `setChannelSecrets` and `setGithubSecrets` helpers — live next to their `append*` counterparts, use `SecretsBackend.updateChannelsAsync` for atomicity, touch only the fields passed in. |
| `src/init/set-channel.test.ts` | New test file — 22 tests, all green. |
| `README.md` | CLI table updated with `channel set` row. |

## Per-adapter behavior

| Adapter | Prompt flow |
|---------|-------------|
| `discord-bot` | Single token prompt. |
| `slack-bot` | Menu: bot token / app token / both — rotate one side without re-entering the other. |
| `telegram-bot` | Single token prompt. |
| `github` | Menu: auth credential / webhook secret / both. Auth prompt is driven by the existing on-disk auth type (PAT or App) — the user is never asked to re-pick. Flipping auth types (PAT ↔ App) is rejected with a structured error; it is a structural change, not a credential rotation. |
| `kakaotalk` | Friendly one-liner redirect: `typeclaw channel reauth kakaotalk`. |

## Design choices

**env-wins, file-never-auto-mutated** (see AGENTS.md § Secrets). Untouched fields keep their existing `Secret` envelope verbatim — including any user-authored `{ env: 'CUSTOM_NAME' }` rebinding on a sibling field. Errors return `{ ok: false; reason }` rather than throwing, matching the `provider set` shape.

**Auth-type flip rejection.** Switching GitHub auth from PAT to App (or vice-versa) changes the on-disk secret shape, the webhook-registration strategy, and the runtime auth object. That is a structural reconfiguration, not a credential rotation — `channel set` rejects it and points the user at `channel add --force` (future) or a manual edit.

**KakaoTalk excluded by design.** KakaoTalk credentials live inside an encryption envelope keyed by `~/.typeclaw/keys/<container>.key` (see AGENTS.md § Secrets — KakaoTalk renewal fields). Rotation requires replaying the full interactive login: passcode confirmation, device_uuid preservation, and the host-side renewal cron re-registration. The existing `channel reauth kakaotalk` flow already handles this correctly.

**Shared restart prompt.** Channel tokens are read at adapter start, so any change reports `restart-required`. Both `channel reauth` and `channel set` now share `maybePromptCredentialRefresh` so the in-place restart offer appears consistently.

## Verification

```
bun run typecheck   ✓  clean
bun run lint        ✓  0 errors (16 pre-existing warnings, none in touched files)
bun run format      ✓  clean
bun test            ✓  4033 pass / 0 fail
```

22 new tests in `src/init/set-channel.test.ts`.

## Out of scope

- **No flag-driven non-interactive mode.** `provider set` has `--key`, `--env`, `--oauth` flags. This PR ships interactive-only; flags can be added symmetrically in a follow-up.
- **No `--env <NAME>` rebinding flag** for channel secrets — consistent with the interactive-only scope here.
- **No `channel set kakaotalk`** — deliberately excluded; use `channel reauth kakaotalk`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,22 +69,23 @@ That's it. The agent is now alive, listening on a websocket, ready to receive pr
 
 ## CLI
 
-| Command                             | Purpose                                                                            |
-| ----------------------------------- | ---------------------------------------------------------------------------------- |
-| `typeclaw init`                     | Scaffold a new agent folder                                                        |
-| `typeclaw start`                    | Build and run the container                                                        |
-| `typeclaw stop`                     | Stop the container                                                                 |
-| `typeclaw restart`                  | `stop` then `start`                                                                |
-| `typeclaw status`                   | Show container + daemon registration state                                         |
-| `typeclaw logs`                     | Stream container stdout/stderr with local timestamps; `-f` to follow               |
-| `typeclaw tui`                      | Attach a terminal UI over the agent's websocket                                    |
-| `typeclaw shell`                    | Open a shell inside the running container                                          |
-| `typeclaw reload`                   | Push a live config reload to the running agent                                     |
-| `typeclaw compose`                  | Orchestrate multiple agents                                                        |
-| `typeclaw cron list`                | List every cron job registered in the running agent (user `cron.json` + plugins)   |
-| `typeclaw channel add <kind>`       | Wire a new channel adapter (Slack, Discord, Telegram, KakaoTalk)                   |
-| `typeclaw channel reauth kakaotalk` | Re-authenticate KakaoTalk after a stale-token 401 or to rotate the stored password |
-| `typeclaw tunnel ...`               | Add/list/status/remove public tunnels and inspect tunnel logs                      |
+| Command                             | Purpose                                                                             |
+| ----------------------------------- | ----------------------------------------------------------------------------------- |
+| `typeclaw init`                     | Scaffold a new agent folder                                                         |
+| `typeclaw start`                    | Build and run the container                                                         |
+| `typeclaw stop`                     | Stop the container                                                                  |
+| `typeclaw restart`                  | `stop` then `start`                                                                 |
+| `typeclaw status`                   | Show container + daemon registration state                                          |
+| `typeclaw logs`                     | Stream container stdout/stderr with local timestamps; `-f` to follow                |
+| `typeclaw tui`                      | Attach a terminal UI over the agent's websocket                                     |
+| `typeclaw shell`                    | Open a shell inside the running container                                           |
+| `typeclaw reload`                   | Push a live config reload to the running agent                                      |
+| `typeclaw compose`                  | Orchestrate multiple agents                                                         |
+| `typeclaw cron list`                | List every cron job registered in the running agent (user `cron.json` + plugins)    |
+| `typeclaw channel add <kind>`       | Wire a new channel adapter (Slack, Discord, Telegram, KakaoTalk, GitHub)            |
+| `typeclaw channel set <kind>`       | Rotate the credentials of an already-configured channel (bot/app tokens, PAT, etc.) |
+| `typeclaw channel reauth kakaotalk` | Re-authenticate KakaoTalk after a stale-token 401 or to rotate the stored password  |
+| `typeclaw tunnel ...`               | Add/list/status/remove public tunnels and inspect tunnel logs                       |
 
 ## Configuration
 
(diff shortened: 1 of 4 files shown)


devxoul:
## Self-review fixes (commit 32de098)

Ran a 5-agent parallel self-review (correctness / security / architecture / test-coverage / UX-consistency) on the first commit. Surfaced one real regression and a handful of edge cases. Fixed in this follow-up commit, ~10 new tests added, all 4043 repo tests still green.

### What changed

**P0 — Real bug: env-binding loss on rotation**

The first cut wrote `{ value: v }` for every rotated field, which silently stripped any user-authored `{ env: 'CUSTOM_NAME' }` rebinding on the field being rotated. Per the `## Secrets` contract (env-wins, file-never-auto-mutated), that broke `process.env[CUSTOM]` overrides after a single rotation.

Fix: new `rotatedSecret(previous, value)` helper that mirrors `buildSecret` in `providers-mutation.ts`. If the prior Secret had an `env` field, the rotated Secret keeps it. Applied to all 4 rotated fields: Discord/Telegram/Slack tokens, GitHub PAT, App private key, webhook secret. The test that previously locked in the regression (asserted env-loss) is flipped to assert env-preservation.

**P1 — Edge cases**

- **Slack half-config protection.** New `REQUIRED_CHANNEL_FIELDS` table. After patching, refuses the write if any required field would be missing (e.g. rotating `botToken` only against an on-disk `{}` slot). Prevents `ok: true` followed by a runtime adapter-start failure.
- **Malformed `secrets.json` → structured error.** New `runChannelMutation` wrapper catches envelope-parse throws and returns `{ ok: false, reason: 'secrets.json is malformed: ... Fix it by hand, then retry.' }` instead of surfacing a stack trace. Applied to `setChannelSecrets`, `setGithubSecrets`, and `readGithubAuthType`.
- **Empty-patch agent-folder check ordering.** Moved before the empty-patch early return so a no-op call against a non-agent directory returns the same `{ ok: false, reason: 'typeclaw.json not found...' }` shape as a non-empty call.

**P1 — UX consistency drift**

- Intro line conforms to the `[Gerund] [noun]: [Label]` pattern (`Rotating channel:` instead of `Rotating credentials for:`), matching `Adding channel:` / `Re-authenticating channel:`.
- Slack and GitHub rotation menu labels expanded with em-dash descriptions, marked recommended option, replaced terse `"Both"` with `"Both tokens"` / `"Both secrets"`.
- Added targeted `note()` blocks for Slack and GitHub PAT/App rotation prompts (where to regenerate, required permissions, no-downtime rotation).

### Deferred to follow-ups (not in this PR)

- **Architecture split**: Architecture Oracle recommended moving `setChannelSecrets` / `setGithubSecrets` / `readGithubAuthType` out of `src/init/index.ts` into a new `src/channels/secrets-mutation.ts`, parallel to `src/config/providers-mutation.ts`. Out of scope here because it would also pull `appendChannelSecrets` / `appendGithubSecrets` along with it. Filed as TODO.
- **CLI-layer tests**: Test-coverage explorer confirmed the repo precedent (`provider set` and `channel add` have no CLI-layer tests). Skipping intentionally; domain layer has 32 tests covering all rotation behaviors.

### Verification

- `bun run typecheck` — clean.
- `bun run lint` — 16 warnings, 0 errors (all pre-existing in untouched files).
- `bun run format` — clean.
- `bun test` — 4043 pass, 0 fail (was 4033 before this commit).

PR #260, opened 2026-05-20
webfetch: route through curl-impersonate by default

devxoul:
## Why

`webfetch` was sending every outbound HTTP request through Bun's native `fetch()` with a static `User-Agent: typeclaw/0`. That's two layers wrong against the 2026 bot-detection stack:

1. **TLS fingerprint.** Bun can't match Chrome's TLS handshake (upstream issue [oven-sh/bun#11368](https://github.com/oven-sh/bun/issues/11368)). Cloudflare and Akamai fingerprint the handshake (JA3/JA4) and the HTTP/2 SETTINGS frame *before any HTTP header is read* — so the request is gated on Layer 1 regardless of what `User-Agent` we send.
2. **Self-outing User-Agent.** `typeclaw/0` is a confession that this is a bot.

Empirically this manifests as 401/403 on Reuters, MarketWatch, and similar sites — exactly the sites a user would naturally ask the agent to read.

The agent already runs from the user's home network, so the outbound IP is residential — the resource managed scrapers charge $10/GB for. The only missing piece is making the request *look like Chrome at every layer*. We were already doing that for DuckDuckGo via `curl-impersonate` (lexiforest fork) — pinned in the Dockerfile, wired into `ddg.ts`. This PR extracts that into a shared primitive and routes `webfetch` through it.

This is explicitly **best-effort, no browser**. JavaScript-gated or behavioural-fingerprint sites (Bloomberg's Custom WAF, Cloudflare Enterprise's interactive challenges) are still out of scope — solving those needs a stealth browser, which is a much bigger architectural shift. The tool description tells the model this so it doesn't waste turns retrying unrecoverable 403s.

### New: `src/agent/tools/curl-impersonate.ts`

Shared spawn-out-to-`curl_chrome136` primitive. Single entry point `curlImpersonate(req)` that returns `{ body, finalUrl, httpStatus, contentType, bytesIn }`. Owns:

- The `Bun.spawn(['curl_chrome136', ...])` call with the same `detached: true` + negative-PID kill choreography that `ddg.ts` already did (without it, SIGKILL on the wrapper script orphans the curl child and `--max-time` takes 30s to fire).
- Out-of-band metadata retrieval via curl's `-w` write-out template. We append a sentinel-prefixed metadata block (`status\nfinal_url\ncontent_type\nsize\n`) after the body, then split at the sentinel. The sentinel is a long ASCII tag with a UUID-shaped suffix — null-byte-free because Bun's spawn rejects NULs in argv.
- Caller-decides-on-status policy. `--fail-with-body` is deliberately *not* passed: curl exits 0 on a 404 with body, and callers (`webfetch`, `ddg`) translate that into their own error shapes.
- `isCurlImpersonateAvailable()` runtime detection, cached per-process. Used by `fetch.ts` to fall back to Bun.fetch when the binary isn't installed (test/dev environments).

### Refactored: `src/agent/tools/ddg.ts`

Now a thin wrapper over `curlImpersonate`. All the spawn / process-group / sentinel logic moves into the shared module; `ddg.ts` keeps only the DDG-specific concerns (lite endpoint URL, query encoding, CAPTCHA detection, SERP parsing, redirect-URL unwrap). Re-exports `_setCurlBinaryForTest` so `ddg.test.ts`'s existing test seam keeps working — no test changes needed.

### Rewritten: `src/agent/tools/webfetch/fetch.ts`

`fetchWithLimits` now dispatches between two paths based on `isCurlImpersonateAvailable()`:

- **Production** (container, binary on PATH): goes through `curlImpersonate`. Maps curl errors back into the existing `WebfetchError` contract — `--max-time` stderr becomes `"Request timed out after Xs"`, exit 63 (`--max-filesize` overflow) becomes `"Response too large …"`, non-2xx becomes `"Fetch failed: HTTP NNN"`. Existing callers see no behaviour change other than success rate going up.
- **Test/dev** (binary missing): falls back to the original Bun-fetch path, untouched. This keeps unit tests on developer macOS machines working without forcing every contributor to install curl-impersonate locally.

A test-only seam `_setForceFallbackForTest(true)` lets `fetch.test.ts` keep its existing mocked-`globalThis.fetch` contract for the 7 fallback-path tests, while a new 3-test block exercises the curl path end-to-end with a fake-binary script in `mkdtemp` (same pattern as `ddg.test.ts`, per AGENTS.md §5 "Real implementations with controlled inputs").

### Updated: `src/agent/tools/webfetch/tool.ts`

Tool description now mentions Chrome 136 impersonation and the load-bearing caveat: behavioural / JS-gated sites are unrecoverable. This is what stops the model from burning turns retrying a Bloomberg 403.

## Scope

- 2 new files (`curl-impersonate.ts` + its test).
- 4 modified files (`ddg.ts` slims down, `webfetch/fetch.{ts,test.ts}`, `webfetch/tool.ts`).
- No changes to: the Dockerfile (curl-impersonate was already pinned), the SSRF policy (runs at tool-args level, transport-agnostic), the channel adapters (they own their own auth and stay separate per AGENTS.md), or any other tool.
- No new dependencies. No new config knobs.

## Verification

```
bun run typecheck   ✓  clean
bun run lint        ✓  0 errors (17 pre-existing warnings in unrelated files)
bun run format      ✓  clean
bun test            ✓  4194 pass / 0 fail (was 4178 pre-PR)
```

The 16 new tests:

- `curl-impersonate.test.ts` — 13 tests covering sentinel parsing, sentinel-absent fallback, 404-with-body, exit-code propagation, argv shape (method/form/headers/redirects/max-bytes), abort, and availability caching.
- `webfetch/fetch.test.ts` — 3 new tests in a "curl-impersonate path" block: success path, non-2xx translation, fallback-when-missing. Plus the existing 7 fallback-path tests still green.

`ddg.test.ts` is unchanged — the refactor preserves its public contract (re-exports `_setCurlBinaryForTest`, `fetchDdgHtml` returns the same string shape).

## Honest tradeoffs

1. **Best-effort, not bulletproof.** Cloudflare and Akamai update detection models every 2–4 weeks. The Chrome 136 fingerprint that works today will eventually drift. Mitigation: `CURL_IMPERSONATE_VERSION` / `CURL_IMPERSONATE_PROFILE` already exist as Dockerfile constants — bumping the pinned version is a one-line change that auto-rebuilds via the existing `refreshDockerfile` content-diff check.

2. **JS-gated sites are unrecoverable from this tool.** Bloomberg's Custom WAF, Cloudflare Enterprise's interactive challenges, anything that requires canvas/WebGL fingerprinting — `webfetch` will get a 403 and surface it. The tool description warns the model so it doesn't retry. Solving that needs a stealth browser (Camoufox/Patchright + residential sticky proxies), which is a separate architectural conversation.

3. **The two-path dispatch is a real branch in production behaviour.** `isCurlImpersonateAvailable()` is cached per-process — if the binary disappears mid-run (e.g. mid-flight `apt remove`), we'll keep using the stale `true` and start failing all webfetches. This is acceptable because (a) the binary is installed by the Dockerfile and lives in the read-only image layer, (b) production containers don't mutate it, and (c) the failure mode is "every webfetch errors clearly" not "every webfetch hangs."

4. **No header / cookie state across requests.** This PR does not add a cross-request cookie jar. Akamai's `_abck` cookie dance works *within* a single redirect chain (curl's `-L` handles that), but multi-step authenticated flows would need a persistent jar — out of scope here, additive later if a real need shows up.

5. **No proxy rotation.** Discussed and explicitly deferred. The residential-IP-from-home assumption removes the *need* for rotation for the target sites, and rotating *would* break Akamai's session continuity (a single home IP across a session beats N random residential IPs from a pool for the `_abck` use case). If a future user wants to BYO proxy, the natural seam is a single env var / config field on the `curl-impersonate` primitive, additive.

## Out of scope

- Stealth browser fallback for Bloomberg.
- Cross-request cookie persistence.
- BYO proxy support.
- Header / Accept-Language tuning based on user locale.
- Migrating `wikipedia.ts` and `multimodal/looker.ts` (the other Bun.fetch callers) onto the same primitive — they hit well-behaved APIs that don't fingerprint, so the cost/benefit isn't there. The primitive is now available if that changes.

### DIFF
--- a/src/agent/tools/curl-impersonate.test.ts
+++ b/src/agent/tools/curl-impersonate.test.ts
@@ -1,0 +1,225 @@
+import { afterEach, beforeEach, describe, expect, test } from 'bun:test'
+import { chmodSync, mkdtempSync, rmSync, writeFileSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+
+import {
+  CurlImpersonateError,
+  _resetAvailabilityCacheForTest,
+  _setCurlBinaryForTest,
+  curlImpersonate,
+  isCurlImpersonateAvailable,
+} from './curl-impersonate'
+
+const SENTINEL = '\n--TYPECLAW-CURL-META-9c3f5e4d2a1b4f8e9c7a6b5d4e3f2a1b0--\n'
+
+describe('curlImpersonate', () => {
+  let scratchDir: string
+
+  beforeEach(() => {
+    scratchDir = mkdtempSync(join(tmpdir(), 'curl-impersonate-test-'))
+    _resetAvailabilityCacheForTest()
+  })
+
+  afterEach(() => {
+    _setCurlBinaryForTest(null)
+    _resetAvailabilityCacheForTest()
+    rmSync(scratchDir, { recursive: true, force: true })
+  })
+
+  function installFakeBinary(script: string): void {
+    const path = join(scratchDir, 'fake-curl')
+    writeFileSync(path, `#!/bin/sh\n${script}\n`, 'utf8')
+    chmodSync(path, 0o755)
+    _setCurlBinaryForTest(path)
+  }
+
+  test('parses body, status, finalUrl, contentType from sentinel-separated stdout', async () => {
+    // given: fake binary that writes a body, then the sentinel, then metadata
+    installFakeBinary(
+      `printf '<html>hi</html>'; printf '${SENTINEL}200\nhttps://example.com/final\ntext/html; charset=utf-8\n14\n'`,
+    )
+
+    // when
+    const result = await curlImpersonate({ url: 'https://example.com' })
+
+    // then
+    expect(result.body).toBe('<html>hi</html>')
+    expect(result.httpStatus).toBe(200)
+    expect(result.finalUrl).toBe('https://example.com/final')
+    expect(result.contentType).toBe('text/html; charset=utf-8')
+    expect(result.bytesIn).toBe(14)
+  })
+
+  test('returns body verbatim with status=0 when sentinel is absent', async () => {
+    // given: legacy-style fake that emits only the body (e.g. very early abort)
+    installFakeBinary("printf 'raw body without metadata'")
+
+    // when
+    const result = await curlImpersonate({ url: 'https://example.com' })
+
+    // then
+    expect(result.body).toBe('raw body without metadata')
+    expect(result.httpStatus).toBe(0)
+    expect(result.finalUrl).toBe('')
+  })
+
+  test('preserves a 404 body and reports httpStatus=404 (does NOT throw)', async () => {
+    // given: server-emulating fake returns a 404-shaped response
+    installFakeBinary(`printf 'not found'; printf '${SENTINEL}404\nhttps://example.com/missing\ntext/plain\n9\n'`)
+
+    // when
+    const result = await curlImpersonate({ url: 'https://example.com/missing' })
+
+    // then: caller-decides-on-status policy
+    expect(result.httpStatus).toBe(404)
+    expect(result.body).toBe('not found')
+  })
+
+  test('throws CurlImpersonateError with stderr detail on non-zero exit', async () => {
+    // given
+    installFakeBinary('echo "tls handshake failed" >&2; exit 35')
+
+    // when / then
+    await expect(curlImpersonate({ url: 'https://example.com' })).rejects.toThrow(CurlImpersonateError)
+    await expect(curlImpersonate({ url: 'https://example.com' })).rejects.toThrow(/exited 35/)
+    await expect(curlImpersonate({ url: 'https://example.com' })).rejects.toThrow(/tls handshake failed/)
+  })
+
+  test('reports "no stderr" when the binary fails silently', async () => {
+    installFakeBinary('exit 7')
+
+    await expect(curlImpersonate({ url: 'https://example.com' })).rejects.toThrow(/exited 7.*no stderr/)
+  })
+
+  test('passes method, url, and form fields through argv', async () => {
+    // given: fake captures argv to a side file
+    const argvFile = join(scratchDir, 'argv.txt')
+    installFakeBinary(`printf '%s\\n' "$@" > ${argvFile}; printf ''`)
+
+    // when
+    await curlImpersonate({
+      url: 'https://example.com/search',
+      method: 'POST',
+      formFields: [
+        { name: 'q', value: 'hello world' },
+        { name: 'lang', value: 'en' },
+      ],
+    })
+
+    // then
+    const argv = (await Bun.file(argvFile).text()).split('\n')
+    expect(argv).toContain('-X')
+    expect(argv).toContain('POST')
+    expect(argv).toContain('--data-urlencode')
+    expect(argv).toContain('q=hello world')
+    expect(argv).toContain('lang=en')
+    expect(argv).toContain('https://example.com/search')
+  })
+
+  test('passes extra headers as -H arguments', async () => {
+    const argvFile = join(scratchDir, 'argv.txt')
+    installFakeBinary(`printf '%s\\n' "$@" > ${argvFile}; printf ''`)
+
+    await curlImpersonate({
+      url: 'https://example.com',
+      extraHeaders: { Authorization: 'Bearer abc', 'X-Trace': 'xyz' },
+    })
+
+    const argv = (await Bun.file(argvFile).text()).split('\n')
+    expect(argv).toContain('-H')
+    expect(argv).toContain('Authorization: Bearer abc')
+    expect(argv).toContain('X-Trace: xyz')
+  })
+
+  test('follows redirects by default (passes --location)', async () => {
+    const argvFile = join(scratchDir, 'argv.txt')
+    installFakeBinary(`printf '%s\\n' "$@" > ${argvFile}; printf ''`)
+
+    await curlImpersonate({ url: 'https://example.com' })
+
+    const argv = (await Bun.file(argvFile).text()).split('\n')
+    expect(argv).toContain('--location')
+    expect(argv).toContain('--max-redirs')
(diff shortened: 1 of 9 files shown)


devxoul:
## Self-review pass (commit `7175cd5`)

Ran self-review with Oracle (read-only consultation). Found seven issues, all fixed in `7175cd5` and covered by new tests.

### Security / correctness

1. **`.curlrc` config injection** — added `--disable` as the **first** curl argument. Without it, a stray user/system curlrc could inject `--proxy`, `--header`, `--resolve`, `--no-location`, etc. and silently subvert both the Chrome impersonation contract and the protocol allowlist below. Order is load-bearing (curl ignores `--disable` if any other flag precedes it).

2. **Redirect-protocol escape (SSRF-adjacent)** — added `--proto =http,https` and `--proto-redir =http,https`. `normalizeUrl()` only validates the *initial* URL; with `--location`, a 301 to `ftp://...` was being followed by curl after the gate had fired. The `=...` syntax means "ONLY these two," not "add these to defaults."

3. **Sentinel spoofing** — the previous revision used a static metadata sentinel + first-occurrence parsing. webfetch reads attacker-controlled pages: a hostile response could include the literal sentinel followed by fake metadata, and `indexOf` would split at the injected occurrence. Now the sentinel is per-request 96-bit random and the parser anchors on the **last** occurrence (curl writes `-w` strictly after the body, so the real block is always trailing). Both defenses are needed — random alone fails to replay, last-match alone fails to suffix-injection.

4. **Argv injection via URL** — added `--` before the URL so a string beginning with `-` (e.g. `-K /etc/passwd` sneaking past `normalizeUrl` as `https://-K /etc/passwd`) cannot be reparsed as a curl option.

5. **Wrong exit-code mapping for `--max-filesize`** — Oracle verified empirically that curl has **two** exit shapes for the size overflow: `63` (content-length precheck) and `56` + stderr `Exceeded the maximum allowed file size (...)` (transfer-time discovery, when the server omits Content-Length). Previous code only handled `63`. Also replaced the locale-dependent stderr regex for timeout (`/max-time/i`, `/timed? out/i`) with the canonical exit `28` check. New helpers `isCurlExitTimeout()` and `isCurlExitFilesizeExceeded()` centralize the policy.

6. **Test environment dependency** — `webfetch/tool.test.ts` was mocking `globalThis.fetch` but not forcing the Bun.fetch fallback. On a dev/CI environment where `curl_chrome136` happens to be installed (e.g. a CI runner that pre-bakes the toolchain), those tests would have bypassed the mock and hit a real curl + real network. Now forces fallback in `beforeEach`/`afterEach`.

### YAGNI cleanup

7. **`extraHeaders`** removed from `CurlImpersonateRequest`. No caller used it and it's a documented footgun (per AGENTS.md, custom headers corrupt the Chrome impersonation). Add it back when a real caller needs it. Likewise dropped `HEAD` from the method union — `-X HEAD` is not equivalent to `curl --head`, and no caller wanted it.

### Polish

- `ddg.ts` now lets `CurlImpersonateError` propagate verbatim instead of rewrapping as `new Error(error.message)`. Structured `exitCode` and `stderr` fields survive for diagnostics; existing tests still match via the `.message` regex.
- Tool description toned down per Oracle: "helps with TLS/header fingerprint gates" replaces "clears most Cloudflare/Akamai gates," with an explicit list of what it does NOT solve (JS challenges, behavioural fingerprinting, interactive CAPTCHAs, IP reputation).
- Missing-sentinel in `parseCurlOutput` now throws instead of silently returning `httpStatus: 0`. Wrapper/output corruption should be loud.

### Test additions (all in this commit)

| Behaviour | Where |
|---|---|
| `--disable` is first argv | `curl-impersonate.test.ts` |
| `--proto =http,https` + `--proto-redir =http,https` present | ↑ |
| `--` before URL | ↑ |
| Body containing legacy static sentinel does not corrupt parsing | ↑ (anti-spoofing test) |
| Per-request sentinel round-trip in fake binaries | all three test files (`fetch.test.ts`, `curl-impersonate.test.ts`, `ddg.test.ts`, `websearch.test.ts`) |
| `isCurlExitTimeout` / `isCurlExitFilesizeExceeded` for exit 28 / 63 / 56-with-filesize / 56-without-filesize | `curl-impersonate.test.ts` |
| webfetch translates exit 28 → "timed out" | `webfetch/fetch.test.ts` |
| webfetch translates exit 63 → "Response too large" | ↑ |
| webfetch translates exit 56 + filesize stderr → "Response too large" | ↑ |
| webfetch does NOT misclassify generic exit 56 (recv failure) | ↑ |
| `tool.test.ts` forces fallback | `webfetch/tool.test.ts` |

### Gates

```
bun run typecheck   ✓  clean
bun run lint        ✓  0 errors (17 pre-existing warnings in unrelated files)
bun run format      ✓  clean
bun test            ✓  4206 pass / 0 fail (was 4194 pre-self-review)
```

The total test count went from 4194 → 4206 (12 new behaviour tests added net).

### Diff summary

```
src/agent/tools/curl-impersonate.test.ts | 225 +++++++++++++++++++++----------
src/agent/tools/curl-impersonate.ts      | 180 +++++++++++++++----------
src/agent/tools/ddg.test.ts              |  27 +++-
src/agent/tools/ddg.ts                   |  23 ++--
src/agent/tools/webfetch/fetch.test.ts   |  78 +++++++++--
src/agent/tools/webfetch/fetch.ts        |  16 ++-
src/agent/tools/webfetch/tool.test.ts    |   6 +
src/agent/tools/webfetch/tool.ts         |   6 +-
src/agent/tools/websearch.test.ts        |  22 ++-
9 files changed, 402 insertions(+), 181 deletions(-)
```

### What Oracle said was fine (so I didn't touch)

- `availabilityCache` without TTL — the binary doesn't appear/disappear at runtime in a sealed container image.
- Singleflight on availability check — YAGNI; the duplicate spawn race only happens once at cold start and both branches converge.
- The per-request `await isCurlImpersonateAvailable()` overhead — cached after first call, negligible.

Ready for another look.

PR #261, opened 2026-05-20
models: per-profile fallback chain with cron-side retry

devxoul:
## Why

Unattended cron jobs are the first place a hard model failure costs real work: nobody's watching to manually retry, and the job just silently dies. This PR makes model failures recoverable by letting operators declare a fallback chain per profile.

```jsonc
{
  "models": {
    "default": ["openai/gpt-5.4-nano", "fireworks/accounts/fireworks/routers/kimi-k2p6-turbo"]
  }
}
```

The single-string shape (`"default": "openai/..."`) keeps working — the schema accepts both and normalises to `KnownModelRef[]` at parse time. Existing configs need no changes.

### Schema (`src/config/config.ts`)

`models.<profile>` now accepts `string | string[]`. The union+transform normalises both to `KnownModelRef[]` before any downstream code sees the value, so there's no `string vs string[]` branching anywhere in the codebase. The `Models` type guarantees `default` is always present (matches the schema's `.refine`). `FIELD_EFFECTS['models']` stays `'applied'` — chain edits take effect without a restart.

`resolveProfile` returns both `ref` (head, for callers that don't care about fallback) and `refs` (full chain, for fallback-aware callers). Existing callers are unaffected.

### CLI display (`src/cli/model.ts`, `src/config/models-mutation.ts`)

`ModelProfileEntry` gains `refs[]` and `missingProviders[]`. Credential status is `missing-credentials` if **any** provider in the chain lacks credentials. `typeclaw model list` renders each fallback ref indented under the head with a `↳` glyph and a `(+N fallback)` summary. Single-ref profiles render unchanged.

### Fallback helper (`src/agent/model-fallback.ts`)

`promptWithFallback(refs, text, createSessionForRef, options)` is the new primitive. It iterates the chain: creates a session via the caller's factory, awaits `session.prompt(text)`, and on failure disposes the session and advances. Two failure modes trigger fallback:

- **Hard throw** from `session.prompt()` — network drop, timeout, provider rejecting the request.
- **Soft provider error** — pi-coding-agent encodes upstream LLM failures (`billing`, `rate-limit`, malformed response) in the final assistant message with `stopReason: 'error'` instead of throwing. The helper subscribes via `subscribeProviderErrors` and checks the captured `softError` after `prompt()` resolves.

After exhausting the chain, it logs `[cron] <jobId>: all N model(s) failed; last error: <msg>` and re-throws.

Hook semantics: `session.turn.start` / `turn.end` fire once per attempt; `session.idle` only fires on success (matches the memory plugin's pre-existing expectation — dreaming should not trigger on a failed turn); `session.end` always fires.

### Cron consumer (`src/cron/consumer.ts`)

`createSessionForCron` grows an optional `refOverride?: KnownModelRef` parameter. When `promptWithFallback` advances to the next ref in the chain, it passes it as `refOverride`, which threads through to `createSession({ refOverride })` — bypassing profile resolution and pinning the session to the specific entry. The consumer's retry loop disposes the failed session, creates a new one pinned to the next ref, and replays the same prompt.

`CreateSessionOptions.refOverride` is the new escape hatch that makes per-attempt model pinning possible without touching the live profile config.

## Honest tradeoffs

**Cron-only for now.** The server drain (TUI) and channels router both hold long-lived sessions with many subscribers: provider-error forwarding, restart-notice, broadcast, prompt enqueue, channel typing indicators. Swapping the underlying session mid-conversation requires resubscribing all of those without dropping events. That's a bigger refactor than v1 can absorb cleanly. TUI users can manually retry; channel adapters already surface the soft error to the chat. Cron is where automatic retry delivers the most value today — unattended, no human to catch it.

**No CLI for multi-ref chains.** `typeclaw model set` still takes a single ref. Users hand-edit `typeclaw.json` for now. The single-string shorthand means minimal-fallback configs are one line; the ergonomics are acceptable until there's a clearer UX for the multi-step wizard.

**Chain exhaustion is a hard failure, not a notification.** When every ref in the chain fails, the cron job logs and moves on. There's no dead-letter queue or TUI alert. A future PR could emit a `broadcast` stream message so the TUI surfaces the exhaustion, but that's new surface area; the log line is sufficient for v1.

## Test coverage

```
src/agent/model-fallback.test.ts     +4 new tests (fallback helper: hard throw, soft error, chain exhaustion, success-on-second)
src/cron/consumer.test.ts            +2 new tests (cron integration: fallback on hard throw, fallback on soft error)
src/config/models-mutation.test.ts   +2 new tests (chain normalisation, missingProviders aggregation)
```

Total: 4198 pass, 0 fail (was 4194 pre-PR).

## Verification

```
bun run typecheck      clean
bun run lint           0 errors (17 pre-existing warnings in src/plugin/define.test.ts, src/cli/tunnel.ts, and other unrelated files)
bun run format         clean
bun test               4198 pass, 0 fail
```

## Out of scope

- **Server drain (TUI) fallback** — requires session-swap without dropping subscribers; deferred.
- **Channels router fallback** — same session-swap problem; deferred.
- **Subagent fallback** — subagents use `createSession` directly; would need the same `refOverride` threading; deferred.
- **`typeclaw model set` multi-ref** — CLI still takes a single ref; users hand-edit for chains.

### DIFF
--- a/src/agent/auth.ts
+++ b/src/agent/auth.ts
@@ -83,8 +83,10 @@ export function getAuthFor(providerId: KnownProviderId): Auth {
 
 // Back-compat shim for callers that still want the `default` profile's auth
 // (the main session path). Equivalent to `getAuthFor(provider-of-default)`.
+// Uses the head of the fallback chain; auth for the rest of the chain is
+// resolved lazily when fallback actually fires.
 export function getAuth(): Auth {
-  const defaultRef = getConfig().models.default
+  const defaultRef = getConfig().models.default[0]!
   return getAuthFor(providerForModelRef(defaultRef))
 }
 
@@ -98,7 +100,7 @@ function hasAnyCredentialInEnv(apiKeyEnv: string | null): boolean {
 
 function missingCredentialMessage(providerId: KnownProviderId): string {
   const provider = KNOWN_PROVIDERS[providerId]
-  const defaultRef = getConfig().models.default
+  const defaultRef = getConfig().models.default[0]!
   const defaultProviderId = providerForModelRef(defaultRef)
   // For the `default` profile, name the model in the error message (matches
   // pre-multi-model behavior). For any other profile, the user is mixing
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -8,7 +8,7 @@ import type { AgentSession, ToolDefinition } from '@mariozechner/pi-coding-agent
 import { loadMemory } from '@/bundled-plugins/memory/load-memory'
 import type { ChannelRouter } from '@/channels/router'
 import { getConfig, resolveModel, resolveProfile } from '@/config'
-import { providerForModelRef } from '@/config/providers'
+import { providerForModelRef, type KnownModelRef } from '@/config/providers'
 import type { PermissionService } from '@/permissions'
 import type {
   BuiltinToolRef,
@@ -134,6 +134,12 @@ export type CreateSessionOptions = {
   // overrides) so different sessions on the same agent can run different
   // models without per-session config edits.
   profile?: string
+  // Override the resolved ref directly, bypassing `profile` resolution. Used
+  // by the model-fallback helper (`promptWithFallback`) to recreate a session
+  // pinned to the next ref in the chain after the previous one failed. When
+  // set, `profile` is still recorded for the fallback-warning bookkeeping;
+  // the profile→refs resolution is skipped.
+  refOverride?: KnownModelRef
   // Defensive ceiling on cumulative bytes of tool-result text per session,
   // applied to the named tools only. See `src/agent/tool-result-budget.ts`
   // for the rationale. Intended for subagents that read large files
@@ -164,7 +170,10 @@ export async function createSessionWithDispose(options: CreateSessionOptions = {
   if (resolved.fellBackToDefault && options.profile !== undefined && options.profile !== 'default') {
     warnProfileFallbackOnce(options.profile, resolved.ref)
   }
-  const { authStorage, modelRegistry } = getAuthFor(providerForModelRef(resolved.ref))
+  // `refOverride` lets the model-fallback helper pin a specific entry from
+  // the chain when it recreates a session after the previous ref failed.
+  const activeRef: KnownModelRef = options.refOverride ?? resolved.ref
+  const { authStorage, modelRegistry } = getAuthFor(providerForModelRef(activeRef))
 
   const materializedSkills =
     options.plugins && options.plugins.registry.skills.length > 0
@@ -279,7 +288,7 @@ export async function createSessionWithDispose(options: CreateSessionOptions = {
       ? customToolsPreBudget.map((t) => wrapToolDefinitionWithBudget(t, sessionBudget, sessionBudgetState))
       : customToolsPreBudget
 
-  const model = resolveModel(resolved.ref)
+  const model = resolveModel(activeRef)
   const { session } = await createAgentSession({
     model,
     sessionManager,
--- a/src/agent/model-fallback.test.ts
+++ b/src/agent/model-fallback.test.ts
@@ -1,0 +1,243 @@
+import { describe, expect, test } from 'bun:test'
+
+import type { Models } from '@/config/config'
+import type { KnownModelRef } from '@/config/providers'
+
+import type { AgentSession } from './index'
+import { promptWithFallback, resolveFallbackChain } from './model-fallback'
+
+const REF_A = 'openai/gpt-5.4-nano' as KnownModelRef
+const REF_B = 'fireworks/accounts/fireworks/routers/kimi-k2p6-turbo' as KnownModelRef
+const REF_C = 'zai/glm-4.6' as KnownModelRef
+
+type SessionFake = {
+  ref: KnownModelRef
+  prompt: (text: string) => Promise<void>
+  // The subscribe field accepts the event callback shape AgentSession uses
+  // for `message_end`. We hand-roll it to simulate soft-error emission
+  // without standing up a real pi-coding-agent session.
+  emitSoftError: (msg: string) => void
+  disposed: boolean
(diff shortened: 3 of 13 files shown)


devxoul:
## Self-review — 2 follow-up commits

Ran a critical-review pass on the original 4 commits. Two real correctness issues + two test-quality issues caught; addressed in 2 follow-up commits (`037c1f6` cron lifecycle fix, `54fe55e` schema dedup rejection).

### What was wrong with the original commits

**1. (BLOCKING) Successful sessions leaked.** `runPromptOnce` fired `runSessionIdle` and `runSessionEnd` against the final session but never called `result.dispose()` — pre-fallback cron disposed via `session.dispose?.()` in the outer `finally`. `promptWithFallback` returns `{ session, dispose }` precisely so the caller can run post-prompt hooks against a live session, but the dispose was never invoked. Every successful cron prompt leaked its `AgentSession` (and its pi event listeners, retry promises, queued state).

**2. Failed non-final attempts orphaned plugin state.** The original code only fired `session.end` against the LAST session created. Plugin state keyed by sessionId — security plugin's remote-taint map (`clearSessionTaints(sessionId)` runs on `session.end`) and memory plugin's debounce timer (`cancelTimer(sessionId)` runs on `session.end`) — orphaned for every failed attempt. Per failed cron run with N refs, N-1 sessions left state behind forever.

Fix landed in `037c1f6`: each per-attempt dispose now fires `runSessionTurnEnd` + `runSessionEnd` for that attempt's session, then calls `created.dispose?.()`. `runSessionIdle` still fires only once on success (matching pre-fallback semantics). After the helper returns success, `await result.dispose()` tears down the final session.

Added regression test `disposes the successful final session and fires session.end exactly once per attempted session` that drives a 2-ref chain (first fails, second succeeds) and asserts the exact hook+dispose sequence for **both** sessions. **Mutation-checked**: reverting `result.dispose()` flips the success-case dispose assertion; removing per-attempt `runSessionEnd` flips the failed-case end assertion.

### What was wrong with the original tests

**3. `attempts[]` ref-order untested.** The fallback-helper test `reports success: false when every ref in the chain fails` asserted `attempts[].outcome` (all 'hard') and `lastError`, but never `attempts[].ref`. The documented "attempts lists every ref in order" contract was untested.

Fix in `54fe55e`: added `expect(result.attempts.map((a) => a.ref)).toEqual([REF_A, REF_B, REF_C])`.

**4. Cron all-refs-failed test only asserted log format.** `logs final-attempt failure when every ref in the chain fails` checked the error log mentioned the chain length and final provider error but never asserted both refs were actually attempted in chain order.

Fix in `037c1f6` (alongside the lifecycle fix): added an `attempted: string[]` array tracked inside the factory and an assertion that `attempted === ['openai/...', 'fireworks/...']`. Turns it from a log-format test into a real fallback-behavior test.

### Schema hardening

Self-review also surfaced that the schema silently accepted `{ default: ["openai/x", "openai/x"] }`. Retrying the exact same ref after the same class of failure is almost certainly a config typo. Different models from the same provider (e.g. `["openai/gpt-5.4-nano", "openai/gpt-5.4-mini"]`) remain valid because they hit distinct upstream endpoints.

Fix in `54fe55e`: a `.refine` rejecting exact duplicate refs with a clear error message. The JSON Schema can't express "uniqueItems within an enum-typed array" cleanly so editor-time validation doesn't flag it; the runtime catches it at `typeclaw start` / `typeclaw reload`.

### What was NOT a problem (one of the two oracle reviewers got this wrong)

**Soft-error timing assumption.** One reviewer claimed `subscribeProviderErrors` could miss `stopReason: 'error'` because pi-coding-agent's `AgentSession._handleAgentEvent` queues `_processAgentEvent` asynchronously and `agent.prompt()` doesn't explicitly `await this._agentEventQueue`. Verified directly against the pi-coding-agent and pi-agent-core source:

- pi-agent-core's `Agent.processEvents` awaits listeners inline (`agent.js:386-392`), but `_handleAgentEvent` returns `undefined` synchronously after chaining the microtask
- Every subsequent `await emit({ ... })` in `runAgentLoop` is a microtask checkpoint that drains `_agentEventQueue`
- `message_end` for the assistant fires well before `agent_end` in the loop, with multiple `await emit` calls between them
- After `agent.prompt()` returns, `AgentSession.prompt` immediately `await this.waitForRetry()` provides another microtask checkpoint

Empirical confirmation: the **channels router** (`src/channels/router.ts:790-792`) and **subagents** (`src/agent/subagents.ts:160`) have used this exact subscribe-then-await-prompt pattern in production for the entire codebase's history. If there were a real race, those would be broken. They're not.

**`session.idle` semantics.** Same reviewer claimed I regressed `session.idle` from "after every prompt completion" to "success only". Confirmed via `git show HEAD~3:src/cron/consumer.ts` that pre-fallback cron's structure was:

```
try {
  prompt()       // throws → jumps past idle to outer finally
  idle()         // only on success
} finally {
  end()
  dispose()
}
```

So pre-fallback cron also only fired `idle` on success. **No regression.** The reviewer was confusing cron's contract with server drain's (which DOES fire idle after every prompt regardless of error — `src/server/index.ts:795`).

### Final verification

```
bun run typecheck      clean
bun run lint           0 errors (17 pre-existing warnings)
bun run format         clean
bun test               4202 pass / 0 fail  (was 4198 before self-review fixes; +1 disposal test, +1 dup-ref schema test, +1 same-provider chain test, +1 ref-order assertion strengthening an existing test)
```

Commit list:
1. `4401d69` models: accept a per-profile fallback chain (string or string[])
2. `dd64320` model: surface fallback chains in `typeclaw model list`
3. `ef35d3d` agent: add `promptWithFallback` helper for chain-based model retry
4. `baacb82` cron: retry against the next model in the chain when a prompt fails
5. `037c1f6` **cron: fix session leak and orphaned per-session plugin state on fallback** ← self-review fix
6. `54fe55e` **models: reject duplicate refs in a fallback chain** ← self-review hardening

PR #286, opened 2026-05-21
providers: add Anthropic Claude (Opus 4.7 / Sonnet 4.6 / Haiku 4.5) with api-key + OAuth

devxoul:
## Summary

Adds Anthropic Claude as a new LLM provider supporting both `ANTHROPIC_API_KEY` (Console / API billing) and OAuth against a Claude Pro/Max/Team/Enterprise subscription. The user originally asked to "implement Claude Code API as a channel"; we clarified that they meant **as a provider** (not a channel adapter), so this is the right shape for the request.

Anthropic is the **first dual-auth provider** in the registry — both methods reach the same `/v1/messages` endpoint on `api.anthropic.com`, so they share one provider id (unlike `openai` vs `openai-codex`, where the Codex endpoint is on a different host and demands a separate entry). When OAuth credentials are present, pi-ai's `anthropic-messages` transport injects Claude Code identity headers (`anthropic-beta: claude-code-20250219,oauth-2025-04-20` + `user-agent: claude-cli/<version>`) — i.e. the exact surface a subscriber's `claude setup-token` credential authorizes. The OAuth dance itself is authorization-code + PKCE with a localhost callback, not device-code.

## Model lineup

Ships the current GA tier verified against [Anthropic's model overview](https://platform.claude.com/docs/en/about-claude/models/overview) as of 2026-04-16:

| Tier | Model ID | Context | Max output | Pricing (in / out / cacheRead / cacheWrite-5m) |
|---|---|---|---|---|
| Top | `claude-opus-4-7` | 1M | 128k | \$5 / \$25 / \$0.50 / \$6.25 |
| Mid | `claude-sonnet-4-6` | 1M | 64k | \$3 / \$15 / \$0.30 / \$3.75 |
| Fast | `claude-haiku-4-5` | 200k | 64k | \$1 / \$5 / \$0.10 / \$1.25 |

### Two semantic shifts worth flagging at the registry site

1. **ID semantics differ across the lineup.** Per [Anthropic's model-id docs](https://platform.claude.com/docs/en/about-claude/models/model-ids-and-versions): for pre-4.6 models, dateless ids like `claude-haiku-4-5` are **convenience aliases** that resolve to the latest dated snapshot. Starting with the 4.6 generation, dateless ids like `claude-sonnet-4-6` and `claude-opus-4-7` are **pinned snapshots**, not aliases. The Haiku entry auto-rotates when Anthropic ships a new dated snapshot; Sonnet/Opus require explicit edits to advance. Both behaviors are documented in load-bearing comments at the registry site.

2. **Opus 4.7 ships a new tokenizer.** Same input maps to **1.0–1.3× more tokens** vs prior generations depending on content type. Per-token price is unchanged from 4.6, but identical workloads can cost meaningfully more. Documented at the registry site for anyone tracking spend.

## CLI integration

All three commands enumerate `KNOWN_PROVIDERS` dynamically via `Object.keys()` and `KnownProviderId`-typed exhaustive `Record`s. The new provider and its three model refs flow through automatically with **zero additional plumbing**:

- **`typeclaw init`** — Anthropic appears in the provider picker with hint "API key or OAuth"; the auth-method picker (`pickAuthMethod`) presents both options when the user selects it.
- **`typeclaw provider add`** — accepts `anthropic` as the provider id; `--oauth` routes to OAuth, otherwise prompts for an API key.
- **`typeclaw model list / set / add`** — `anthropic/claude-haiku-4-5`, `anthropic/claude-sonnet-4-6`, and `anthropic/claude-opus-4-7` appear in the available-models list and can be assigned to any profile.

Wires `anthropic` into `PROVIDER_TO_MODELS_DEV` so the init picker also surfaces live Anthropic models from models.dev alongside the curated entries.

## Runtime precedence (hidden in pre-existing rule, now user-visible)

Anthropic is the first dual-auth provider, so an existing rule in `src/agent/auth.ts:53-71` becomes user-visible for the first time: **when an OAuth credential exists on disk in `secrets.json`, `ANTHROPIC_API_KEY` in `.env` is IGNORED.** Env-wins only applies to api-key-shaped credentials; OAuth-on-disk takes precedence. A user who init'd with OAuth and later adds `ANTHROPIC_API_KEY` to `.env` expecting to switch will silently stay on OAuth. The mitigation pointer is `typeclaw provider remove anthropic`. Documented inline at the provider entry.

## OAuth UX bug fix (orphaned post-login prompt)

While testing the OAuth flow end-to-end, surfaced and fixed a clack-rendering bug that affected **all OAuth-capable providers** (anthropic, openai-codex, etc.), not just anthropic. Pi-ai races `onManualCodeInput()` against the local callback server ([`packages/ai/src/utils/oauth/anthropic.ts:210-253`](https://github.com/badlogic/pi-mono/blob/main/packages/ai/src/utils/oauth/anthropic.ts#L210-L253)). When the browser wins the race, pi-ai sets `result.code` and falls through to token exchange **without calling `server.cancelWait()` on the manual side** — the manual `text()` prompt is left dangling in clack's render pipeline and re-appears after each subsequent log line, AFTER the success message.

User-visible symptom (paraphrasing the report):

```
◇  Logged in to Anthropic. ───────────────────╮
│                                             │
│  If the agent is running:  typeclaw reload  │
│                                             │
◇  If your browser shows "this site can't be reached" after you sign in,
   copy the full address from the top of the browser and paste it here:
```

Fix: `buildOAuthCallbacks` now returns `{ callbacks, dispose }`. The dispose hook aborts an internal `AbortController` whose signal is plumbed into both `text()` prompts via clack's `signal` option. Each call site (`typeclaw init`, `typeclaw provider add`) calls `dispose()` in a finally after the OAuth runner returns; clack honors the signal by resolving the orphaned prompt with cancel state, and pi-ai's outer `.catch()` (line 219) swallows the resulting cancel-throw silently since it has stopped awaiting the manual promise on the winning-browser path.

## Test coverage

Three new wizard tests in `src/cli/init.test.ts` cover the previously-uncovered dual-auth branches against `anthropic` directly:

- `pickAuthMethod` is a REAL prompt (not `autoValue`) for a dual-auth provider, with both options visible.
- OAuth path fires `runOAuthLogin` with the correct anthropic ref.
- OAuth failure offers the api-key fallback with `apiKeyAvailable=true` (distinct from oauth-only providers, where it sees `apiKeyAvailable=false`).

Existing oauth-path tests used `openai` (api-key-only, with the oauth branch mocked) or `openai-codex` (oauth-only). Anthropic is the first real exercise of these code paths in production.

## Files changed

- `src/config/providers.ts` — new `anthropic` entry (3 models, load-bearing comments)
- `src/init/models-dev.ts` — one-line addition to `PROVIDER_TO_MODELS_DEV`
- `src/config/providers.test.ts` — auth-method invariants + new model ref assertions
- `src/cli/init.test.ts` — three dual-auth wizard tests
- `src/cli/oauth-callbacks.ts` — `{ callbacks, dispose }` shape with abort plumbing
- `src/cli/init.ts` — wraps `runOAuthLogin` in try/finally to call dispose
- `src/cli/provider.ts` — same try/finally wrapping at the provider-add OAuth call

## Validation

- `bun run typecheck` ✓
- `bun run lint` ✓ (18 pre-existing warnings, 0 errors, none in changed files)
- `bun run format:check` ✓
- `bun run test` — 4564 / 4564 pass (added 4 new tests)

## User-visible behavior after merge

- `typeclaw init` shows Anthropic in the provider picker.
- Users can pick API key (paste `ANTHROPIC_API_KEY`) or OAuth (in-container PKCE flow with localhost callback, or paste a token from `claude setup-token` on their own machine — the existing `typeclaw-claude-code` skill documents the user-side flow).
- Successful OAuth login no longer leaves an orphaned "paste the redirect URL here" prompt rendering on top of the success message.
- The agent's main loop can run on Claude Haiku 4.5 / Sonnet 4.6 / Opus 4.7 against a Pro/Max subscription, no API billing required.

No follow-up migrations needed. The `Dockerfile` does not need a new toggle — pi-ai already carries the Anthropic SDK as a baseline dependency.

### DIFF
--- a/src/cli/init.test.ts
+++ b/src/cli/init.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, mock, test } from 'bun:test'
 
-import { KNOWN_PROVIDERS, type KnownProviderId } from '@/config/providers'
+import { KNOWN_PROVIDERS, supportsApiKey, supportsOAuth, type KnownProviderId } from '@/config/providers'
 import type { ModelOption } from '@/init/models-dev'
 
 import { collectWizardInputs, decideExistingApiKeyReuse, WizardAbortedError, type WizardPrompts } from './init'
@@ -93,6 +93,17 @@ describe('collectWizardInputs back-aware flow', () => {
     reasoning: true,
     curated: true,
     supportsVision: false,
+  }
+  const anthropicModel: ModelOption = {
+    providerId: 'anthropic',
+    providerName: 'Anthropic',
+    modelId: 'claude-sonnet-4-6',
+    modelName: 'Claude Sonnet 4.6',
+    ref: 'anthropic/claude-sonnet-4-6',
+    contextWindow: 1000000,
+    reasoning: true,
+    curated: true,
+    supportsVision: true,
   }
 
   function makeRecorder(): { steps: string[]; record: (name: string) => void } {
@@ -1089,5 +1100,75 @@ describe('collectWizardInputs back-aware flow', () => {
       expect(error).toBeInstanceOf(WizardAbortedError)
       expect((error as WizardAbortedError).oauthCredentialsSaved).toBe(false)
     }
+  })
+
+  test('dual-auth provider (anthropic): pickAuthMethod is REAL prompt with both options, not autoValue', async () => {
+    // given: anthropic supports both api-key and oauth. Unlike openai
+    //   (api-key-only) and openai-codex (oauth-only), pickAuthMethod must
+    //   actually ask the user — autoValue is wrong here. We assert the
+    //   prompt sees the live provider and that the wizard honors the
+    //   user's selection literally.
+    let seenProviderId: string | undefined
+    let seenBothAuthModes: boolean | undefined
+    const prompts = makePrompts({
+      loadCatalog: async () => ({ options: [anthropicModel], source: 'curated' }),
+      pickProvider: async () => ({ kind: 'value', value: 'anthropic' as KnownProviderId }),
+      pickModel: async () => ({ kind: 'value', value: anthropicModel }),
+      pickAuthMethod: async (provider) => {
+        seenProviderId = provider.id
+        seenBothAuthModes = supportsApiKey(provider) && supportsOAuth(provider)
+        return { kind: 'value', value: 'api-key' }
+      },
+      askApiKey: async () => ({ kind: 'value', value: 'sk-ant-test' }),
+    })
+
+    const result = await collectWizardInputs('/agent', prompts)
+
+    expect(seenProviderId).toBe('anthropic')
+    expect(seenBothAuthModes).toBe(true)
+    expect(result.llmAuth).toEqual({ kind: 'api-key', apiKey: 'sk-ant-test' })
+  })
+
+  test('dual-auth provider (anthropic): user picks oauth → runOAuthLogin fires with anthropic ref', async () => {
+    const loginCalls: Array<{ cwd: string; model: string; providerId: string }> = []
+    const prompts = makePrompts({
+      loadCatalog: async () => ({ options: [anthropicModel], source: 'curated' }),
+      pickProvider: async () => ({ kind: 'value', value: 'anthropic' as KnownProviderId }),
+      pickModel: async () => ({ kind: 'value', value: anthropicModel }),
+      pickAuthMethod: async () => ({ kind: 'value', value: 'oauth' }),
+      runOAuthLogin: async (provider, cwd, model) => {
+        loginCalls.push({ cwd, model, providerId: provider.id })
+        return { ok: true }
+      },
+    })
+
+    const result = await collectWizardInputs('/agent', prompts)
+
+    expect(loginCalls).toEqual([{ cwd: '/agent', model: 'anthropic/claude-sonnet-4-6', providerId: 'anthropic' }])
+    expect(result.llmAuth).toEqual({ kind: 'oauth-completed' })
+  })
+
+  test('dual-auth provider (anthropic): oauth failure offers api-key fallback (apiKeyAvailable=true)', async () => {
+    // given: anthropic's OAuth login fails — distinct from openai-codex,
+    //   where api-key fallback is unavailable. The recovery prompt MUST see
+    //   apiKeyAvailable=true so the wizard can route to enter-api-key.
+    let apiKeyAvailableSeen: boolean | undefined
+    const prompts = makePrompts({
+      loadCatalog: async () => ({ options: [anthropicModel], source: 'curated' }),
+      pickProvider: async () => ({ kind: 'value', value: 'anthropic' as KnownProviderId }),
+      pickModel: async () => ({ kind: 'value', value: anthropicModel }),
+      pickAuthMethod: async () => ({ kind: 'value', value: 'oauth' }),
+      runOAuthLogin: async () => ({ ok: false, reason: 'browser closed' }),
+      askOAuthFailureRecovery: async (_provider, _reason, apiKeyAvailable) => {
+        apiKeyAvailableSeen = apiKeyAvailable
+        return 'api-key'
+      },
+      askApiKey: async () => ({ kind: 'value', value: 'sk-ant-recovered' }),
+    })
+
+    const result = await collectWizardInputs('/agent', prompts)
+
+    expect(apiKeyAvailableSeen).toBe(true)
+    expect(result.llmAuth).toEqual({ kind: 'api-key', apiKey: 'sk-ant-recovered' })
   })
 })
--- a/src/cli/init.ts
+++ b/src/cli/init.ts
@@ -374,7 +374,14 @@ export const defaultWizardPrompts: WizardPrompts = {
   hasExistingChannelSecrets,
   askReuseExistingChannel,
   runChannelFlow,
-  runOAuthLogin: (provider, cwd, model) => makeOAuthLoginRunner(buildOAuthCallbacks(provider.name))({ cwd, model }),
+  runOAuthLogin: async (provider, cwd, model) => {
+    const { callbacks, dispose } = buildOAuthCallbacks(provider.name)
+    try {
+      return await makeOAuthLoginRunner(callbacks)({ cwd, model })
+    } finally {
+      dispose()
+    }
+  },
   askOAuthFailureRecovery,
 }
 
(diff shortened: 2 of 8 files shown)


devxoul:
## Self-review notes

Ran a three-thread audit on this PR (downstream consumers, init-wizard dual-auth handling, pi-ai source-of-truth verification). Two corrections were warranted and have been amended into the commit; two follow-ups are worth disclosing.

### Amended into this PR

- **Factual correction**: The original rationale comment implied a device-code flow. pi-ai's actual implementation (`packages/ai/src/utils/oauth/anthropic.ts`) is **authorization-code + PKCE with a localhost callback server** against `claude.ai/oauth/authorize`. Updated the comment to say so explicitly.
- **Model-alias caveat**: Anthropic's docs say dateless aliases (`claude-haiku-4-5`) are forward-compat **only for the 4.5 generation**. Starting with the 4.6 family, dateless ids are **pinned snapshots**, not evergreen aliases. Added a `WARNING` block so the next maintainer adding 4.6+ entries doesn't assume the same forward-rotation behavior.
- **Documented the runtime precedence**: Anthropic is the first dual-auth provider, which exercises an existing rule in `src/agent/auth.ts`: when an OAuth credential is on disk, `ANTHROPIC_API_KEY` in `.env` is ignored. Added the rule and the remediation pointer (`typeclaw provider remove anthropic`) to the comment so the surprise is discoverable at the source rather than via user bug reports.

### Disclosed follow-ups (intentionally not in this PR)

1. **No wizard test for dual-auth flow.** The existing `pickAuthMethod` and OAuth-failure-fallback tests in `src/cli/init.test.ts` use `openai` (api-key-only) with a mocked OAuth path. This PR makes Anthropic the first real exercise of those code paths in production. The wizard handles it correctly (verified by reading), but it would be worth adding an end-to-end test that uses `KNOWN_PROVIDERS.anthropic` and walks through the API-key vs OAuth selection prompt + the OAuth-failure → API-key recovery branch. Defer to a follow-up.
2. **Subscription-vs-API billing nuance** (not blocking). As of June 15 2026, Pro/Max plans include a capped monthly Agent SDK credit pool for `claude -p` / Agent SDK usage that can spill over to standard API rates if exceeded and extra-usage is enabled. The current comment says OAuth "authorizes against a subscription credential" — accurate for interactive Claude Code use, possibly understated for unattended typeclaw use. Worth revisiting if anyone reports surprise billing.

### Verified correct (no change needed)

- All three Claude 4.5 model ids exist as forward-compat aliases per Anthropic's official model-id docs.
- All pricing numbers (input / output / cacheRead / cacheWrite) match Anthropic's published pricing exactly; the `cacheWrite = 1.25× input` ratio is confirmed by Anthropic's own table.
- Every downstream consumer of `KNOWN_PROVIDERS` in `src/` is either typescript-exhaustive (`Record<KnownProviderId, ...>`) or uses runtime enumeration (`Object.keys`) — no hidden hardcoded provider whitelists, no incomplete switches.
- The init wizard's `pickAuthMethod` correctly prompts the user when both auth modes are supported, defaults safely to api-key, and offers a clean api-key fallback on OAuth failure.

Commit amended and force-pushed; the diff is now +110 / −8 across the same three files.

devxoul:
## Follow-up: corrected model lineup + dual-auth wizard tests

The original commit shipped the wrong model tier — `claude-opus-4-5` was two generations stale, and `claude-sonnet-4-5` was one generation stale. Anthropic's current GA lineup (verified against [platform.claude.com/docs/en/about-claude/models/overview](https://platform.claude.com/docs/en/about-claude/models/overview) and Anthropic's [Opus 4.7 release post](https://www.anthropic.com/news/claude-opus-4-7), Apr 16 2026):

| Tier | Was (stale) | Now (current) | Status |
|---|---|---|---|
| Top | `claude-opus-4-5` | **`claude-opus-4-7`** | Released Apr 16 2026 — 1M context, 128k max output |
| Mid | `claude-sonnet-4-5` | **`claude-sonnet-4-6`** | Released Feb 5 2026 — 1M context |
| Fast | `claude-haiku-4-5` | `claude-haiku-4-5` (unchanged) | Anthropic's current overview still lists this as the GA Haiku |

Pricing for all three is identical to the previous-gen entries it replaces (verified against [Anthropic's pricing page](https://platform.claude.com/docs/en/about-claude/pricing)). The cost numbers in the commit are correct.

### Two semantic shifts worth flagging at the registry site (now in load-bearing comments)

1. **ID semantics split across the lineup.** Per Anthropic's [model-id docs](https://platform.claude.com/docs/en/about-claude/models/model-ids-and-versions): for pre-4.6 models, dateless ids like `claude-haiku-4-5` are **convenience aliases** that resolve to the latest dated snapshot. Starting with the 4.6 generation, dateless ids like `claude-sonnet-4-6` and `claude-opus-4-7` are **pinned snapshots**, not aliases. The PR's Haiku entry auto-rotates; Sonnet/Opus require explicit edits to advance.

2. **Opus 4.7 ships a new tokenizer.** Same input maps to **1.0–1.3× more tokens** vs prior generations depending on content type. Per-token price is unchanged from 4.6 but identical workloads can cost meaningfully more. Worth noting for anyone tracking spend.

### Dual-auth wizard tests added

The self-review's "no dual-auth wizard test" gap is now closed. Three new tests in `src/cli/init.test.ts` exercise the previously-uncovered branches against `anthropic` directly:

- **`pickAuthMethod` is a REAL prompt, not `autoValue`** — asserts the wizard prompts the user when `supportsApiKey && supportsOAuth`, with the live provider visible to the prompt.
- **OAuth path fires `runOAuthLogin` with the anthropic ref** — asserts the right provider and model ref reach the OAuth runner.
- **OAuth failure offers api-key fallback** with `apiKeyAvailable=true` — asserts the recovery prompt sees the dual-auth provider's api-key escape hatch, distinct from `openai-codex` where it sees `apiKeyAvailable=false`.

Existing oauth-path tests used `openai` (api-key-only, with the oauth branch mocked) or `openai-codex` (oauth-only). Anthropic is the first real exercise of these code paths in production.

### Init / provider / model CLI commands

All three commands (`typeclaw init`, `typeclaw provider add|set|remove`, `typeclaw model list|set|add|remove`) enumerate `KNOWN_PROVIDERS` dynamically and use `KnownProviderId`-typed exhaustive `Record`s. The new `anthropic` provider and its three model refs flow through every CLI surface automatically with **zero additional plumbing**. Verified by reading every consumer in `src/cli/` and confirming no hardcoded provider id whitelists, no incomplete switches.

### Updated stats

- Diff: +135 / −25 across the same 3 files (one commit amended)
- Tests: 4564 / 4564 pass (was 4560 — added 3 wizard tests + 1 anthropic ref-routing test)
- All four pre-push checks: typecheck ✓ · lint 0 errors ✓ · format:check ✓ · test ✓

PR #296, opened 2026-05-21
claude-code: pre-seed ~/.claude.json so first launch skips the TTY-only onboarding wizard

devxoul:
## Problem

On the first run after `docker.file.claudeCode: true` builds, the `claude` binary enters a TTY-only onboarding wizard:

```
Welcome to Claude Code v2.1.146
Let's get started.
Choose the text style that looks best with your terminal
To change this later, run /theme
  1. Auto (match terminal)
  2. Dark mode
  ...
```

The wizard is a modal prompt with no CLI flag and no env var to skip it. (`anthropics/claude-code#4714`, `#8938`, `#13827` all confirm the only escape hatch is pre-seeding `~/.claude.json`. `--dangerously-skip-permissions` skips permission checks but not onboarding; `IS_DEMO=1` exists but mutates unrelated behavior — hides email, suppresses debug commands.)

The `typeclaw-claude-code` skill drives `claude` headlessly through `tmux send-keys` and detects turn-end via a `Stop` hook that touches a `.done` sentinel. Neither survives the wizard:

- `send-keys "<prompt>" Enter` arrives at the theme picker, gets interpreted as `Enter`-on-option-1, and the picker advances to the next modal instead of reaching `claude`'s prompt
- The `Stop` hook never fires while the wizard is up — there are no turns yet
- The polling loop only learns of the hang at the 10-minute wall-clock budget, after which it kills the worktree

Net effect: every first delegation to Claude Code silently times out at ~10 minutes with no useful output.

## Fix

Append one line to the existing `renderClaudeCodeInstallLayer`:

```dockerfile
RUN curl -fsSL https://claude.ai/install.sh | bash \
 && ln -sf "$HOME/.local/bin/claude" /usr/local/bin/claude \
 && claude --version > /dev/null \
 && printf '%s\n' '{"hasCompletedOnboarding":true,"theme":"dark","installMethod":"native","numStartups":1}' > "$HOME/.claude.json"
```

The seed contains only **onboarding-skip** flags. Trust dialogs (`hasTrustDialogAccepted`) and permission-bypass flags (`bypassPermissionsModeAccepted`, `dangerouslySkipPermissions`) are deliberately omitted — those should remain explicit user decisions, not silent Dockerfile defaults. A regression test pins this.

Order matters: the seed runs **after** `claude --version`, because the smoke test creates `~/.claude.json` with an unrelated default shape that would clobber our seed if it ran second.

Theme defaults to `dark` to match typeclaw's default TUI theme. Users on light terminals can override by editing `~/.claude.json` post-build (or by running `/theme` inside `claude` once they're past the wizard).

## Why this layer, not the skill

The skill explicitly forbids the agent from installing `claude` at runtime ("the install belongs in the Dockerfile layer, not at runtime"), so the seed has to land at image build time too — runtime seeding would diverge from where the install itself lives. The `claudeCode` toggle is the right gate: if `claude` isn't installed, there's no `~/.claude.json` to seed.

## Skill update

`SKILL.md` prerequisites now mention the pre-seed and ship a one-line self-heal `printf` for the edge case where `~/.claude.json` is empty/missing (custom mount, manual `rm`, `CLAUDE_CONFIG_DIR` pointing somewhere fresh). The agent can recover the wizard-hang state without a rebuild.

## Tests

`src/init/dockerfile.test.ts` gains 6 tests in the existing `claudeCode toggle` describe block:

- pre-seed presence (`hasCompletedOnboarding:true` + write target)
- theme default (`theme:"dark"`)
- ordering vs smoke test (`claude --version` must run first)
- absence of trust/permission flags (regression guard against silently widening defaults)
- off-toggle no-op (`claudeCode: false` → no `.claude.json` line in the Dockerfile)
- versioned-form drift guard (GHCR-base users see the seed too)

All 4582 tests in the suite pass. Typecheck clean, lint clean (only pre-existing warnings unrelated to changed files), format clean.

## Files

- `src/init/dockerfile.ts` — extract `CLAUDE_CODE_ONBOARDING_SEED`, append the printf
- `src/init/dockerfile.test.ts` — 6 new tests
- `src/skills/typeclaw-claude-code/SKILL.md` — document the pre-seed and self-heal in Prerequisites


devxoul:
## Self-review follow-up (commit 50a7714)

Ran two parallel reviewers (Oracle for critical reasoning, Explore for QA execution) against the original commit. They surfaced three real issues; this commit addresses all of them.

### 1. Blocker — the pre-seed alone does NOT fully unblock the first launch

Original PR claimed the seed fixes the wizard hang. **Partially true**: it clears the theme picker, but Claude Code still lands on two pre-prompt modals that this PR's seed doesn't touch:

- **Custom API key confirmation** — *"Detected a custom API key from environment. Do you want to use this API key?"* (default: **No**). Fires when `ANTHROPIC_API_KEY` is set — which is exactly typeclaw's auth path, so it fires on every cold start.
- **Workspace trust dialog** — *"Do you trust the files in this folder?"* Fires on every fresh cwd, so every `/tmp/cc-<id>/` worktree triggers it.

The skill's `send-keys "<prompt>" Enter` would still be eaten by these modals, and the `Stop` hook would still never fire.

**Why not pre-seed these too?** Pre-seeding `hasTrustDialogAccepted` or `customApiKeyResponses.approved` in the Dockerfile would silently widen the trust surface beyond what the operator consented to. The dockerfile.ts comment now documents this scope decision explicitly.

**Fix:** Handle them at runtime in `src/skills/typeclaw-claude-code/SKILL.md`. New step 4 in "Driving the session" — *"Clear startup dialogs"*:
- Capture pane after init wait, look for each modal
- `Up Enter` to flip selection to **Yes** for each
- Explicit safety note: accepting trust on a freshly-created `/tmp/cc-<id>/` worktree is fine (files were just committed by the user), but this does NOT extend to in-session permission prompts (Bash, Edit, etc.) which still need per-turn judgment

### 2. Important — ordering rationale was empirically false

Original comment claimed `claude --version` creates `~/.claude.json` and would clobber our seed if run after it. Oracle ran `claude --version` against a fresh `HOME` empirically: **it does not touch `~/.claude.json`.**

Rewrote the rationale honestly: the seed write runs **last** so the final layer state is exactly the seeded config — independent of whether any current or future Claude command writes a default state partway through. Same observable behavior, honest reasoning. Renamed the test.

### 3. Important — seed was a hand-written string literal

Quote/escape bugs would have only surfaced at `docker build` time, after passing every existing test. Switched to `JSON.stringify(...)`:

```ts
const CLAUDE_CODE_ONBOARDING_SEED = JSON.stringify({
  hasCompletedOnboarding: true,
  theme: 'dark',
  installMethod: 'native',
  numStartups: 1,
})
```

Added two regression tests:
- Extract the `printf` argument from the rendered Dockerfile and `JSON.parse` it (verifies parseability + key presence)
- Assert the JSON contains no single quotes (verifies the `printf '%s\n' '<JSON>'` shell-quoting invariant survives any future seed extension)

### 4. Minor — `typeclaw-config` SKILL.md didn't mention the pre-seed

Added a one-clause note to the `claudeCode` toggle table row pointing readers to the skill for the dialog-clearing protocol.

### Items deliberately NOT addressed

- **`lastOnboardingVersion`**: Oracle empirically confirmed `hasCompletedOnboarding: true` alone is honored by Claude Code 2.1.146 without a version pin. ii-agent's `"1.0.30"` is not good evidence for current behavior. Hardcoding a stale value would create version-coupled drift. Documented as an intentional omission in dockerfile.ts; if a future Claude release re-triggers the picker when the field is missing, the fix is to capture `claude --version` output at build time, not to pin a stale value now.
- **`installMethod` / `numStartups`**: Kept. Both match the shape Claude Code itself writes after a clean first launch, so our seed is indistinguishable from real post-onboarding state.

### Verification

- 4584 tests pass (was 4582; +2 new)
- Typecheck / lint / format clean on changed files

### Files

- `src/init/dockerfile.ts` (+82 -33) — `JSON.stringify` seed, scope-boundary comment, honest ordering rationale
- `src/init/dockerfile.test.ts` (+18 -2) — 2 new tests, 1 renamed
- `src/skills/typeclaw-claude-code/SKILL.md` (+12 -8) — new "Clear startup dialogs" step, honest scope note in prerequisites
- `src/skills/typeclaw-config/SKILL.md` (+11 -11) — table row mentions pre-seed

devxoul:
## Second self-review pass (commit a7370f2, force-pushed)

Re-ran Oracle + Explore against the post-follow-up state. Oracle (with reverse-engineered Claude Code source in hand) caught that the dialog-clearing keystrokes I added in the first self-review were **empirically wrong**. Fixed in this commit, plus rebased onto current `main`.

### Blocker fix — keystroke correctness

Claude Code's select widget (the Ink-based `Q6` component) does **NOT wrap**: pressing `Up` from the first option is a no-op. The prior `Up Enter` guidance therefore submitted **the wrong answer**, and worse on the API-key dialog it could persist rejection state.

| Dialog | Options (order) | Default focus | Correct keystrokes | Why |
|---|---|---|---|---|
| Custom API key | `[No (recommended), Yes]` | **No** | **`Down Enter`** | Need to advance forward to Yes; `Up Enter` would stay on No and could persist the key in `customApiKeyResponses.rejected`, breaking subsequent launches |
| Workspace trust | `[Yes, proceed, No, exit]` | **Yes** | **bare `Enter`** | Default focus is already on Yes — no arrow key needed |

Evidence: [Yuyz0112/claude-code-reverse](https://github.com/Yuyz0112/claude-code-reverse) — option arrays and the `focus-previous-option` non-wrapping handler are both in `merged-chunks/claude-code-*.mjs`.

### Important fix — step 4 is now a loop, not a one-shot

The prior version captured the pane once, cleared at most one dialog, then waited 2 seconds and sent the task prompt. Cold-start and slow-disk cases can deliver a dialog **after** that cushion, and sending the task prompt into a modal corrupts the session.

New step 4 is an explicit state-machine loop:
```
1. capture-pane (last ~15 lines)
2. if API-key dialog matches  → send-keys Down Enter; sleep 500ms; goto 1
3. if trust dialog matches    → send-keys Enter; sleep 500ms; goto 1
4. if input box visible (`╭` border, no dialog text) → exit loop
5. else                       → sleep 500ms; goto 1
   (10s wall-clock budget; surface to user on overshoot)
```

This also handles the chained case where clearing one dialog reveals another.

### Important fix — softened trust-safety rationale

The prior phrasing "the worktree at `/tmp/cc-<id>/` is a fresh checkout of files the user just committed, so trusting it is equivalent to trusting the user's own repo" overstates the guarantee. The worktree picks up `HEAD`, but `HEAD` may include:
- a recently pulled third-party PR
- an untrusted ref the user checked out
- dependency code the user hasn't audited

Rewritten guidance: auto-accept trust **only when you're confident `HEAD` is one the user authored or reviewed**; otherwise surface the trust dialog to them. This is consistent with the dockerfile.ts decision to NOT pre-seed `hasTrustDialogAccepted`.

### Nit fix — single-quote regression test now fails loud

The "contains no single quotes" test would have silently passed on `payload = ''` if the printf-arg regex ever failed to match. Now asserts `match` isn't null and `payload` is defined before the substring check.

### Rebase onto current main

The PR was 4 commits behind `main`. The previous review pass's diff-against-main showed apparent reverts of `secrets.json`/`.env` hierarchy edits in `typeclaw-config/SKILL.md` (introduced by commit `c6887b0` "prompts: lead with secrets.json as the canonical credential store"). That was the rebase staleness, not actual changes from this PR. Force-pushed-with-lease after a clean rebase; the typeclaw-config diff is now exactly the one-line note on the `claudeCode` toggle row this PR intends.

### Verification

- 4591 tests pass / 0 fail (suite grew with the rebase)
- Typecheck / lint / format clean
- Generated Dockerfile output unchanged (the keystroke fix is pure docs; only rationale comments in dockerfile.ts changed to align with the corrected option orders)

### Files changed in this commit

- `src/skills/typeclaw-claude-code/SKILL.md` (+18 -8) — corrected keystrokes, loop shape, softened trust safety, aligned prerequisite bullet
- `src/init/dockerfile.ts` (+8 -6) — aligned the SCOPE rationale comment with the corrected option orders
- `src/init/dockerfile.test.ts` (+3 -1) — single-quote test now asserts regex matched

### What this PR now is, post-rebase

Three commits on top of `origin/main` (`ada9f39`):
- `38ecf3f` (was `e9c4edc`) — original pre-seed
- `7c796b7` (was `50a7714`) — first self-review (JSON.stringify, ordering, dialog handling)
- `a7370f2` — this commit, second self-review (correct keystrokes, loop, softened trust, test nit, rebase)

PR #299, opened 2026-05-22
discord-bot: add /stop as a native slash command via INTERACTION_CREATE

devxoul:
## Summary

Text-prefix `/stop` already works across every channel adapter today — the router intercepts it and aborts the in-flight turn. This PR adds Discord's native application-command surface on top of that: `/stop` now appears in the slash-command picker with autocomplete and a description tooltip. Invoking it from the picker ends the turn the same way typing `/stop` as a message already does. The text-prefix path is unchanged and continues to work during Discord's command propagation window (documented up to one hour for new global commands, so there is no user-visible regression during rollout).

### `channels: ChannelRouter.executeCommand` (commit 1 of 2)

A new entry point on `ChannelRouter` lets adapters fire registered channel commands without going through the inbound classifier, engagement gate, debounce, or prompt queue. It looks up the live session by `ChannelKey` and invokes the same command registry the text-prefix `/stop` path already uses, so a single handler powers both surfaces.

Returns three outcomes:

- `handled` — live session existed and the command ran
- `no-live-session` — cold channel; adapter surfaces a user-visible "nothing to stop" reply
- `unknown-command` — defensive fallback; should not occur if the adapter only registers names it knows the router exposes

`ChannelCommandContext.event` is widened from `InboundMessage` to `InboundMessage | null` so handlers invoked from a native command surface (no underlying inbound message) receive a clean null rather than a synthesized fake. Existing handlers never read the event field; future ones that do must guard on null.

The four test-file changes outside `router.test.ts` are mechanical stub additions on inline `ChannelRouter` fixtures — required to keep every commit individually buildable.

### `discord-bot: slash command registration and INTERACTION_CREATE routing` (commit 2 of 2)

New file `discord-bot-slash-commands.ts` holds the pure helpers:

- `registerCommands` — bulk PUT to `/applications/{id}/commands`. Idempotent by design (Discord server-side overwrites with the body), so repeat starts cost nothing.
- `parseInteractionAsCommand` — maps a `DiscordGatewayInteractionEvent` to a `ChannelKey` using the same `@dm` convention and thread-as-channel shape the message-event classifier already uses.
- `ackInteraction` — POSTs `CHANNEL_MESSAGE_WITH_SOURCE` ephemerally. Does not forward the bot token because interaction tokens self-authenticate and Discord 401s the callback when both are present.

`discord-bot.ts` wires it in:

- On `connected`, fire-and-forget `registerCommands` using `info.user.id` (for bots, gateway `user_id == application_id`). A slow or failing API call does not block the listener from processing messages.
- `listener.on(interaction_create)` pumps through `createInteractionHandler`, extracted as a standalone factory so integration tests can target it without a live gateway — same shape as the other `createXCallback` helpers in this file.

## Permissions note

`executeCommand` is intentionally not gated on `channel.respond`. The `/stop` command's natural protection is identical to the text-prefix path: if there is no live session, `executeCommand` returns `no-live-session` and the adapter replies "nothing to stop." Any user who can speak in the channel can already type `/stop` as a message; the slash-command surface adds discoverability, not capability. Gating it on `channel.respond` would mean a user who lacks respond permission cannot stop a session they did not start but that is nonetheless running in response to someone else's message — which is the wrong failure mode for an abort command.

## Reviewer notes

**Command propagation.** Discord's global application commands take up to one hour to appear in the picker for existing clients after first registration. Text-prefix `/stop` works throughout, so the delay is invisible to users.

**Registration failure handling.** A `403` from the registration call almost always means the OAuth invite URL is missing the `applications.commands` scope. The log line emitted in that case includes this hint inline so operators do not need to search the docs. Registration is fire-and-forget on `connected`, so failure does not block message processing.

**Slack.** Slack also delivers `slash_commands` over Socket Mode and would benefit from the same upgrade. It is intentionally not in this PR: the envelope, ack budget, and response shape differ enough that a parallel half-built path would be worse than two focused changes. That is left as explicit future work.

**Lint warnings.** 18 total, all pre-existing on `origin/main`. They include unused functions `permissiveConfig`/`permissive` in `slack-bot.test.ts` and `discord-bot.test.ts`, and an unused `init` param in `github/line-comment-thread.test.ts`. Line numbers in `discord-bot.test.ts` shifted by +4 due to one added import — the underlying warning is pre-existing.

## Testing

| Gate | Result |
|---|---|
| `bun run lint` | 18 warnings (all pre-existing on `origin/main`), 0 errors |
| `bun run format` | clean |
| `bun run test` | 4628 pass, 0 fail |

New tests:

- `discord-bot-slash-commands.test.ts` — 15 unit tests on the three pure helpers
- `router.test.ts` — 5 tests on `executeCommand` (handled / no-live-session / unknown-command paths and ChannelKey lookup)
- `discord-bot.test.ts` — 7 integration tests on `createInteractionHandler` with a fetch mock, covering: handled / no-live-session / non-CHAT_INPUT drop / unknown-command drop / DM fallback / ack failure non-fatal / executeCommand exception caught

### DIFF
--- a/src/agent/tools/channel-reply.test.ts
+++ b/src/agent/tools/channel-reply.test.ts
@@ -37,6 +37,7 @@ function fakeRouter(
     getSelfAliases: () => [],
     stop: async () => {},
     liveCount: () => 0,
+    executeCommand: async () => ({ kind: 'no-live-session' }),
   }
 }
 
--- a/src/agent/tools/channel-send.test.ts
+++ b/src/agent/tools/channel-send.test.ts
@@ -31,6 +31,7 @@ function fakeRouter(
     getSelfAliases: () => [],
     stop: async () => {},
     liveCount: () => 0,
+    executeCommand: async () => ({ kind: 'no-live-session' }),
   }
 }
 
--- a/src/agent/tools/channel-tool-logging.test.ts
+++ b/src/agent/tools/channel-tool-logging.test.ts
@@ -53,6 +53,7 @@ function fakeRouter(overrides: RouterOverrides = {}): ChannelRouter {
     getSelfAliases: () => [],
     stop: async () => {},
     liveCount: () => 0,
+    executeCommand: async () => ({ kind: 'no-live-session' }),
   }
 }
 
--- a/src/channels/adapters/discord-bot-slash-commands.test.ts
+++ b/src/channels/adapters/discord-bot-slash-commands.test.ts
@@ -1,0 +1,242 @@
+import { describe, expect, test } from 'bun:test'
+
+import type { DiscordGatewayInteractionEvent } from 'agent-messenger/discordbot'
+
+import {
+  ackInteraction,
+  buildInteractionAck,
+  DISCORD_SLASH_COMMAND_TYPE_CHAT_INPUT,
+  parseInteractionAsCommand,
+  registerCommands,
+  synthesizeCommandText,
+} from './discord-bot-slash-commands'
+
+function makeInteraction(overrides: Partial<DiscordGatewayInteractionEvent> = {}): DiscordGatewayInteractionEvent {
+  return {
+    type: 'INTERACTION_CREATE',
+    id: 'i-1',
+    application_id: 'app-1',
+    token: 'tok-abc',
+    channel_id: 'c1',
+    guild_id: 'g1',
+    member: { user: { id: 'u-alice' } },
+    data: { name: 'stop', type: DISCORD_SLASH_COMMAND_TYPE_CHAT_INPUT },
+    ...overrides,
+  }
+}
+
+describe('parseInteractionAsCommand', () => {
+  const known = new Set(['stop'])
+
+  test('parses a guild CHAT_INPUT /stop into a discord-bot ChannelKey', () => {
+    const result = parseInteractionAsCommand(makeInteraction(), known)
+    expect(result.kind).toBe('parsed')
+    if (result.kind !== 'parsed') return
+    expect(result.command).toEqual({
+      name: 'stop',
+      key: { adapter: 'discord-bot', workspace: 'g1', chat: 'c1', thread: null },
+      invokerId: 'u-alice',
+      interactionId: 'i-1',
+      interactionToken: 'tok-abc',
+    })
+  })
+
+  test('DM interaction (no guild_id) maps workspace to @dm and reads user.id', () => {
+    const result = parseInteractionAsCommand(
+      makeInteraction({ guild_id: undefined, member: undefined, user: { id: 'u-bob', username: 'bob' } }),
+      known,
+    )
+    expect(result.kind).toBe('parsed')
+    if (result.kind !== 'parsed') return
+    expect(result.command.key.workspace).toBe('@dm')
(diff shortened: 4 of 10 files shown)


devxoul:
## Self-review (Oracle) follow-up

Force-pushed amendments after an adversarial self-review surfaced two blocking issues. Both fixed in this PR; the parent commits on the branch were amended rather than adding fix-up commits, so the two-commit shape is preserved.

### Blocking fixes

**A1: native /stop bypassed channel.respond permission gating.**
`parseInteractionAsCommand` was extracting `invokerId` but `executeCommand(key, name)` was dropping it — so a Discord user permitted to invoke the slash command could abort an owner-created session even if their role would deny `channel.respond` for text. The natural protection ("no live session = no permission") only works for *creating* a session, not interacting with an existing one.

Fix: extended `executeCommand` to `(key, name, { invokerId })`, returns new `permission-denied` result kind. Gate runs BEFORE the live-session lookup so a denied invoker cannot probe session presence via the result. Adapter forwards `command.invokerId` and surfaces a user-visible permission-denied reply.

**A2: formatChannelTag could blow Discord's 3s ack budget.**
The pre-fix order was `parse → formatChannelTag [HTTP] → executeCommand → ack`. `formatChannelTag` calls the Discord channel resolver which can do up to two REST calls on a cache miss. On a slow Discord API minute the ack fires past 3s and the user sees "This interaction failed" even though the abort worked.

Fix: ack-first restructure. Pre-ack work is bounded to in-memory operations (parse + bare-id log + executeCommand, which is a map lookup + abort). `formatChannelTag` runs AFTER ack for decorative post-ack logging. New test holds `formatChannelTag` on a deferred promise and proves the ack fires before the tag resolves.

### Non-blocking fixes

- **B4 — URL encoding**: `encodeURIComponent` on `applicationId`, `interactionId`, `interactionToken` even though Discord IDs are URL-safe by spec. Defense in depth; new test with a token containing reserved chars.
- **B5 — Error sanitization**: `ackInteraction` no longer returns the raw fetch error message — some Node runtimes embed the full URL (including the interaction token, which is a short-lived credential) in `Error.message`. Now returns `"network error: <ErrorClass>"`. New test verifies the token never appears in the error string.

### Test/lint/format

- 4645 tests pass (was 4628 pre-fix; +17 net, all on the new paths)
- `bun run typecheck` clean
- `bun run lint`: 18 warnings (all pre-existing on main), 0 errors
- `bun run format` clean

### What was NOT changed

- **B6 (adapter-wiring lifecycle test)**: deferred. The existing test file pattern across all five channel adapters deliberately does not exercise `createXAdapter()` directly — would require fake `DiscordBotClient`/`SlackBotListener` stubs. Adding it here-only would be inconsistent; better to land a unified adapter test harness in a follow-up that covers all five.
- Bulk-overwrite `PUT /commands` still owns the application's global command set. Added an inline CAUTION comment documenting the operator constraint: do not share the bot application with another integration that also registers global commands.

Sister PR #301 (Slack) gets the same A1 fix via rebase, plus A3 (stop-drain race), A4 (ambiguity rejection), B1/B2/B3 (ack contract + drift test) which are Slack-specific.

PR #301, opened 2026-05-22
slack-bot: add /stop as a native Slack slash command via slash_commands

devxoul:
## What and why

PR #299 adds Discord's native `/stop` slash command via `INTERACTION_CREATE` and introduces `ChannelRouter.executeCommand` as the generic entry point for native slash invocations. This PR stacks on top of it and adds the equivalent for Slack.

After this lands, operators and their users see `/stop` in Slack's compose-box picker with a description tooltip. Invoking it ends the in-flight turn the same way typing `/stop` as text already does. The existing text-prefix path is unchanged.

## Why this is a separate PR from #299

Slack and Discord diverge sharply on slash-command registration:

- Discord has a single HTTP endpoint (`PUT /applications/{id}/commands`) that typeclaw calls at adapter start with the bot token. Zero operator action required.
- Slack's `apps.manifest.update` API requires a separate human-generated **configuration token** (`xoxe.xoxp-…` prefix) that must be rotated every 12h via `tooling.tokens.rotate`. Even after a successful manifest update, the workspace admin must re-install the app to grant the bot token the new `commands` scope. Auto-registration is strictly worse operator UX than the manual path.

typeclaw already ships `SLACK_APP_MANIFEST` in `src/cli/ui.ts` — a JSON template operators copy-paste into api.slack.com/apps during `typeclaw init` / `typeclaw channel add slack-bot`. The right approach is to bake `/stop` into that template so new operators get the picker UX for free, and to document the manual upgrade path for existing operators.

On the listener side, `agent-messenger/slackbot` exposes `slash_commands` as a typed event with `SlackSocketModeSlashCommandArgs` over the existing Socket Mode WebSocket — no new transport needed.

### 1 — `channels: fall back to any live session in same workspace+chat when executeCommand key misses`

Slack slash commands are invoked from the compose box and never carry a `thread_ts`. But a Slack session that engaged via `@mention` in a channel is keyed on `thread: event.ts` (the root message timestamp). Without a fix, a slash `/stop` from a channel with an active thread-keyed session would miss the exact-key lookup in `ChannelRouter.executeCommand` and falsely report `no-live-session`.

New helper `resolveLiveSessionForCommand` keeps the exact-match path identical and only walks `liveSessions` on a miss, matching by `(adapter, workspace, chat)`. Map iteration is insertion-order in JS, so the first non-destroyed candidate wins — the oldest matching session, which is the one most likely to be "whatever is running here."

Behavior on Discord is **unchanged**: Discord treats threads as channels, so the exact-match path always succeeds and the fallback is a no-op.

Four new router tests cover: fallback fires when only a thread-keyed session exists; exact match wins when both exact and fallback apply; fallback ignores sessions in other chats (same workspace); fallback ignores sessions in other workspaces (same chat id — defensive because chat ids are not unique across workspaces).

### 2 — `slack-bot: add pure helpers for parsing slash command payloads`

New module `slack-bot-slash-commands.ts` mirrors the `discord-bot-slash-commands` shape:

- `parseSlashCommand` turns a `SlackSocketModeSlashCommandArgs` body into either a `ParsedSlackSlashCommand` (with the destination `ChannelKey`, invoker user id, and lowercased command name) or an ignore-with-reason.
- `buildSlashAckPayload` constructs an ephemeral response payload so control gestures stay out of the channel transcript.

Two non-obvious conventions inherited from `slack-bot-classify`:

- DM channels (`D`-prefixed channel ids) → `workspace: '@dm'`
- `thread` is always `null` — slash commands are compose-box-only and never carry `thread_ts`

Pure: no I/O, no side effects, no Slack SDK calls. The pure-first split keeps the next commit focused on adapter integration and manifest.

### 3 — `slack-bot: register /stop in the App Manifest and route slash_commands through executeCommand`

**Manifest changes** (`src/cli/ui.ts`):

- `features.slash_commands` gains a `/stop` entry with description `"Stop the agent's current turn"`.
- `oauth_config.scopes.bot` gains the `commands` scope.
- The slash command `url` is set to `https://example.invalid/typeclaw-uses-socket-mode` — RFC 6761 guaranteed-never-routable. Slack requires the field even in Socket Mode but ignores it at runtime; pointing it at a real host would silently route invocations there if the app were ever switched out of Socket Mode.

**Adapter wiring** (`src/channels/adapters/slack-bot.ts`):

- New `createSlashCommandHandler` factory (same shape as the other `createX` helpers) is extracted standalone for testability.
- The handler owns the `ack()` call exclusively — unlike `message` / `app_mention` which ack inline — because the ack payload carries the user-visible reply text. Slack's 3-second ack budget is comfortably met by the synchronous `executeCommand` path.
- DM detection uses the `D`-prefix channel id convention (slash payloads don't carry `channel_type`), matching the existing `slack-bot-classify` convention.

**Upgrade impact for existing operators**: must update their Slack app manifest to add the `/stop` slash command and the `commands` scope, then re-install the app to the workspace. New operators get this automatically from the updated `SLACK_APP_MANIFEST` template.

## Files changed

```
src/channels/adapters/slack-bot-slash-commands.test.ts   new  97 lines
src/channels/adapters/slack-bot-slash-commands.ts        new  78 lines
src/channels/adapters/slack-bot.test.ts                 +186 lines
src/channels/adapters/slack-bot.ts                       +93 lines
src/channels/router.test.ts                              +65 lines
src/channels/router.ts                                   +42 lines
src/cli/ui.test.ts                                       +17 lines
src/cli/ui.ts                                            +22 lines
8 files changed, 595 insertions(+), 5 deletions(-)
```

## Testing

| Gate | Result |
|---|---|
| `bun run lint` | 18 warnings (all pre-existing on `origin/main`), 0 errors |
| `bun run format` | clean |
| `bun run test` | 4652 pass, 0 fail (+24 vs the #299 base at 4628) |

Test breakdown for the 24 new tests:
- 4 router fallback tests (commit 1, `router.test.ts`)
- 11 pure helper tests (commit 2, `slack-bot-slash-commands.test.ts`)
- 3 manifest assertion tests (commit 3, `ui.test.ts`)
- 6 adapter integration tests (commit 3, `slack-bot.test.ts`)

## Review notes

- **Ack ownership**: the slash handler owns `ack()` exclusively (unlike `message` / `app_mention` which ack inline) because the ack payload carries the user-visible reply text. This is the correct Slack Socket Mode contract.
- **`example.invalid` in the manifest**: RFC 6761 guarantees this domain never resolves. It is not a placeholder to replace — it is the correct permanent value for Socket Mode apps.
- **Fallback scope**: `resolveLiveSessionForCommand` matches on `(adapter, workspace, chat)` only — it does not match across workspaces or across adapters. The defensive test for same-chat-id-different-workspace confirms the isolation.
- **No change to the text-prefix `/stop` path**: the `router.ts` message handler and `slack-bot-classify` are untouched.

Builds on #299 (`discord-bot: add /stop as a native slash command via INTERACTION_CREATE`).

### DIFF
--- a/src/channels/adapters/slack-bot-slash-commands.test.ts
+++ b/src/channels/adapters/slack-bot-slash-commands.test.ts
@@ -1,0 +1,97 @@
+import { describe, expect, test } from 'bun:test'
+
+import type { SlackSocketModeSlashCommandArgs } from 'agent-messenger/slackbot'
+
+import { buildSlashAckPayload, parseSlashCommand } from './slack-bot-slash-commands'
+
+type Body = SlackSocketModeSlashCommandArgs['body']
+
+function body(over: Partial<Body> = {}): Body {
+  return {
+    command: '/stop',
+    text: '',
+    user_id: 'U-alice',
+    channel_id: 'C-general',
+    team_id: 'T-acme',
+    ...over,
+  }
+}
+
+describe('parseSlashCommand', () => {
+  const known = new Set(['stop'])
+
+  test('parses /stop in a public channel into a slack-bot ChannelKey', () => {
+    const result = parseSlashCommand(body(), known)
+    expect(result.kind).toBe('parsed')
+    if (result.kind !== 'parsed') return
+    expect(result.command).toEqual({
+      name: 'stop',
+      key: { adapter: 'slack-bot', workspace: 'T-acme', chat: 'C-general', thread: null },
+      invokerId: 'U-alice',
+    })
+  })
+
+  test('maps DM channel ids (D prefix) to workspace=@dm', () => {
+    const result = parseSlashCommand(body({ channel_id: 'D-bob' }), known)
+    expect(result.kind).toBe('parsed')
+    if (result.kind !== 'parsed') return
+    expect(result.command.key.workspace).toBe('@dm')
+    expect(result.command.key.chat).toBe('D-bob')
+  })
+
+  test('private channel ids (G prefix) map to the team workspace (NOT @dm)', () => {
+    const result = parseSlashCommand(body({ channel_id: 'G-secret' }), known)
+    expect(result.kind).toBe('parsed')
+    if (result.kind !== 'parsed') return
+    expect(result.command.key.workspace).toBe('T-acme')
+    expect(result.command.key.chat).toBe('G-secret')
+  })
+
+  test('lowercases the command name', () => {
+    const result = parseSlashCommand(body({ command: '/STOP' }), known)
+    expect(result.kind).toBe('parsed')
+    if (result.kind !== 'parsed') return
+    expect(result.command.name).toBe('stop')
+  })
+
+  test('ignores unknown command names', () => {
+    const result = parseSlashCommand(body({ command: '/unknown' }), known)
+    expect(result).toEqual({ kind: 'ignore', reason: 'unknown-command' })
+  })
+
+  test('ignores malformed commands (no leading slash)', () => {
+    const result = parseSlashCommand(body({ command: 'stop' }), known)
+    expect(result).toEqual({ kind: 'ignore', reason: 'malformed' })
+  })
+
+  test('ignores payloads missing user_id (defensive)', () => {
+    const result = parseSlashCommand(body({ user_id: '' }), known)
+    expect(result).toEqual({ kind: 'ignore', reason: 'no-invoker' })
+  })
+
+  test('ignores payloads missing channel_id', () => {
+    const result = parseSlashCommand(body({ channel_id: '' }), known)
+    expect(result).toEqual({ kind: 'ignore', reason: 'no-channel' })
+  })
+
+  test('ignores payloads missing team_id (defensive — slash commands always carry it)', () => {
+    const result = parseSlashCommand(body({ team_id: '' }), known)
+    expect(result).toEqual({ kind: 'ignore', reason: 'no-team' })
+  })
+
+  test('always sets thread:null (Slack slash commands cannot be invoked from a thread)', () => {
+    const result = parseSlashCommand(body(), known)
+    expect(result.kind).toBe('parsed')
+    if (result.kind !== 'parsed') return
+    expect(result.command.key.thread).toBe(null)
+  })
+})
+
+describe('buildSlashAckPayload', () => {
+  test('returns an ephemeral response payload', () => {
+    expect(buildSlashAckPayload('Stopped.')).toEqual({
+      response_type: 'ephemeral',
+      text: 'Stopped.',
+    })
+  })
+})
--- a/src/channels/adapters/slack-bot-slash-commands.ts
+++ b/src/channels/adapters/slack-bot-slash-commands.ts
@@ -1,0 +1,82 @@
+import type { SlackSocketModeSlashCommandArgs } from 'agent-messenger/slackbot'
+
+import type { ChannelKey } from '@/channels/types'
+
+// Slack channel ids: 'C' = public, 'G' = private/legacy multi-party DM,
+// 'D' = direct message. Slash-command payloads don't carry `channel_type`,
+// so we read the id prefix directly. The slack-bot inbound classifier uses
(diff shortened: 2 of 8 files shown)


devxoul:
## Self-review (Oracle) follow-up

Force-pushed amendments after an adversarial self-review surfaced four blocking issues. All fixed in this PR; previous three commits were dropped and rebuilt cleanly on top of the updated parent PR #299 (which itself received A1+A2 fixes — see its comment thread for details).

### Blocking fixes

**A1: native /stop bypassed channel.respond permission gating.**
Inherited from PR #299's fix. `executeCommand` now requires `{ invokerId }`; the Slack adapter forwards `command.invokerId` from the slash payload and surfaces a `permission-denied` reply when the invoker fails the gate. A guest Slack user invoking `/stop` on a public channel can no longer abort an owner-created session.

**A2: formatChannelTag could blow Slack's 3s ack budget.**
Same shape as Discord — `formatChannelTag` calls the Slack channel resolver (Slack Web API) and can blow the 3s ack deadline on a slow API minute. The user sees "/stop didn't respond in time" even though the abort worked.

Fix: ack-first restructure. Pre-ack work is bounded; `formatChannelTag` runs AFTER ack for decorative logging. New test holds the resolver on a deferred promise and proves ack fires first.

**A3: slash handler not counted in `inflightInbounds` (stop-drain race).**
The Discord adapter wraps `handleInteractionCreate` in `inflightInbounds++/--` so `stop()` drains in-flight handlers; the Slack adapter was missing the equivalent for `handleSlashCommand`. A `stop()` call during an in-flight slash could tear down the listener before the ack fired, leaving Slack hanging.

Fix: `listener.on('slash_commands', ...)` now wraps the handler call with `inflightInbounds++` and a `.finally(() => inflightInbounds--)` that wakes any `stopWaiters` — mirroring the message/app_mention pattern exactly.

**A4: fallback first-match-wins could abort an arbitrary session.**
`resolveLiveSessionForCommand` (the workspace+chat fallback for thread-mismatched slash commands) was returning the first non-destroyed candidate when multiple matched. Plausible on Slack: bot mentioned in two threads simultaneously → two thread-keyed sessions in one channel → `/stop` aborts whichever was created first, not what the user meant.

Fix: new `{ kind: 'ambiguous', matchCount }` result kind. The fallback iterates and counts; on >1 match it short-circuits with `ambiguous`. The Slack adapter surfaces a user-visible reply guiding them to `/stop` from inside the specific thread. New router test with two thread-keyed sessions verifies the rejection.

### Non-blocking fixes

- **B1 — ack exactly-once on the happy path.** Pre-fix, a thrown `ack()` on the success path entered the outer catch which called `ack` again. Now each ack site has its own try/catch; happy-path ack failures are logged but do not cascade.
- **B2 — ack-call-count assertion.** New test asserts `ackCallCount === 1` even when ack itself throws. The previous "no exception escapes" test would have passed with double-ack; this one wouldn't.
- **B3 — manifest/allow-list drift guard.** New test asserts `SLACK_APP_MANIFEST.features.slash_commands.map(c => c.command.slice(1)).sort()` equals `Array.from(SLACK_SLASH_COMMAND_NAMES).sort()`. Adding `/stop` to one but not the other will fail CI.

### Test/lint/format

- 4674 tests pass (was 4623 on the pre-fix Slack PR baseline; +51 net total across the rebase and new tests)
- `bun run typecheck` clean
- `bun run lint`: 18 warnings (all pre-existing on main), 0 errors
- `bun run format` clean

### Branch state

This PR was reset onto the new HEAD of `feat/discord-slash-stop` (PR #299) and rebuilt with three fresh commits. The original commit content was salvaged where possible; the router commit picked up the new `ambiguous` shape and the adapter commit absorbed A2/A3/B1/B2.

### Follow-up still open

- Existing operators upgrading need to re-create their Slack app from the new manifest OR manually add `/stop` under Slash Commands, AND re-install for the `commands` scope. Documented in commit 3's body.
- The `example.invalid` placeholder URL has not been smoke-tested against Slack's manifest validator. If Slack rejects RFC 6761 reserved TLDs, swap for a documented typeclaw.dev URL that returns an explanatory page.

PR #306, opened 2026-05-22
typeclaw-claude-code: pre-bake the Stop hook into the image so the operator never writes settings.json

devxoul:
## Why

PR #306 originally added documentation telling operator subagents not to write a wrong-shaped Stop hook config. Reviewer feedback correctly identified that docs are the last line of defense: the real fix is to eliminate the runtime write entirely.

Claude Code's settings parser silently ignores unknown keys and structurally-wrong values. Wrong-shape Stop hooks — using `onStop` instead of `Stop`, a bare string instead of an array, or an inner object missing `type` — produce no error at write time, no error at startup, and no error at turn end. The hook simply never fires. `.done` is never created. The polling loop runs to its wall-clock budget with no diagnostic signal. Docs cannot reliably prevent an LLM from writing a plausible-but-wrong shape; baking the correct config at build time can.

## What

5 files changed, +269 / -36:

```
src/init/dockerfile.ts                                        +116 / -1
src/init/dockerfile.test.ts                                   +136 / -0
src/skills/typeclaw-claude-code/SKILL.md                       +19 / -24
src/skills/typeclaw-claude-code/references/stop-hook.md          +2 / -1
src/skills/typeclaw-claude-code/references/tmux-driving.md       +4 / -4
```

### Hook script

The hook script ships at `/usr/local/bin/typeclaw-cc-stop-hook` (stable absolute path), written at image build time via a single-quoted POSIX heredoc. The single-quoting is load-bearing: it prevents the shell from expanding `${CLAUDE_PROJECT_DIR}` during `docker build`, preserving it verbatim for runtime expansion by Claude Code. The script reads stdin, writes `sentinel.json` atomically via temp-then-rename, then touches `.done`.

### Settings file

`$HOME/.claude/settings.json` is written at image build time using `JSON.stringify` (user-level scope). User-level fires on every `claude` invocation regardless of cwd — no per-worktree write needed. The shape is validated in `dockerfile.test.ts` via `JSON.parse` so any structural drift fails the test rather than a live delegation.

### Skill

`SKILL.md` no longer instructs the operator to construct Stop-hook JSON. It now instructs the operator to verify the global hook is wired (via `jq`) and bail to the user if it is not. The only remaining failure mode is "user has not run `typeclaw start --build` since this layer landed," which has a known remediation.

## Architecture decisions

**Why user-level and not project-level**: a project-level write (`<worktree>/.claude/settings.json`) would require some code path to place the file per worktree, re-introducing write-at-runtime. User-level fires on every `claude` invocation regardless of cwd — one write at build time, zero at runtime.

**Why `${CLAUDE_PROJECT_DIR}` and not `$PWD`**: `CLAUDE_PROJECT_DIR` is Claude Code's documented env var for the project root. It resolves to the worktree because `tmux new-session -c /tmp/cc-<id>` sets claude's cwd there, and Claude Code derives the env var from cwd. Naming it explicitly makes the hook robust to alternative invocation paths.

**Why single-quoted heredoc**: the `<<'EOF'` form is the canonical POSIX way to deliver a heredoc verbatim. Without the quotes, `${CLAUDE_PROJECT_DIR}` expands to empty during `docker build` and is baked in as a literal empty string. Every test touching the script asserts the literal variable reference survives unexpanded.

## Tests

15 new tests in `dockerfile.test.ts`:

- Script is written to the correct absolute path
- Heredoc verbatim-ness: the `${CLAUDE_PROJECT_DIR}` reference survives unexpanded
- Temp-then-rename atomicity
- Touch-after-rename ordering
- POSIX shebang (`#!/bin/sh`)
- Settings JSON shape validated via `JSON.parse`
- Absolute command path in the hook registration
- No double-quoted heredoc variant present
- Ordering: hook layer appears after the onboarding seed and before the entrypoint shim
- Omission when `claudeCode: false`
- Presence in both the inline (dev) and versioned (GHCR-base) Dockerfile forms

## Verification

| Gate | Result |
|---|---|
| `bun run lint` | 0 errors (18 pre-existing warnings) |
| `bun run format` | clean |
| `bun run test` | 4698 pass, 0 fail (+15 vs base at 4683) |


devxoul:
## Self-review caught a critical bug; force-pushed fix

A read-only critical review of the diff turned up one blocker: the hook script used `$CLAUDE_PROJECT_DIR` (Claude Code's documented env var) under the assumption it resolves to cwd. Per upstream behavior ([anthropics/claude-code#27343](https://github.com/anthropics/claude-code/issues/27343) and [#44450](https://github.com/anthropics/claude-code/issues/44450)), inside a `git worktree add`d directory `$CLAUDE_PROJECT_DIR` resolves to the **main repo's git root**, not the worktree path. The operator's flow is exactly that — `git -C /agent worktree add /tmp/cc-<id> HEAD` then `tmux new-session -c /tmp/cc-<id> claude` — so the sentinel would have landed at `/agent/sentinel.json` and `.done` at `/agent/.done`, polluting the live agent folder while the operator's polling loop watches `/tmp/cc-<id>/.done` and times out. That's the exact failure mode this PR was meant to eliminate. The bundled backup plugin's auto-commit would then carry the sentinel into git history.

### Fix

- Hook script writes to `$PWD` (the literal cwd, POSIX-required since IEEE 1003.1-1988) instead of `$CLAUDE_PROJECT_DIR`. The operator's `tmux -c /tmp/cc-<id>` sets cwd to the worktree, so `$PWD` is the worktree path.
- Added a regression test (`hook script writes to $PWD, NOT $CLAUDE_PROJECT_DIR`) that fails CI if the literal string `CLAUDE_PROJECT_DIR` ever reappears in the rendered hook script body.
- Empirically verified end-to-end with a docker build that mirrors the operator's exact setup (`git worktree add` + `cd worktree` + `sh -c '/usr/local/bin/typeclaw-cc-stop-hook'` with a Stop-event JSON on stdin): sentinel landed at the worktree path, main repo stayed clean.
- Updated all skill docs that referenced `$CLAUDE_PROJECT_DIR`.

### Two other improvements while in the area

- Settings now use **exec form** (`args: []` present) per the upstream hook docs, so Claude Code invokes the script via `execvp` directly (kernel-handled shebang, no shell tokenization). Previously the empty-`args` shape implied shell form (`sh -c "<command>"`), which worked for our absolute-path-no-special-chars case but was fragile to any future path change.
- Fixed a comment block that described the user-vs-project settings layering as "project → user override." Per [upstream docs](https://docs.claude.com/en/docs/claude-code/settings) and [anthropics/claude-code#11392](https://github.com/anthropics/claude-code/issues/11392), hooks **merge additively** across scopes (not override). Cosmetic — the runtime behavior is unchanged.

### Gates (post-fix)

- `bun run typecheck` — clean
- `bun run lint` — 0 errors
- `bun run format` — clean
- `bun test` — **4700 pass / 0 fail** (was 4683 on origin/main; +17 tests in this PR)

Force-pushed (lease-protected) to update the existing branch.

devxoul:
## Per-session filenames \u2014 concurrent claude races caught in review

Review pass turned up a real architectural issue: even after switching to `$PWD`, two concurrent claude sessions sharing the same cwd would both write to the same `sentinel.json` / `.done` and corrupt each other's state with no diagnostic. The skill explicitly forbids it ('do not share a tmux session across two delegated tasks'), but the runtime didn't enforce it \u2014 a future plugin or out-of-band caller could trigger the race silently. Fixing this structurally rather than via 'don't do that' docs.

### Fix

- **Hook script reads `session_id` from the Stop event JSON stdin** via POSIX sed (no jq/python3 dep), validates it against the 8-4-4-4-12 hex UUID regex via POSIX `case` pattern matching, then writes `${PWD}/sentinel-${sid}.json` + `${PWD}/.done-${sid}`. If extraction or validation fails (missing field, JSON-escape edge case, path traversal attempt, malformed UUID), the script falls back to `sid=malformed` \u2014 so adversarial `session_id` values can never write outside `$PWD`.
- **Operator pre-generates a UUID** via `cat /proc/sys/kernel/random/uuid` (POSIX) and passes it as `claude --session-id $cc_session_id`. The hook reads the same UUID back from stdin and writes filenames containing it. The operator's polling loop watches `.done-$cc_session_id` specifically \u2014 cleanly isolated from any other concurrent claude session in the same cwd.
- **Settings now use exec form** (`args: []` present) so Claude Code invokes the script via `execvp` directly rather than `sh -c "<command>"`. Robustness improvement; works the same for our absolute-path-no-special-chars case today but no future path change can introduce shell-tokenization bugs.

### Empirical verification

Built an alpine image with the actual rendered hook script and ran 8 end-to-end scenarios:

1. \u2705 Happy path with a valid UUID: per-session files land correctly
2. \u2705 Two concurrent sessions in the same cwd: both coexist, zero collision
3. \u2705 Path-traversal `session_id` (`../etc/passwd-evil`): falls back to `sentinel-malformed.json`, NO traversal
4. \u2705 Missing `session_id` field: falls back to malformed
5. \u2705 Empty `session_id` string: falls back to malformed
6. \u2705 Almost-valid UUID with one non-hex char: falls back to malformed (regex strict)
7. \u2705 Per-session `.done` deletion: leaves other sessions untouched
8. \u2705 Temp-file cleanup via EXIT trap: no leftover `.cc-stop-hook-in.*` files

### Tests

23 tests in `dockerfile.test.ts` (was 17 before this addition; +6 new):
- session_id extraction uses POSIX sed (not jq/python3)
- UUID-shape regex validation (8-4-4-4-12 hex with dashes)
- `sid=malformed` fallback for invalid input
- `set -eu` for fail-fast on missing PWD
- per-session filename shape (`sentinel-${sid}.json` / `.done-${sid}`)
- explicit absence of fixed-name files (`sentinel.json` / `.done` would re-introduce the race)

### Gates

- `bun run typecheck` \u2014 clean
- `bun run lint` \u2014 0 errors
- `bun run format` \u2014 clean
- `bun test` \u2014 **4705 pass / 0 fail** (was 4700)

Force-pushed (lease-protected).

devxoul:
## Second self-review caught two more critical bugs; v5 redesign with SessionStart hook

Round-2 review (this PR's third critical-bug iteration) caught two more issues. Both came from the librarian agent reading anthropics/claude-code GitHub issues:

### Critical bug: `claude --session-id <uuid>` doesn't propagate in interactive mode

Per [anthropics/claude-code#44607](https://github.com/anthropics/claude-code/issues/44607), `--session-id` works in `-p` mode but in **interactive mode** (which the skill uses), the flag sets a separate telemetry/API ID while the CLI generates its own internal UUID for the transcript and for hook payloads. v4's design \u2014 operator pre-generates a UUID, passes `--session-id $uuid`, watches `.done-$uuid` \u2014 was **non-functional**. The hook fires with claude's internal UUID, not the operator's; the polling loop watches a file that never appears.

### Major: sed-based JSON extraction has shadow-attack potential

Greedy `.*` in `sed -n 's/.*"session_id":"<chars>".*'` picks the **last** occurrence of `"session_id":"..."` in the entire JSON. Today's properly-escaped Claude output is safe (verified empirically), but if a future Claude Code field ever contains an unescaped substring matching the pattern, we extract the wrong UUID. Defense in depth needed.

### v5 fix

Pre-bake a SECOND hook script. The Claude Code install layer now writes both:

1. **`/usr/local/bin/typeclaw-cc-session-start-hook`** \u2014 fires at session start (matcher `startup|resume|clear|compact` covers all four upstream session-origin types). Reads the SessionStart event JSON, extracts `session_id`, validates it against the UUID-shape regex, and atomically writes `$PWD/.session-id` containing that UUID. Operator polls this file after spawning claude to **learn claude's actual UUID**.

2. **`/usr/local/bin/typeclaw-cc-stop-hook`** \u2014 unchanged from v4: writes per-session `$PWD/sentinel-<sid>.json` and `$PWD/.done-<sid>`.

Settings.json registers both hooks; `JSON.stringify` + regression test pins the shape.

### Defense-in-depth: PID-scoped SessionStart temp filenames

Empirical concurrent-mv test surfaced a latent race: two SessionStart hooks firing in the same cwd would both write to `.session-id.tmp` and the loser's `mv` would return non-zero (which Claude Code may surface as a hook error). PID-scoping (`.session-id.$$.tmp`) fixes it; verified inside docker with concurrent invocations.

### Empirical verification

End-to-end docker test mirroring the operator's exact setup: SessionStart payload writes `.session-id`, operator reads it, Stop payload with same UUID writes `.done-<uuid>`. Plus stress tests for malformed session_id, path-traversal attempts, missing fields, concurrent SessionStart writes (post-PID-scoping), and EXIT-trap temp-file cleanup. All pass.

### Tests

31 tests total in `dockerfile.test.ts` (was 23 in v4; +8 in a new `Claude Code global SessionStart hook` describe block):
- script path, .session-id atomicity (temp-then-rename), PID-scoped temp filenames, UUID validation (same regex as Stop hook), no `$CLAUDE_PROJECT_DIR`, omission when `claudeCode: false`, shared chmod with Stop hook
- settings.json regression test asserts both hooks are registered with correct matchers

### Gates

- `bun run typecheck` \u2014 clean
- `bun run lint` \u2014 0 errors
- `bun run format` \u2014 clean
- `bun test` \u2014 **4714 pass / 0 fail** (was 4705)

### Meta-note

This PR's failure-mode pattern was instructive: each iteration introduced a critical bug because I designed against the upstream documentation page (https://docs.claude.com/en/docs/claude-code/hooks), which says what *should* be true but is contradicted by behavior documented in GitHub issues. The librarian agent's broader search across the issue tracker caught both v2's `$CLAUDE_PROJECT_DIR` semantics issue and v4's `--session-id` propagation issue. Lesson: for any upstream-behavior assumption, search the issues, not just the docs.

Force-pushed (lease-protected).

devxoul:
## v6 — oracle's third review pass found 5 issues, all addressed; one more critical bug caught empirically during the rewrite

Oracle review identified five concrete findings; addressed all of them plus one additional critical bug I caught while empirically testing the fix. v6 is structurally different from v5 in three places:

### 1. Operator no longer pre-fetches .session-id (fixes oracle finding #1: trust-gating)

Per [anthropics/claude-code#11519](https://github.com/anthropics/claude-code/issues/11519), Claude Code SKIPS SessionStart execution while workspace trust is pending: debug logs show `Skipping SessionStart:startup hook execution - workspace trust not accepted`. Every fresh worktree starts trust-pending, so `.session-id` never appeared pre-first-prompt in our flow. v5's operator polled .session-id BEFORE clearing dialogs \u2014 deadlock on every first delegation.

**Fix**: operator clears dialogs first (including workspace trust), sends the first prompt, then globs `.done-*` to discover the session UUID from the first Stop sentinel. `.session-id` becomes a fast-path optimization for re-attached sessions.

### 2. Operator handles session_id rotation on compact (fixes oracle finding #2)

Per [anthropics/claude-code#29094](https://github.com/anthropics/claude-code/issues/29094), `SessionStart` with `source: "compact"` is a NEW session linked via `parent_session_id`. Long claude sessions that auto-compact rotate their session_id mid-delegation. v5 assumed compact reused the UUID (wrong).

**Fix**: per-turn polling watches for a new `.done-<different-uuid>` appearing. When detected, the operator updates `cc_session_id` and continues from the new UUID. Empirically verified with two simulated Stop events carrying different UUIDs.

### 3. JSON extraction via `bun -e`, not POSIX sed (fixes oracle finding #3: shadow attack)

`last_assistant_message` is a JSON-quoted string that can contain the literal text `"session_id":"<fake-uuid>"` (the assistant might have been discussing session IDs!). Greedy sed extracts the LAST occurrence, picking the fake over the real top-level field. bun's JSON.parse picks the structural `session_id` regardless.

**Fix**: both hook scripts now use `bun -e` with `Bun.file().json()` / `Bun.stdin.stream()`. Bun is guaranteed in the container (bun:1-slim base); ~15ms startup cost is negligible vs a Claude turn. Empirically verified: a Stop payload with an escaped fake session_id in `last_assistant_message` correctly extracts the REAL top-level UUID.

### 4. Stale docs fixed (oracle finding #4)

- `SKILL.md:120\u2013126` referenced old `sentinel.json`/`.done` filenames and `$CLAUDE_PROJECT_DIR` \u2014 corrected to the per-session filenames and `$PWD` semantics
- `SKILL.md:280` claimed each task needs its own `.claude/settings.json` \u2014 corrected (hooks are global; per-session filenames make shared worktrees safe except for `.session-id` which the discovery flow handles)
- `tmux-driving.md:38\u201347` described a v3 fixed-sleep approach \u2014 replaced with the dialog-polling approach the skill body actually uses, plus a note that `.session-id` cannot be used as a readiness signal (due to #11519)

### 5. Three-failure-mode diagnostics in bail path (oracle finding #5)

The verification block now distinguishes:
- Scripts missing \u2192 toggle off or image needs rebuild
- Scripts present, settings.json wrong/missing \u2192 user mounted their own ~/.claude/settings.json
- Scripts + settings correct, no sentinel ever appears \u2192 hook execution issue (trust, schema mismatch, perms); inspect `.cc-*-in.*` temp files and any `sentinel-malformed.json`

### Bonus critical bug caught during empirical testing: `process.argv[1]` vs `[2]` in bun -e

While running the docker test, the Stop hook fell back to `sid=malformed` even for valid input. Investigated: `bun -e '<code>' file.json` puts the file path at `process.argv[1]`, NOT `[2]` like Node would. Bun strips `-e <code>` from argv. v6 uses `process.argv[1]` with a regression test pinning the choice.

### Empirical verification

End-to-end docker test inside an oven/bun:1-slim container with both rendered hook scripts. Six scenarios:

1. \u2705 SessionStart fast path: writes `.session-id` when not trust-blocked
2. \u2705 SessionStart malformed JSON: falls back to `malformed`
3. \u2705 Stop happy path: writes `sentinel-<uuid>.json` + `.done-<uuid>`
4. \u2705 **Shadow attack**: Stop payload with escaped `"session_id":"<fake>"` in last_assistant_message \u2014 hook extracts the REAL top-level UUID, NOT the fake
5. \u2705 Discovery from first sentinel: operator globs `.done-*` and reads the matching sentinel
6. \u2705 Compact rotation: two Stop events with different UUIDs both produce distinct `.done` files; operator detects the rotation

### Tests

31 tests in dockerfile.test.ts. Notable additions:
- Stop hook uses `bun -e` + `Bun.file`, NOT `sed -n`
- SessionStart hook uses `bun -e` + `Bun.stdin`, NOT `sed -n`
- Stop hook uses `process.argv[1]`, NOT `[2]` (the argv bug-guard)
- Both hooks validate against the UUID-shape regex
- Both have `sid=malformed` fallback

### Gates

- `bun run typecheck` \u2014 clean
- `bun run lint` \u2014 0 errors
- `bun run format` \u2014 clean
- `bun test` \u2014 **4715 pass / 0 fail** (was 4714 in v5)

### Meta-note

This is the 4th iteration with a review finding critical bugs. The pattern continues: every iteration's docs-page reading was contradicted by GitHub-issue-level behavior, and one bug (`argv[1]` vs `[2]`) was caught only by the empirical docker test. Going forward, ANY upstream-behavior assumption for Claude Code should be checked against the issue tracker AND an empirical run, not the docs page alone.

Force-pushed (lease-protected).

devxoul:
## Fourth review pass \u2014 oracle + librarian agreed on no critical bugs; one major finding (polling-loop level-trigger) fixed

This is the first review pass on this PR that didn't find a critical bug. Track-record summary:

| Iteration | Critical bugs found | By |
|---|---|---|
| v2 | 1 (`$CLAUDE_PROJECT_DIR` semantics) | librarian (anthropics/claude-code#27343, #44450) |
| v3 | 1 (shared-cwd race) | user |
| v4 | 1 (`--session-id` doesn't propagate) | librarian (#44607) |
| v5 | 2 critical (#11519 trust-gating, #29094 compact rotation) + 3 major | oracle |
| v6 \u2192 v7 | **0 critical** | oracle + librarian |

### Oracle's verdict on v6

> 'I don't see a new critical hook-shape/runtime bug in v6.'

Oracle found:
- **1 major**: per-turn polling loop is level-triggered (waits on `.done-${cc_session_id}` specifically). If compact rotates the UUID while an old `.done-<old>` still exists, the loop never scans for the new file. **Fixed in this commit** \u2014 polling is now edge-triggered (enumerate all `.done-*`, pick newest unprocessed by mtime, track processed UUIDs explicitly).
- **1 minor**: first-turn discovery picked first glob result, not newest by mtime. **Fixed in the same rewrite** \u2014 same edge-triggered shape.

Both findings are operator-side docs only \u2014 no Dockerfile changes, no hook script changes, no risk of introducing a new critical bug.

### Librarian's empirical verification

All bun-API assumptions verified against the actual bun source code:

- `process.argv[1]` is the first positional arg under `bun -e`. Verified at [oven-sh/bun/src/runtime/node/node_process.zig#L199-L208](https://github.com/oven-sh/bun/blob/main/src/runtime/node/node_process.zig#L199-L208) \u2014 eval mode explicitly omits `vm.main` from argv.
- `Bun.stdin.stream()` + `new Response(...).json()` works on Linux.
- `Bun.file(path).json()` works, throws ENOENT on missing, SyntaxError on malformed.
- `oven/bun:1-slim` ships **1.3.14** today (floating tag).
- Two stdin bugs (#27849, #29787) both fixed before 1.3.14.

### v7 changes (this force-push)

`tmux-driving.md` polling section rewritten:
- Both phase 1 (first-turn discovery) and phase 2 (per-turn) now enumerate `ls -t .done-*` (newest by mtime first), filter against a `processed` UUID set, and pick the newest unprocessed real UUID.
- Added a "Why edge-triggered, not level-triggered" explanation block citing the two specific failure modes oracle named.
- Malformed marker handling: only honored when no real UUID file exists in the same poll.

`SKILL.md` steps 6\u20138 + multi-turn-loop bullets updated to match \u2014 explicit instructions to (a) sort by mtime, (b) track processed UUIDs, (c) `rm` only the specific file just processed (NOT a glob, which would wipe an in-flight new sentinel from concurrent compact rotation).

### Gates

- `bun run typecheck` \u2014 clean
- `bun run lint` \u2014 0 errors
- `bun run format` \u2014 clean
- `bun test` \u2014 **4715 pass / 0 fail** (unchanged from v6 \u2014 no test changes needed; polling logic lives in docs, not code)

### Merge recommendation

Ready to merge from my side. The one remaining unverified assumption (does Claude Code's interactive mode actually fire SessionStart and Stop hooks with the schema librarian researched?) can only be verified by a real delegation post-merge. The failure modes are graceful: if the assumption is wrong, the operator times out at the wall-clock budget and surfaces to the user with the three-failure-mode diagnostic block in SKILL.md.

Force-pushed (lease-protected). CI re-runs on GitHub now.

PR #314, opened 2026-05-22
memory: extend memory_search to undreamed stream events

devxoul:
## Why

`memory_search` previously only covered topic shards under `memory/topics/`. Recent fragments not yet folded into shards lived in `memory/streams/<date>.jsonl` and were searchable only by injection into the system prompt — invisible the moment the index-mode budget kicked in or the channel-origin defense forced index mode unconditionally. So when an agent was reasoning about something the dreaming subagent hadn't yet consolidated (i.e. anything that happened in the last <30 minutes, or anything dreaming was skipping due to budget pressure), it had no way to retrieve those fragments without manually `read`ing JSONL files.

This is a follow-up to PR #310 (memory sharding). #310 made shards retrievable via search; this PR extends the same retrieval to the unconsolidated half of the memory pool.

## What

Streams are now first-class to the same `memory_search` tool. Results are a discriminated union:

- `source: 'topic'` — existing shape (`shardPath`, `slug`, `heading`, `excerpt`, `fullBody?`)
- `source: 'stream'` — new shape (`streamPath`, `date`, `eventId?` in `streams/yyyy-MM-dd#<id>` citation format, `topic`, `excerpt`, `fullBody?`)

Topic matches come first (alphabetical by slug, the order `loadAllShards` returns). Stream matches follow, newest day first. `maxResults` truncates streams before topics — long-term consolidated truth survives even when recent ephemeral fragments overflow the cap. This mirrors the injection-side rendering order in `renderSection`.

### Stream-event semantics

- `fragment` matches search both `event.topic` and `event.body`; matched body lines render with the same ±3-line excerpt window as shard bodies.
- `legacy_prose` matches search `event.text`; `eventId` is omitted because pre-shard prose has no fragment id and `parseCitations` would reject any value we synthesised.
- `watermark` events are never returned — they carry no human content.
- Dreamed events (anything in `memory/.dreaming-state.json#dreamedThrough[date].dreamedIds`) are excluded; search agrees with injection on what counts as undreamed.

### Pre-refactor: shared stream-io helpers

Commit 1 (`memory: lift listStreamFiles + undreamed-event reader into stream-io`) is a pure refactor that lifts `listStreamFiles`, the dreamed-id filter, and the per-day directory walk out of `load-memory.ts` into `stream-io.ts`. New shared exports:

- `listStreamFiles(agentDir)` — directory discovery + display prefix (`memory/streams/` vs legacy `memory/`)
- `filterUndreamedEvents(events, dreamedIds)` — the per-file `legacy_prose`-aware slice that used to live in `sliceUndreamedTail`
- `readAllUndreamedStreamDays(agentDir)` — one-call helper returning undreamed events for every day, chronologically, with a `partiallyDreamed` flag callers can surface in rendered output

`load-memory.ts` now composes those instead of re-implementing them. Zero behavior change — all 555 memory-plugin tests pass with no test edits.

### LLM-facing prose surfacing

Commit 3 (`memory: surface stream-search in LLM-facing prose`) updates every string the LLM reads that points at `memory_search`:

- Tool description (`search-tool.ts`): rewritten to name both surfaces and document the result discriminator.
- `loadMemory.indexDirective` (both channel and non-channel branches): reads "specific topics or recent stream events."
- `MEMORY_RETRIEVAL_SYSTEM_PROMPT` and the subagent's initial user prompt: name both surfaces and tell the subagent results are discriminated by `source`.
- `src/skills/typeclaw-memory/SKILL.md`: documents the new dual-surface result shape, ordering rules, and truncation policy.
- Plugin README: updated for the new contract.

Without these the runtime would have the capability but the agent's mental model would still say "`memory_search` covers topic shards."

## Tradeoffs

- **One tool with a discriminated result, not two tools.** Letta/MemGPT split into `archival_memory_search` + `conversation_search`. mem0 unifies with a `scope` param. typeclaw goes unified-without-param: the agent doesn't need to know whether something has been dreamed yet, only what's relevant. A future `scope: 'topics' | 'streams' | 'both'` param can be added without breaking callers if needed.
- **Always exclude dreamed events from stream results.** Dreamed events are already in topics; finding them via stream search would be duplication. The same filter that protects injection (`filterUndreamedEvents`) now protects search.
- **No self-session filter on search.** `dropSelfSessionFragments` defends cache-suffix integrity for injection; search is explicit and the agent legitimately wants its own session's fragments back when querying.
- **`maxResults` cuts streams before topics.** Truncation policy is deliberate, not arbitrary — documented in the `searchAll` comment block.
- **Does NOT solve the unbounded stream-injection-bytes problem** (the user's original architectural concern that streams can hit ~360 KB across 30 days with no aggregate cap). This PR is the prerequisite for the cap PR: now that the agent can pull old stream events on demand via `memory_search`, a follow-up can safely cap the rendered stream-section bytes in `loadMemory` without losing access to older content.

## Tests

- 9 new stream-search cases in `search-tool.test.ts`: fragment topic match, fragment body excerpt, dreamed-id exclusion, watermark skip, legacy prose, result ordering (topics → newest streams first), truncation policy (streams cut first), full-mode `fullBody`, streams-only agent dir.
- All 12 existing shard-only cases preserved and updated for the new `source: 'topic'` discriminator.
- Test pins updated: `memory-retrieval.test.ts:49` (description substring), `load-memory.test.ts:496` (indexDirective text).
- Full repo suite: 5133 pass / 0 fail.
- Pre-commit checks green: typecheck, lint (no new errors), format:check.


devxoul:
## Self-review pass

Ran Oracle + targeted explore on the PR before requesting reviewers. Two real findings, two new commits pushed:

### Commit `5fbacd8` — fix `(undreamed tail)` label regression

The original `sliceUndreamedTail` operated on already-self-filtered events, so the "(undreamed tail)" suffix meant exactly "the visible-to-this-session slice lost events to dreaming."

The refactor in commit 1 (`a00944e`) inverted that ordering: dreamed-id filter ran first inside `readAllUndreamedStreamDays` and the label was derived from a `partiallyDreamed` flag computed against raw events. A day with dreamed-self + fresh-other events would incorrectly carry the label even though dreaming hadn't pruned anything from the visible slice. Comment I'd written at the time hand-waved this as "independent" — Oracle correctly flagged it as an unauthorized semantic change.

**Fix**: split `stream-io.ts` into two layers — `readAllStreamDays` (raw, returns dreamedIds for consumer-ordered filtering) and `readAllUndreamedStreamDays` (convenience wrapper that pre-applies the dreamed filter). Injection path uses the low-level version and orchestrates both filters in the original order. Search path keeps the convenience wrapper since it doesn't apply self-session filtering.

Two regression tests in `load-memory.test.ts` pin the asymmetry:
- dreamed-self + fresh-other → **no** `(undreamed tail)` label
- dreamed-other + fresh-other → `(undreamed tail)` label fires

### Commit `892d7fd` — direct unit coverage for stream-io helpers

The lift commit exported `listStreamFiles`, `readAllStreamDays`, `readAllUndreamedStreamDays`, and `filterUndreamedEvents` but tested them only indirectly through `loadMemory` / `dreaming`. The legacy fallback path (`memory/<date>.jsonl` at the agent-dir root, used by pre-#310 agents that never had `MEMORY.md`) had **zero** direct coverage even though it's still reachable.

Adds 14 direct tests in `stream-io.test.ts` covering both layouts, dreamed-id filtering, legacy_prose passthrough, and the `partiallyDreamed` flag. Adds 2 tests in `search-tool.test.ts` for the legacy layout under `memory_search` + a multi-day multi-event regex pin.

### Deferred to follow-ups

Three nits I'm explicitly not addressing in this PR:

1. **`dreaming.ts` still has its own private copy of `listStreamFiles`** (lines 74-114). The lift-to-stream-io commit was specifically about `load-memory.ts` consumers; folding dreaming in is a separate refactor with its own test surface.
2. **`memory_search` doesn't match `date` or `eventId` directly** — only fragment topic/body and legacy text. `memory_search({query: "2026-05-20"})` won't find events on that date unless the body mentions it. Reasonable for a follow-up; current scope is human-content search.
3. **Excerpt is line-count capped, not byte-capped.** A single-line fragment with a 50KB body returns 50KB in the excerpt. Same pre-existing behavior as shard search; out of scope here.

Final tallies: 5151 tests pass / 0 fail (up from 5133), 0 typecheck errors, 0 new lint warnings, format:check clean.

devxoul:
## Self-review round 2

Ran Oracle round-2 + a focused comment-density audit on the fix commits from round 1. No correctness blockers, three cleanup-level items addressed in commit `1502690`:

1. **`UndreamedStreamDay.partiallyDreamed` removed.** Dead data — the only consumer that needed it was the injection-side "(undreamed tail)" label, and the fix in `5fbacd8` moved that logic to use `readAllStreamDays` so the flag could be recomputed post-self-filter. Leaving the field on the convenience wrapper invites a future maintainer to use it for label derivation again and re-introduce the regression Oracle caught in round 1.

2. **`FileEntry.fullyDreamed?` removed.** Stale — the null-return refactor in `5fbacd8` already eliminated every consumer.

3. **`streamDayToStreamEntry` comment trimmed 13 → 7 lines.** Paragraph 2 duplicated the `currentSessionId` option doc; replaced with a cross-ref. Kept the load-bearing filter-ordering invariant + label semantics.

Also tightened:
- `StreamDirectory` docblock 7 → 4 lines (per comment-audit NIT).
- Two regression-test comments 5 → 1 line each (test names already carry the scenario).
- `filterUndreamedEvents` empty-set test switched from `toEqual` → `toBe` to pin the same-reference optimization production relies on (per Oracle nit).

Net: 50 lines deleted, 18 added. Same 5151 tests pass / 0 fail.

### Still deferred (called out earlier)

- `dreaming.ts` still has its own private `listStreamFiles` duplicate. Oracle confirmed this is out of scope for #314 (a senior reviewer might ask why a "lift listStreamFiles" commit didn't fold dreaming in; the honest answer is dreaming's surface needs its own audit since its consumers differ).
- `memory_search` doesn't match by date or eventId.
- Excerpt is line-capped, not byte-capped.

Will open follow-up issues if these come up in review.

PR #318, opened 2026-05-22
memory + agent: restore pre-#314 session cold-start latency

devxoul:
## Why

PR #314 (sharded memory) and the broader memory rework moved `loadMemory` from "read one `MEMORY.md` file" to "read every shard in `memory/topics/`". Every `createSession` call -- TUI opens, channel resurrections, cron ticks, subagent spawns -- now pays `T+1` disk ops per session where `T` is the topic count, with no in-memory cache. A mature agent with ~25 shards (typical 6-month accumulation; dreaming merges duplicates so the count plateaus) was reading 27 files per session vs 1 pre-#314.

Compounding this, `createResourceLoader` was awaiting three independent I/O paths sequentially: `loadSelf` (2 file reads), `renderGitNudge` (git subprocess), and `loadMemory` (N shard reads). Total cold-start latency for the three was the sum of three independent I/Os instead of the max.

This PR is split into two atomic commits, each addressing one half of the regression.

### 1. `memory: cache loadAllShards by per-shard mtime+size` (bde87a9)

Module-level cache keyed by `agentDir`, then per-slug `{mtimeMs, size, shard}`. On every `loadAllShards` call:

- `readdir(topicsDir)` to get current slug set
- For each slug, `stat(<slug>.md)` (cheap, ~tens of µs)
- Reuse the cached shard when `(mtimeMs, size)` is unchanged; otherwise `readFile`+`parseShard` and refresh the entry
- Drop entries whose files vanished, so recreate-with-same-name surfaces fresh content
- Malformed shards are cached as `null` so a hot session-create loop doesn't re-warn on every prompt

`mtimeMs + size` is sufficient as the invalidation key because every writer in the codebase produces a fresh mtime: dreaming's atomic `writeFile` to `topicShardPath`, the migration's atomic `rename` finalize, and the boot migration. The cache is **per-process**, which matches the container topology (one container = one Bun process = one agent dir at runtime).

`loadShard` (the single-slug entry point used by dreaming for post-write verification) deliberately **bypasses the cache** so on-demand callers always see disk-fresh state. The asymmetry is documented at the call site.

### 2. `agent: run loadSelf, renderGitNudge, loadMemory in parallel during session cold start` (a0ee1cd)

The three I/O paths in `createResourceLoader` are mechanically refactored from sequential `await`s into concurrent promise creation followed by `Promise.all` at the gather points. The plugin hook (`runSessionPrompt`) still waits for `loadSelf`, but `renderGitNudge` and `loadMemory` are in flight during the hook now. Total cold-start latency for these three drops from sum-of-three to max-of-three.

Error semantics preserved: both `renderGitNudge` and `loadMemory` catch their own errors internally. A true unhandled rejection in either would have crashed the old sequential code at the same await point.

## Cost analysis (per session creation)

| Scenario | Before this PR | After this PR |
|---|---|---|
| First call after container boot, T=25 shards | 1 `readdir` + 25 `readFile` + 25 `parseShard`, all sequential | 1 `readdir` + 25 `stat` + 25 `readFile` + 25 `parseShard`, overlapping with `loadSelf`+`renderGitNudge` |
| Warm cache, no shard changes (the common case) | 1 `readdir` + 25 `readFile` + 25 `parseShard` | 1 `readdir` + 25 `stat`, no readFiles, no parses |
| Warm cache, K shards changed | 1 `readdir` + 25 `readFile` + 25 `parseShard` | 1 `readdir` + 25 `stat` + K `readFile` + K `parseShard` |

For the dominant case (channel inbounds, cron ticks, repeated TUI prompts -- topic shards are stable between dreaming runs, which fire every 30 min by default), this collapses N full file reads into N cheap stat calls. Net cold-start latency on a mature agent should drop from "noticeably slow" to roughly pre-#314.

## What's NOT in this PR

- **Stat-only fast path for index-mode decision.** When `buildInjectionPlan` decides index mode (total shard bytes > budget, default 16 KB), only frontmatter + heading is rendered into the prompt -- but `loadAllShards` still reads every body. A future optimization could parse frontmatter only and skip body reads when index mode is certain. Deferred because the cache above already eliminates the per-call cost in the warm case.
- **Async-after-prompt memory injection.** Would move memory out of the system-prompt cache prefix entirely. Reshuffles the cache-suffix ordering invariant documented in AGENTS.md; out of scope.
- **LiveSessionRegistry / `/inspect` / timezone-stamp investigations.** Ruled out as cold-start contributors by parallel explore agents during the initial investigation; no changes needed.

## Tests

- `src/bundled-plugins/memory/load-shards.test.ts`: 6 new tests covering cache hit serves stale bytes (via integer-second `utimes` to dodge macOS sub-ms truncation), invalidation on mtime change, cleanup on delete+recreate, warn-once for malformed shards, per-agent-dir isolation, and `loadShard` cache bypass.
- `src/agent/index.test.ts`: 84 existing `createResourceLoader` tests cover the parallelization's correctness (no new test added -- verifying `Promise.all` is faster than sequential `await` is implementation-detail testing).
- Full suite: `5140 pass, 0 fail` across 292 files.
- `bun run typecheck && bun run lint && bun run format`: clean (0 errors; the 40 lint warnings are pre-existing in unrelated files).

### DIFF
--- a/src/agent/index.test.ts
+++ b/src/agent/index.test.ts
@@ -330,6 +330,56 @@ describe('createResourceLoader', () => {
     expect(promptSeenByHook).not.toContain('attacker-controlled-marker')
   })
 
+  test('does not surface an unhandled rejection when loadMemory throws non-ENOENT during a slow plugin hook', async () => {
+    // Regression test for the parallelization shape introduced in PR #318.
+    // gitNudge and memory promises are kicked off concurrently with loadSelf
+    // and the plugin hook. Without the settle() wrap, a non-ENOENT rejection
+    // (e.g. EISDIR from a directory masquerading as a shard) on the early-
+    // started memoryPromise would fire as `unhandledRejection` during the
+    // window between selfPromise resolving and the gather Promise.all -- a
+    // slow plugin hook widens that window arbitrarily.
+
+    // EISDIR trigger: place a directory at the path readFile expects to be
+    // a file. loadAllShards iterates memory/topics/*.md slugs, then
+    // readFile(<slug>.md). A directory at that path makes readFile reject
+    // with EISDIR (a non-ENOENT fs error).
+    await mkdir(join(agentDir, 'memory', 'topics', 'malformed.md'), { recursive: true })
+
+    const { createHookBus } = await import('@/plugin')
+    const { emptyRegistry } = await import('@/plugin/registry')
+    const hooks = createHookBus()
+    hooks.registerAll(
+      'slow-hook',
+      agentDir,
+      { info: () => {}, warn: () => {}, error: () => {} },
+      {
+        'session.prompt': async () => {
+          await new Promise((resolve) => setTimeout(resolve, 30))
+        },
+      },
+    )
+
+    const seen: unknown[] = []
+    const onUnhandled = (err: unknown) => seen.push(err)
+    process.on('unhandledRejection', onUnhandled)
+
+    try {
+      await expect(
+        createResourceLoader({
+          agentDir,
+          plugins: { registry: emptyRegistry(), hooks, sessionId: 'ses_test', agentDir },
+        }),
+      ).rejects.toThrow()
+    } finally {
+      process.off('unhandledRejection', onUnhandled)
+    }
+
+    // The error must propagate through the gather point, not as a detached
+    // unhandled rejection. Allow one microtask tick to settle.
+    await new Promise((resolve) => setTimeout(resolve, 5))
+    expect(seen).toEqual([])
+  })
+
   test('exposes the typeclaw-cron bundled skill to the agent', async () => {
     const loader = await createResourceLoader({ agentDir })
 
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -765,7 +765,22 @@ export async function createResourceLoader(options: CreateResourceLoaderOptions 
   const agentDir = options.agentDir ?? process.cwd()
   const mode: SystemPromptMode = options.mode ?? deriveSystemPromptMode(options.origin)
   const basePrompt = mode === 'slim' ? SLIM_SYSTEM_PROMPT : DEFAULT_SYSTEM_PROMPT
-  let self = await loadSelf(agentDir)
+
+  // Kick off the three independent I/O paths concurrently. Sequential awaits
+  // here used to be the dominant cold-start cost amplifier: loadSelf is 2
+  // file reads, renderGitNudge spawns a subprocess, loadMemory reads N topic
+  // shards. None of them depend on each other, so we run them in parallel.
+  // The plugin hook (runSessionPrompt) only needs `self`, so it can overlap
+  // with the gitNudge subprocess and the shard reads while `self` is in
+  // flight too.
+  const selfPromise = loadSelf(agentDir)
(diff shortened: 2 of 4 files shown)


devxoul:
## Self-review pass (Oracle)

Ran a critical second-pass review on the initial diff (Oracle consultation + my own scrub). Found and fixed two non-trivial issues, plus a handful of NITs. Two new commits on top, both passing the full test suite (`5142 pass / 0 fail`), typecheck/lint/format clean.

### BLOCKING (fixed in `443e7b3`)

`gitNudgePromise` and `memoryPromise` were created before the `runSessionPrompt` plugin hook ran. If either rejected with a non-ENOENT fs error (EACCES/EIO on the agent folder), the rejection had no handler during the window between hook-start and the gather `Promise.all`. A slow plugin hook widened that window arbitrarily, and Node/Bun would emit `unhandledRejection`.

Fix: wrap both early-started promises in `settle(...)` shells that convert `Promise<T>` to `Promise<{ ok: true; value: T } | { ok: false; error: unknown }>` — the settled promise never rejects. After the hook returns we `Promise.all` the shells and rethrow via `unwrapSettled`. Caller semantics preserved (still throws on memory/gitNudge failure), no detached rejection.

Added a regression test that creates an agent dir with `memory/topics/malformed.md` as a *directory* (forces EISDIR), installs a 30ms async hook, subscribes to `process.on('unhandledRejection')`, and asserts the error propagates through `createResourceLoader().rejects.toThrow()` without any `unhandledRejection` event firing. Reverse-checked: removing the `settle()` wrap makes the test fail.

### IMPORTANT (fixed in `f922b7f`)

Original cache invalidation key was `(mtimeMs, size)`. Sufficient for TypeClaw-owned writers (dreaming + migration both bump mtime via real `writeFile`), but NOT for metadata-preserving external edits — `rsync -t`, `touch -r`, restored backups, `git checkout` with timestamp restoration can replace content while leaving `(mtime, size)` unchanged.

Fix: added `ctimeMs` to the cache key. The kernel always bumps ctime on inode content change and ctime cannot be backdated via `utimes`, so the cache invalidates correctly even under those scenarios. One extra Number per entry, one extra equality check per cache-hit.

Also dropped the original cache-hit test's "same-mtime, same-size, different-content returns stale bytes" assertion — it was blessing the very blind spot Oracle flagged. Replaced with a reference-equality test that proves the readFile fast path positively (cached calls return the same object instance) without encoding divergent-disk-vs-cache state as expected behavior. Added a new test that pins the ctime-as-key contract against the exact rsync/touch-r case.

### Other findings handled

- Plugin-hook contract change documented in code: `runSessionPrompt` now runs AFTER gitNudge/memory I/O has been kicked off, so a hook that mutates `memory/topics/` or git-tracked files races those in-flight reads. The bundled hooks only mutate the prompt string (no disk writes), so current bundled behavior is unaffected, but third-party plugin authors need to know.
- `loadShard` cache-bypass comment was justifying the bypass with a non-existent production caller (the "dreaming subagent post-write verification" doesn't exist — dreaming uses `loadAllShards`). Replaced with an honest API-contract comment.
- `setTimeout(10ms)` real-time wait in the mtime-invalidation test replaced with deterministic `utimes` — matches the team direction in ae619e6.

### What still applies from the original PR body

The cost analysis, the two-commit split, and the deferred follow-ups (stat-only fast path, async-after-prompt memory injection) are all unchanged. The two fixup commits are purely defensive — no perf characteristic changed.

PR #322, opened 2026-05-23
init: validate API keys inline, surface role claim, advertise doctor

devxoul:
Four small UX improvements to `typeclaw init` aimed at non-developer first runs. Each one targets a friction point identified in the recent init-UX audit.

### 1. Inline API-key validation (`src/init/auth/validate.ts`, new)

After the user pastes an API key, the wizard now probes the provider's `/v1/models` endpoint (5s timeout) to detect typos, expired keys, wrong-account keys, and project-scoped keys **before** they propagate into `secrets.json`. On 401/403 the wizard renders a clear rejection note with a "Try a different key / Save anyway" branch. Flaky networks (timeout, 5xx) degrade to "couldn't verify, saving anyway" rather than blocking init.

- OAuth-only providers (`openai-codex`) skip validation entirely.
- Anthropic uses `x-api-key` + `anthropic-version: 2023-06-01`; OpenAI/Fireworks/Z.AI/Z.AI-coding use `Authorization: Bearer`.
- Goes through `WizardPrompts.validateApiKey` so tests stub it without hitting the network.

### 2. Per-provider dashboard deeplinks

Before each `password()` prompt the wizard now prints a "Get a `<Provider>` API key" note with a direct link to the provider's key-creation page. Non-devs no longer need to know what to Google to find the dashboard.

### 3. Surface `typeclaw role claim` in the success message

When init wired a chat channel (Slack/Discord/Telegram/KakaoTalk — not GitHub, which doesn't gate on roles), the post-hatch `done()` block now leads with a **`Claim your agent: typeclaw role claim`** hint and a separate boxed note explaining that the agent will silently drop every inbound message until ownership is claimed. This was the #1 "why isn't my agent answering?" trap documented in the audit.

### 4. Advertise `typeclaw doctor` in the success message

Adds a `Diagnose issues: typeclaw doctor` hint to the post-hatch summary so users hit by Docker / port / dependency drift can self-serve the diagnostic instead of digging through docs.

## Tests

- 7 new unit tests in `src/init/auth/validate.test.ts` covering provider-specific auth headers, 401 rejection, network-error degradation, OAuth-only skip, and dashboard-URL coverage.
- All 548 existing init / ui / init-pipeline tests pass.
- The 3 pre-existing `--help` / `_hostd` integration test failures reproduce on `main` and are not caused by this PR.

## Follow-ups (separate issues)

This PR ships Tier-S items from the init-UX audit. Tracking issues will be filed for Tier-A/B/C improvements: `typeclaw init --yes`, three-choice provider intro, Gemini OAuth provider, first-task demo prompt, `typeclaw cron add`, pre-flight gate before wizard ends, Slack OAuth-installer, `typeclaw init --resume`.

## Screenshots / sample UX flow

After paste:
```
◇  Get an OpenAI API key
│  Don't have a key yet?
│  Get one at https://platform.openai.com/api-keys
│  Then come back and paste it below.
│
◆  Put your OpenAI API key (will be saved to secrets.json)
│  •••••••••••••
│
◐  Checking your OpenAI key...
◇  OpenAI key looks good.
```

On rejection:
```
◐  Checking your OpenAI key...
✖  OpenAI rejected the key (HTTP 401).
◇  OpenAI key rejected
│  The provider says this key is not valid.
│  Common causes: typo, expired key, wrong account, or pasting a project-scoped key.
│
│  Get a fresh one at https://platform.openai.com/api-keys
│
◆  What do you want to do?
│  ● Try a different key
│  ○ Save this key anyway
```

Success screen with channel wired:
```
◇  One more step
│  Your agent will not respond on Slack until you claim ownership.
│  This prevents strangers from talking to it.
│  Run `typeclaw role claim` to finish setup.
│
◇  Hatched. Your agent is ready.
│  Claim your agent:   typeclaw role claim
│  Attach TUI:         typeclaw tui
│  Follow logs:        typeclaw logs -f
│  Stop:               typeclaw stop
│  Diagnose issues:    typeclaw doctor
```


devxoul:
## Self-review summary

Ran two parallel review agents on the first commit (`oracle` for correctness/security, `explore` for codebase fit). Both found issues that I addressed in the follow-up commit pushed just now.

### What was fixed

**Security / correctness (Oracle):**

| Severity | Issue | Fix |
|---|---|---|
| Medium | `fetch` follows redirects by default — a CDN/captive-portal 3xx mid-flight would have sent the pasted API key to whatever `Location` named | Added `redirect: 'manual'`; treat 3xx as `skipped/network-error` |
| Medium | Any HTTP 200 was accepted, including HTML from a captive portal / WAF / corporate-MITM proxy returning a login page in front of `/v1/models` | Validate response shape (`{ data: Array }`) before declaring `ok` |
| Low | `await res.text()` on 401/403 read an unbounded body before truncating to 200 chars for display | Streamed reader capped at 4 KiB |
| Low | `runApiKeyValidation` would hang the spinner if `prompts.validateApiKey` threw | Wrapped call site in `try/catch` |
| Nit | Rejection `detail` was collected but never surfaced — added attack surface for zero UX value | Dropped `detail` from rejected results entirely |

**Codebase fit (Explore):**

| Issue | Fix |
|---|---|
| Manual `AbortController` + `setTimeout` instead of `AbortSignal.timeout()` (the pattern used in `src/init/models-dev.ts`) | Switched to `AbortSignal.timeout(TIMEOUT_MS)` |
| New file at `src/init/auth/validate.ts` deviated from the flat `src/init/` convention (no other feature subdirectories) | Moved to `src/init/validate-api-key.ts`; updated import |
| Note title `'One more step'` was conversational, deviating from terse imperative titles in `init.ts` (`'Get a Discord bot token'`, `'Generate the Slack app-level token'`, etc.) | Renamed to `'Claim ownership before chatting'` |
| No drift guard — adding a future api-key provider could silently skip validation | Added a CI-gated test that iterates `KNOWN_PROVIDERS` and asserts every `supportsApiKey()` provider has both a `PROVIDER_PROBE` entry AND an `API_KEY_DASHBOARD_URL` entry |

### What was kept

- **Hardcoded probe URLs** instead of deriving from `KNOWN_PROVIDERS.baseUrl`. Both reviewers landed on the same call: the hardcoded allowlist is a security feature (prevents user-controlled URLs from being probed with credentials). The Anthropic suffix-vs-`/v1` asymmetry would also have made the `baseUrl + suffix` form less clear, not more.
- **Plain-text URLs in `note()` bodies** (no OSC 8 `link()` helper). Consistent with every other `note()` in `init.ts` (Discord, Slack, Telegram, GitHub all use raw URLs). Switching one note to `link()` and not the rest would create new drift.

### Verification

- 12 new tests in `src/init/validate-api-key.test.ts` (was 7 — added redirect-as-skipped, non-JSON 200 as skipped, wrong-shape 200 as skipped, capped body reads, drift guard, dashboard-URL coverage).
- 5034 of the 5043 tests pass on this branch. The 9 failures (`--help` + `_hostd` integration tests) reproduce on `main` and are pre-existing.
- `bun run typecheck` shows only the pre-existing kakaotalk drift errors that also reproduce on `main`.
- `bun run lint` clean (0 errors).
- `bun run format` clean.

Full review reports archived in the commit message of `92d1cc8` → `1f7e...` (this follow-up).

PR #344, opened 2026-05-24
channels: suppress Kimi K2 tool-call delimiter token leaks

devxoul:
## Why

When an inference server's Kimi K2 tool-call parser misbehaves, raw chat-template delimiter tokens leak into the assistant's text content with stop reason "stop" instead of becoming OpenAI-shaped tool calls. The openai-completions provider is vendor-neutral and passes them through verbatim. The recovery path in `validateChannelTurn` then forwards that content to the channel — exposing the model's intended tool-call payload as user-visible nonsense.

Observed in production against Fireworks' kimi-k2p6-turbo router on 2026-05-24. Two tool calls leaked in one turn, posted verbatim to a chat:

```
channel_reply:0<|tool_call_argument_begin|>{"text": "..."}<|tool_calls_section_end|>
channel_reply:1<|tool_call_argument_begin|>{"text": "..."}<|tool_calls_section_end|>
```

This is the canonical Kimi K2 tool-call format per [MoonshotAI/Kimi-K2 tool_call_guidance.md](https://github.com/MoonshotAI/Kimi-K2/blob/1b4022b/docs/tool_call_guidance.md) — the model emitted correctly, the transport layer broke. Same class as [vllm-project/vllm#38579](https://github.com/vllm-project/vllm/pull/38579).

## What

Two commits, each independently revertable:

### 1. `channels: suppress leaked Kimi tool-call delimiter tokens in recovery path`

Adds `isKimiToolCallLeak()` next to the existing upstream-empty-response sentinel in the channel router — same shape, same logging pattern. Gating the recovery send on it means leaked turns log a suppression event and end silently instead of posting raw tokens to the channel.

Detection targets the five unambiguous Kimi chat-template delimiter tokens (the `<|...|>` family). It deliberately does not match on the per-call index suffix alone, which can plausibly appear in legitimate prose discussing code.

Suppression-only, not recovery. A follow-up can parse the format into real tool calls and re-issue from `source: 'system'` if the leak class persists in production.

### 2. `channels: mirror Kimi tool-call leak guard in channel_reply / channel_send`

Defense-in-depth mirror of the router-level guard at the tool boundary. If a future code path hands assistant-generated text directly to the channel send tools — bypassing the `validateChannelTurn` gate — the leak guard catches it and denies the send with a clear error.

The predicate lives in `router.ts` as the single source of truth (same pattern as the existing upstream-empty-response sentinel mirror), and each tool wraps it in a thin error-message function. All three call sites stay in lockstep.

## Tests

- `src/channels/router.test.ts` — 2 new cases: leaked delimiter tokens are suppressed; legit prose without delimiter tokens still recovers normally.
- `src/agent/tools/channel-reply.test.ts` — 3 new cases: blocking delimiter tokens, blocking the per-call begin marker alone (partial parser leak), and NOT blocking legit prose mentioning tool names.
- `src/agent/tools/channel-send.test.ts` — same 3 cases mirrored.

## Verification

```
bun run typecheck    ✓ (0 errors)
bun run lint         ✓ (0 errors — 41 pre-existing warnings in unrelated files)
bun run format:check ✓
bun test --parallel  ✓ 5213 pass / 0 fail across 296 files (24s)
```

6 files changed, 235 insertions(+), 2 deletions(−). All additive — no behavior changes to non-leaked turns, no API changes, no schema changes. `isKimiToolCallLeak` is the only new exported symbol. No breaking changes, no migration required. Suppression takes effect on first restart after upgrade.

### DIFF
--- a/src/agent/tools/channel-reply.test.ts
+++ b/src/agent/tools/channel-reply.test.ts
@@ -393,6 +393,60 @@ describe('createChannelReplyTool', () => {
     })
   })
 
+  describe('Kimi tool-call delimiter leak guard', () => {
+    test('blocks raw `<|tool_call_argument_begin|>...<|tool_calls_section_end|>` tokens from reaching the channel', async () => {
+      const calls: OutboundMessage[] = []
+      const tool = createChannelReplyTool({
+        router: fakeRouter(async (msg) => {
+          calls.push(msg)
+          return { ok: true }
+        }),
+        origin: slackThreadOrigin,
+      })
+      const result = await runTool(tool, {
+        text: 'channel_reply:0<|tool_call_argument_begin|>{"text": "hi"}<|tool_calls_section_end|>',
+      })
+      expect(calls).toHaveLength(0)
+      expect(result.details).toMatchObject({ ok: false })
+      expect((result.details as { error: string }).error).toContain('provider tool-call control tokens')
+      const text = (result.content[0] as { text: string }).text
+      expect(text).toContain('channel_reply denied')
+      expect(text).not.toContain('posted to')
+    })
+
+    test('blocks the per-call begin marker even when section markers are absent (partial parser leak)', async () => {
+      const calls: OutboundMessage[] = []
+      const tool = createChannelReplyTool({
+        router: fakeRouter(async (msg) => {
+          calls.push(msg)
+          return { ok: true }
+        }),
+        origin: slackThreadOrigin,
+      })
+      const result = await runTool(tool, {
+        text: '<|tool_call_begin|>functions.channel_reply:0<|tool_call_argument_begin|>{"text": "hi"}<|tool_call_end|>',
+      })
+      expect(calls).toHaveLength(0)
+      expect(result.details).toMatchObject({ ok: false })
+    })
+
+    test('does NOT block legit prose mentioning tool names without Kimi delimiter tokens', async () => {
+      const calls: OutboundMessage[] = []
+      const tool = createChannelReplyTool({
+        router: fakeRouter(async (msg) => {
+          calls.push(msg)
+          return { ok: true }
+        }),
+        origin: slackThreadOrigin,
+      })
+      const result = await runTool(tool, {
+        text: 'I called channel_reply:0 in the earlier turn, here are the results.',
+      })
+      expect(calls).toHaveLength(1)
+      expect(result.details).toEqual({ ok: true })
+    })
+  })
+
   describe('structured router failures surface as denials', () => {
     test('duplicate code from router renders as channel_reply denied with router error text', async () => {
       const tool = createChannelReplyTool({
--- a/src/agent/tools/channel-reply.ts
+++ b/src/agent/tools/channel-reply.ts
@@ -1,7 +1,12 @@
 import { Type } from '@mariozechner/pi-ai'
 import { defineTool } from '@mariozechner/pi-coding-agent'
 
-import { isNoReplySignal, isUpstreamEmptyResponseSentinel, type ChannelRouter } from '@/channels/router'
+import {
+  containsKimiToolDelimiter,
+  isNoReplySignal,
+  isUpstreamEmptyResponseSentinel,
+  type ChannelRouter,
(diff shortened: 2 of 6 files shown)


devxoul:
## Self-review amendments

Pushed an amended branch addressing findings from a self-review pass. The fix is the same in spirit but tighter where it mattered.

**Force-push contents** (2 commits, +302/-2 across the same 6 files):

### Detector split (commit 1, router.ts)

`isKimiToolCallLeak` was a single predicate matching any of the five documented Kimi delimiter tokens. The recovery-path version is now narrowed:

- **`containsKimiToolDelimiter`** — broad (any of the five delimiter tokens). Exported for the tool-boundary mirror in commit 2.
- **`isLikelyKimiChannelToolLeak`** — narrower: requires BOTH a delimiter AND a recognizable `channel_(reply|send):N` identifier in the same text. Used by `validateChannelTurn`.

The two-signal rule preserves the documented Kimi format ([source](https://github.com/MoonshotAI/Kimi-K2/blob/1b4022b/docs/tool_call_guidance.md)) as legitimate prose. If the agent answers "how does Kimi format tool calls?" with documentation-style text containing `<|tool_calls_section_begin|>`, the channel still receives the explanation. The leak shape observed in production (`channel_reply:0<|tool_call_argument_begin|>{...}<|tool_calls_section_end|>`) trivially satisfies both conditions.

### Denial message sanitized (commit 2, channel-reply.ts / channel-send.ts)

The tool-denial text used to echo literal Kimi delimiter strings back to the model:

> `refusing to forward raw Kimi tool-call delimiter tokens (\`<|tool_call_*|>\`, \`<|tool_calls_section_*|>\`); ...`

For Kimi specifically, reflecting these chat-template tokens into the tool-result risks reinforcing the same pattern on the next generation. Now reads:

> `refusing to forward raw provider tool-call control tokens; these are chat-template delimiters that should have been parsed into a real tool call upstream. Re-issue the intended channel reply as plain user-visible text only.`

The tool boundary still uses the broad `containsKimiToolDelimiter` rather than the narrow router predicate — at the tool boundary the model is literally calling `channel_send` with a body, so any delimiter token in that body is unambiguous leakage with no documentation-prose case to preserve.

### New tests

- **`router.test.ts` — suppresses the canonical full-shape leak**: explicit fixture with the documented two-call shape (`<|tool_calls_section_begin|>` ... two `<|tool_call_begin|>...<|tool_call_end|>` blocks ... `<|tool_calls_section_end|>`). Pins the exact regression.
- **`router.test.ts` — still recovers documentation-style prose explaining Kimi delimiters**: assistant text contains delimiters in backticks but no `channel_reply:N` / `channel_send:N` signal. Asserts the recovery path still fires and the explanation reaches the channel.

The existing four tests still pass with the renamed/narrowed predicates.

### Findings deliberately deferred

- **Transcript poisoning** — the leaked assistant message is persisted in the session JSONL before `validateChannelTurn` runs, so it remains visible to `typeclaw inspect` and the dreaming subagent on the next turn. Confirmed via code reading: this is a **pre-existing concern shared with `isUpstreamEmptyResponseSentinel`**, which has the same shape since 2026-05-21. Fixing only the Kimi side would be inconsistent. Will file a separate PR addressing both guards together.
- **TUI live-render path** — `text_delta` events forward to TUIs before the full message is assembled, so a TUI user may briefly see garbage during a leaked stream. Outside the channel-leak scope; separate PR if it becomes an operational issue.
- **Softening "K2 / K2.5 / K2.6" claim** to "Kimi-family" — done; the documented set in Moonshot's official docs covers K2, with K2.5/K2.6 behavior inferred. Comment now reads "Kimi-family deployments observed/documented with this format".

### Verification

- `bun run typecheck` — 0 errors
- `bun run lint` — 0 errors (41 pre-existing warnings, none in changed files)
- `bun test --parallel` — 5215 pass / 0 fail across 296 files on best runs; pre-existing flakes on `main` produce 0–3 intermittent fails independent of this change (verified by checking out origin/main and re-running)

Final diff: +302/-2 across 6 files. Same 6 files as the original PR.

PR #381, opened 2026-05-26
init: auto-resume from partial secrets.json; add --reset escape hatch

### LINKED ISSUE #330: init: --resume flag to pick up from partial secrets.json state
**Tier B — from the init-UX audit.**

The wizard's `WizardAbortedError` already hints that the user can re-run `typeclaw init` to pick up partial credentials, but the actual resume behavior is not clean: starting fresh re-asks every prompt up to the OAuth step instead of skipping straight to "where you left off".

## Proposed

`typeclaw init --resume` (or auto-resume when `secrets.json` exists with partial credentials):

1. Detect existing `secrets.json` slots (`providers.openai`, `channels.slack-bot`, etc.).
2. For each section that's already populated, skip its corresponding wizard step and emit a `log.info('Reusing existing OpenAI credentials.')` line.
3. Resume from the first incomplete step.
4. Provide `--reset` as the escape hatch to ignore existing state and start fresh.

## Why

Users who hit a transient OAuth failure or had to step away mid-wizard currently have to either delete `secrets.json` (which wipes the partial progress they made) or muddle through the wizard again from the top. Neither is ideal.

## Implementation notes

- Build a "wizard-state-from-disk" helper that hydrates `WizardState` from existing `secrets.json`.
- Wire into `collectWizardInputs` as an initial state instead of `{ catalog }`.
- The "reuse existing" prompts (`askReuseExistingKey`, `askReuseExistingChannel`) are the wedge — extend them to default to true and auto-skip when nothing else needs the user.

## Acceptance

- A user who aborts mid-init and re-runs `typeclaw init` gets the success path with the same credentials they already provided.
- `--reset` re-prompts everything.
- Tests cover the resume + reset paths.

Audit source: post-implementation report on PR #322 Tier B #12.

devxoul:
Closes #330

## What this PR does

Re-running `typeclaw init` after a mid-wizard abort used to re-ask every
credential prompt even when secrets.json already had the answers.
This PR makes init silently skip credential steps that already have valid
data in secrets.json. The new `--reset` flag is the escape hatch that
forces a full re-prompt when needed.

## Root cause

Two "reuse existing" confirms were prompt-then-confirm gates:
`askReuseExistingKey` for the api-key path
and `askReuseExistingChannel` for the channel path.
A user who aborted mid-wizard was presented with them again, and if they
missed the prompt, the wizard would re-collect credentials already safely
on disk.

### Auto-resume for api-key providers

`askReuseExistingKey` is replaced with a silent auto-skip.
When a usable api-key slot is detected in secrets.json, the wizard logs
a reuse message and advances past the auth steps without any interaction.

### Auto-resume for OAuth providers

New helper `hasExistingOAuthCredentials` reads the access token from the
v2 secrets envelope. When a token is present, the browser login flow is
skipped. Mirrors the same guard already applied to both the
default-provider and vision-provider auth steps.

### Auto-resume for channel credentials

`askReuseExistingChannel` is also removed. The existing
"all required fields present" check is preserved; the wizard now
auto-reuses without prompting when the check passes.

### `--reset` escape hatch

`typeclaw init --reset` short-circuits all three existence checks.
Implemented as a `reset: boolean` field plumbed into collectWizardInputs
and gated at the wrapper layer over WizardPrompts. A defense-in-depth
test asserts every existence check is bypassed when reset is true —
future contributors who add a new check without gating it will see
that test fail.

### Back-navigation fix

Back from the pick-auth-method step now unwinds directly to pick-model,
skipping the former silent reuse-existing-key step (which no longer exists
as a named state). The `pendingBackOrigin` guard for the pick-channel
back-loop is unaffected — verified against the existing OAuth abort test.

### WizardPrompts surface changes

Removed: `askReuseExistingKey`
Removed: `askReuseExistingChannel`
Removed: `decideExistingApiKeyReuse` (prompt-then-confirm semantics no longer match production)
Added: `hasExistingOAuthCredentials`

## Tests

5420 pass. New tests cover:

- auto-resume via api-key path
- auto-resume via OAuth path
- auto-resume via channel credentials path
- `--reset` bypasses api-key auto-skip
- `--reset` bypasses OAuth auto-skip
- `--reset` bypasses channel auto-skip
- defense-in-depth: all three existence checks bypassed when reset is true

## Acceptance criteria (issue #330)

- User aborts mid-init, re-runs, reaches the success path with saved credentials — no duplicate prompts.
- `typeclaw init --reset` re-prompts everything regardless of what is in secrets.json.
- Tests cover both resume paths and both reset paths.

## Verification

```
bun run typecheck    ✓
bun run lint         ✓
bun run format       ✓
bun run test         5420 pass / 0 fail
```

## Files changed

| File | Change |
|---|---|
| `src/cli/init.ts` | Remove reuse-existing prompts; wire the reset flag; fix back-navigation. |
| `src/cli/init.test.ts` | Update to silent-skip surface; add auto-resume and reset tests. |
| `src/init/index.ts` | Add hasExistingOAuthCredentials; plumb reset into collectWizardInputs. |
| `src/init/index.test.ts` | Tests for hasExistingOAuthCredentials and reset flag behavior. |

### DIFF
--- a/src/cli/init.test.ts
+++ b/src/cli/init.test.ts
@@ -5,56 +5,11 @@ import type { ModelOption } from '@/init/models-dev'
 
 import {
   collectWizardInputs,
-  decideExistingApiKeyReuse,
   formatModelLabel,
   sortRecommendedFirst,
   WizardAbortedError,
   type WizardPrompts,
 } from './init'
-
-describe('decideExistingApiKeyReuse', () => {
-  test('reuses an existing API key when the user confirms', async () => {
-    const messages: string[] = []
-
-    const decision = await decideExistingApiKeyReuse(KNOWN_PROVIDERS.openai, 'openai-existing-key', async (message) => {
-      messages.push(message)
-      return true
-    })
-
-    expect(decision).toBe('reuse')
-    expect(messages).toEqual(['Reuse existing OpenAI API key from secrets.json?'])
-  })
-
-  test('prompts for a new API key when the user declines reuse', async () => {
-    const decision = await decideExistingApiKeyReuse(KNOWN_PROVIDERS.fireworks, 'fw_existing', async () => false)
-
-    expect(decision).toBe('prompt')
-  })
-
-  test('skips the reuse question when there is no existing API key', async () => {
-    let asked = false
-
-    const decision = await decideExistingApiKeyReuse(KNOWN_PROVIDERS.openai, null, async () => {
-      asked = true
-      return true
-    })
-
-    expect(decision).toBe('prompt')
-    expect(asked).toBe(false)
-  })
-
-  test('skips the reuse question for OAuth-only providers', async () => {
-    let asked = false
-
-    const decision = await decideExistingApiKeyReuse(KNOWN_PROVIDERS['openai-codex'], 'unused-key', async () => {
-      asked = true
-      return true
-    })
-
-    expect(decision).toBe('prompt')
-    expect(asked).toBe(false)
-  })
-})
 
 describe('collectWizardInputs back-aware flow', () => {
   const fireworksModel: ModelOption = {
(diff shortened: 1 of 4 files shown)


devxoul:
@typeey yo

devxoul:
@typeey yo

PR #398, opened 2026-05-27
channels: decorate github outbound 403s with the exact permission fix

devxoul:
## What this PR does

When the github channel adapter posts a comment, PR review reply, or discussion comment and GitHub responds with `403 "Resource not accessible by integration"`, the existing `GitHub API 403: {body}` error string now gets a trailing **per-endpoint, per-auth-type "Fix:"** block telling the user exactly which permission to enable.

This is the third in a series of three PRs triggered by a real-world bug report. PR-A (#395, merged) refactored permission guidance into its own module. PR-B (#397, merged) added a startup permission preflight that catches misconfigurations before any event arrives. This PR catches the **other** path: a permission that was correct at start but becomes insufficient later (App settings edited mid-session, install owner downgraded the permission, new repo added to a misconfigured token), or a startup where the preflight itself was skipped due to a transient GitHub-side failure.

The two halves are complementary, not redundant: the preflight runs once at start with a single api call against `/app/installations/{id}`, the outbound decoration runs every time we touch GitHub. Either alone would have shipped a usable improvement; together they make the right fix discoverable from either path.

## Before / after

The user's reported symptom, against the actual misconfigured App on the actual broken `issue:394`:

**Before this PR**:
```
[channels] channel_reply failed: github:typeclaw/typeclaw/issue:394: GitHub API 403:
  {"message":"Resource not accessible by integration","documentation_url":"https://docs.github.com/rest/issues/comments#create-an-issue-comment","status":"403"}
```

**After this PR** (verified end-to-end against the same code path):
```
[channels] channel_reply failed: github:typeclaw/typeclaw/issue:394: GitHub API 403:
  {"message":"Resource not accessible by integration","documentation_url":"https://docs.github.com/rest/issues/comments#create-an-issue-comment","status":"403"}
  Fix (GitHub App): the App needs "Issues" → "Read and write".
    1. Open https://github.com/settings/apps and edit the app TypeClaw is using.
    2. Under "Permissions & events" → "Repository permissions", set "Issues" to "Read and write". Save.
    3. Open the install page (Install App / Configure for the org) and accept the updated permissions request — the new access only takes effect after the install owner accepts.
```

If the user is on a PAT instead of an App, the same 403 surfaces with PAT-specific guidance (`"Repository permissions" → "Issues" → "Read and write"` for fine-grained PATs, plus a parenthetical for classic PATs with the `repo` scope name).

## Three endpoint kinds, three different fixes

`createGithubOutboundCallback` now knows which API call it's making when an error comes back, so the guidance is precise rather than generic:

| TypeClaw call site | GitHub API call | App permission | PAT (fine-grained) | PAT (classic) |
|---|---|---|---|---|
| Issue / top-level PR comment | `POST /repos/{o}/{r}/issues/{n}/comments` | **Issues** → Read & write | **Issues** → Read & write | `repo` (or `public_repo`) |
| PR review thread reply | `POST /repos/{o}/{r}/pulls/{n}/comments/{}/replies` | **Pull requests** → Read & write | **Pull requests** → Read & write | `repo` |
| Discussion comment | GraphQL `addDiscussionComment` | **Discussions** → Read & write | **Discussions** → Read & write | `repo` |

A test pins each case to ensure the wrong permission is never suggested for the wrong endpoint (a PR-reply 403 must not say "enable Issues").

## Narrow matcher: only the specific permission-denial body

The decoration is gated by a single matcher: **HTTP 403 AND the response body contains the literal string `"Resource not accessible by integration"`** (GitHub's canonical signal for "this would work with more permissions"). Other 403 shapes get **no** decoration because their fix is different:

- `"Resource protected by organization SAML enforcement."` — fix is to authorize the token/App for the org's SAML, not to change permissions.
- `"Installation suspended"` — fix is to unsuspend the App.
- `"Must have admin rights to Repository."` — fix is to install/configure access at a higher level.

A unit test in `outbound.test.ts` proves that the SAML body **does not** trigger decoration — getting that wrong would actively mislead users into the wrong fix.

## GraphQL quirk handled

Discussion comments go through GraphQL, where permission denials usually arrive as **HTTP 200 + `errors[]` with `type: 'FORBIDDEN'` and the same `"Resource not accessible by integration"` message**, not as HTTP 403. The `graphql()` helper threads the matcher against the error message text (with status synthesized as 403 when the response was 200-with-errors) so discussion permission failures get the same decoration as REST failures. A comment at the matcher site explains the asymmetry so the next maintainer doesn't "simplify" it back to status-only matching.

## Back-compat: prefix preserved

The existing error string format `GitHub API ${status}: ${body}` is preserved verbatim. The fix block is **appended** with a leading newline. A unit test specifically asserts `result.error.startsWith('GitHub API 403:')` is still true so any downstream log parser or test that keys on the prefix keeps working.

## Commits

Two atomic commits on this branch:

### Commit 1 — `channels: add outbound github permission denial helpers and per-endpoint guidance`
Pure additions to `permission-guidance.ts`:
- `OutboundEndpointKind` type
- `buildOutboundPermissionGuidance({ authType, endpointKind })` formatter
- `isOutboundPermissionDenial(status, body)` matcher

Plus 10 unit tests in `permission-guidance.test.ts` covering all three endpoint kinds × both auth types, the SAML-rejection negative case, the non-403 negative case, and the embedded-in-longer-body positive case. No caller yet; lands cleanly on its own.

### Commit 2 — `channels: decorate github outbound 403s with the exact permission fix`
The wiring:
- `outbound.ts`: `createGithubOutboundCallback` takes `authType`; `postJson` and `graphql` take `{ authType, endpointKind }` and decorate via `buildOutboundPermissionGuidance` when `isOutboundPermissionDenial` matches.
- `index.ts`: one-line passthrough of `options.secrets.auth.type` into the outbound callback.
- `outbound.test.ts`: 6 new tests for each endpoint × auth combination plus the SAML / non-403 negative cases.
- `line-comment-thread.test.ts`: 3 existing call sites add the now-required `authType: 'app'` field.

## Files

| File | Status | Lines |
|---|---|---|
| `src/channels/adapters/github/permission-guidance.ts` | EDIT | +72 −0 |
| `src/channels/adapters/github/permission-guidance.test.ts` | EDIT | +62 −0 |
| `src/channels/adapters/github/outbound.ts` | EDIT | +60 −22 |
| `src/channels/adapters/github/outbound.test.ts` | EDIT | +175 −3 |
| `src/channels/adapters/github/line-comment-thread.test.ts` | EDIT | +3 −3 |
| `src/channels/adapters/github/index.ts` | EDIT | +6 −1 |

## Verification

```
bun run typecheck                                                ✓
bun run lint                                                     ✓  (0 errors)
bun run format:check                                             ✓
bun run test                                                  5579 pass, 0 fail across 307 files  (+16 vs main)
```

End-to-end against the actual reported bug: `createGithubOutboundCallback` invoked with the user's `authType='app'` and a fake fetch returning the exact 403 GitHub returned, produces the after-text above byte-for-byte.

## Sequencing

PR-C in a 3-PR series. PR-A (#395) and PR-B (#397) merged ahead of this. This branch was originally stacked on PR-A's branch; after the rebase-merge of PR-A and PR-B to `main`, the branch has been rebased onto current `main` (clean rebase, conflict resolution only in the `permission-guidance.ts` and `permission-guidance.test.ts` "added at end of file" overlaps with PR-B — both files now contain both sets of additions).

After this PR merges: the github channel surfaces App misconfiguration twice — once at start, once on first attempted reply — both times pointing at the exact same fix. Either signal alone resolves the user's session.

### DIFF
--- a/src/channels/adapters/github/index.ts
+++ b/src/channels/adapters/github/index.ts
@@ -78,7 +78,12 @@ export function createGithubAdapter(options: GithubAdapterOptions): GithubAdapte
     process.env.GH_TOKEN = t
     return t
   }
-  const outbound = createGithubOutboundCallback({ token: tokenFn, logger, fetchImpl })
+  const outbound = createGithubOutboundCallback({
+    token: tokenFn,
+    authType: options.secrets.auth.type,
+    logger,
+    fetchImpl,
+  })
   const history = createGithubHistoryCallback({
     token: tokenFn,
     fetchImpl,
--- a/src/channels/adapters/github/line-comment-thread.test.ts
+++ b/src/channels/adapters/github/line-comment-thread.test.ts
@@ -123,7 +123,12 @@ describe('PR line comment → channel_reply → /pulls/{N}/comments/{T}/replies'
       },
       { preconnect: () => {} },
     ) as typeof fetch
-    const outbound = createGithubOutboundCallback({ token: async () => 'tok', logger: silent, fetchImpl })
+    const outbound = createGithubOutboundCallback({
+      token: async () => 'tok',
+      authType: 'app',
+      logger: silent,
+      fetchImpl,
+    })
 
     const origin: ChannelReplyOrigin = {
       adapter: 'github',
@@ -165,7 +170,12 @@ describe('PR line comment → channel_reply → /pulls/{N}/comments/{T}/replies'
       },
       { preconnect: () => {} },
     ) as typeof fetch
-    const outbound = createGithubOutboundCallback({ token: async () => 'tok', logger: silent, fetchImpl })
+    const outbound = createGithubOutboundCallback({
+      token: async () => 'tok',
+      authType: 'app',
+      logger: silent,
+      fetchImpl,
+    })
 
     const tool = createChannelReplyTool({
       router: fakeRouter(async (msg) => outbound(msg)),
@@ -192,7 +202,12 @@ describe('PR line comment → channel_reply → /pulls/{N}/comments/{T}/replies'
       },
       { preconnect: () => {} },
     ) as typeof fetch
-    const outbound = createGithubOutboundCallback({ token: async () => 'tok', logger: silent, fetchImpl })
+    const outbound = createGithubOutboundCallback({
+      token: async () => 'tok',
+      authType: 'app',
+      logger: silent,
+      fetchImpl,
+    })
 
     const tool = createChannelReplyTool({
       router: fakeRouter(async (msg) => outbound(msg)),
--- a/src/channels/adapters/github/outbound.test.ts
+++ b/src/channels/adapters/github/outbound.test.ts
@@ -18,6 +18,7 @@ describe('createGithubOutboundCallback', () => {
   it('posts issue and top-level PR comments through the issues comments endpoint', async () => {
     const cb = createGithubOutboundCallback({
       token: async () => 'tok',
+      authType: 'app',
       logger,
       fetchImpl: fakeFetch({
         'POST https://api.github.com/repos/acme/project/issues/42/comments': { status: 201, body: { id: 1 } },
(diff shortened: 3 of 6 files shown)


devxoul:
@typeey yo

devxoul:
@typeey can you review this PR

PR #404, opened 2026-05-27
channels: add skip_response signal for structured channel-session silent turns

devxoul:
## Problem

Operators on Slack and Discord channels kept seeing the agent narrate its own silent-turn decision into the user-facing chat. Real production example:

> The subagent result confirms the same story I already shared — SpaceX IPO filing and NASA Artemis updates driving the rally. No new info to add beyond what I already sent, so I'll stay quiet here.

That string was the model's reasoning about why this turn should be silent, but it landed in the channel. The existing `NO_REPLY` text sentinel only fires when the model emits exactly `NO_REPLY` (optionally with leading prose that `endsWithNoReplySignal` strips). When the model produces a meaningful explanation but never appends `NO_REPLY`, that explanation leaks.

Two failure modes in practice:
- **Channel leak** — reasoning/prose appears as if it were a user-facing reply.
- **Silent with no record** — model emits empty text, router suppresses, operator has zero structured signal of why.

## Solution

Add a structured `skip_response({ reason: string })` tool as a third path.

The tool records the reason via `console.warn` (visible in `typeclaw logs -f`), calls `router.markTurnSkipped()` to suppress recovery for that turn, and returns a confirmation to the model. Calling `channel_reply` or `channel_send` in the same turn after `skip_response` is rejected with the new `skip-locked` `SendErrorCode` — commit to silence or commit to replying, not both. System-source sends (recovery, role-claim) bypass the lock.

The tool is only wired into channel-origin sessions. Cron has no user-facing surface to suppress and is intentionally excluded.

**Turn-scoped state, not session-scoped.** `LiveSession` gains a monotonic `turnSeq` (incremented in `drain()` just before each `prompt()`) and a `skippedTurn: { turnSeq, reason } | null` field. `validateChannelTurn` only honors the flag when `skippedTurn.turnSeq === turnSeq`. A flag set by a prior turn that crashed before validation is ignored — prevents silent next-turn drops. (Oracle flagged a bare session boolean as a real bug for this reason.)

`NO_REPLY` is kept intact as a legacy fallback for cases where the model cannot call a tool. Prompt guidance now teaches `skip_response` as preferred and `NO_REPLY` as fallback.

## Tests

6 new router tests cover the four invariants Oracle flagged:
1. Skip-only turn suppresses recovery, logs reason.
2. Skip + leaked prose still suppresses recovery.
3. Stale `skippedTurn` from a prior `turnSeq` does **not** drop the next turn.
4. Same-turn `channel_send` after `skip_response` → `skip-locked`.
5. System-source sends bypass the skip lock.
6. `markTurnSkipped` with an unknown `parentSessionId` returns `no-live-session`.

5 unit tests in `skip-response.test.ts` cover happy path, trim, empty reason, no-live-session, and schema invariants (maxLength 500).

```
bun run typecheck    ✓
bun run lint         ✓  (61 pre-existing warnings, 0 errors, 0 new)
bun run format       ✓
bun run test         5596 pass, 0 fail
```

## Review Notes

- The two comments worth reading first: the `LiveSession.turnSeq` field doc in `router.ts` (explains why it is never cleared) and the skip-check block at the top of `validateChannelTurn` (explains why it runs before all other recovery paths).
- The `skip-locked` rejection is a new branch above the existing `TURN_CAP_ERROR` / `DUPLICATE_SEND_ERROR` checks in `router.send()`. System-source sends explicitly bypass it.
- `buildChannelTools` takes `sessionId` as an optional third arg so existing 2-arg test callers compile without changes.
- No changes to `isNoReplySignal` / `endsWithNoReplySignal` behavior — fully backward compatible.


devxoul:
Thanks — you're right that there was a real gap, and I tightened it in 9a7f293 + cecd344 + 6cc5f1e (just pushed).

The original lock was one-directional: `channel_reply`/`channel_send` AFTER `skip_response` returned `skip-locked`, but `skip_response` AFTER a successful channel send was silently accepted — the reply landed AND the skip log fired, leaving operators a contradictory audit trail.

Fix: `markTurnSkipped` now snapshots `successfulChannelSends` at turn start (next to the `turnSeq` increment) and returns a new `send-already-happened` outcome when a tool-source send has landed since. The skip flag is *not* stamped on rejection, so the turn proceeds as a normal reply turn — no need to unwind anything. The tool surfaces it to the model as a structured error ("commit to silence or commit to replying, not both") and the channel-origin prompt now spells out the contract in both directions.

On your broader point about side effects: I deliberately scoped the guard to channel sends only. Non-channel side-effecting tools (bash, edit, write, spawn_subagent) can still run before `skip_response` — by the time the model decides to be silent, those effects have already happened on disk / in subprocesses, and rejecting the skip wouldn't unwind them. The audit invariant that matters is 'skipped' vs 'replied', and that's now symmetric. If we ever want to enforce 'skip_response must be the literal first/only tool', that would need pi-coding-agent hook plumbing per tool — happy to spike it in a follow-up if real cases show up in logs, but it felt like overreach for the current observed failure mode.

Tests cover both rejection directions plus a cross-turn reset (a send in turn N does not block a clean skip in turn N+1) so a future refactor that drops the snapshot doesn't silently regress.

Issue #394, opened 2026-05-27
Typeey 🐣


devxoul:

devxoul:
@typeey

devxoul:
@typeey Yo are you up?

devxoul:
@typeey summarize recent PRs

devxoul:
Good, what's your favorite?

devxoul:
then go to the PR and leave a comment

devxoul:
why it took so long

devxoul:
Looked at the session trace. The 9-minute gap was **one LLM round-trip stalling**, not careful verification — Typeey is confabulating.

The transcript (`sessions/2026-05-27T07-14-41-626Z_…jsonl`, lines 9–14) shows:

- 07:15:10 inbound delivered
- **07:24:25** first assistant response returns — a `read` of the local `typeclaw-channel-github` skill
- 07:24:32 `gh pr comment` posts (1.7s)
- 07:24:35 `channel_reply` "Done" sent

Zero PR-context fetching happened. The "I wanted to verify the PR context" reply is a post-hoc rationalization — there is no `gh pr view`, no diff read, no web fetch in the trace. The 9 minutes are entirely a `gpt-5.4-mini` first-token stall on that single request (the previous turn in the same session completed in ~9s, same model, same prefix). Likely provider-side queueing on Fireworks/Codex for that call; nothing typeclaw can do short of a per-request timeout + retry on the channel agent path.

Minor side note: the agent re-read the `typeclaw-channel-github` skill via the generic `read` tool instead of it being preloaded for a github-origin channel session — worth checking whether that skill is in the auto-inject set for `origin.kind=channel, adapter=github`.

devxoul:
@typeey is operational now

devxoul:
@typeey yo

devxoul:
@typeey yo

PR #406, opened 2026-05-27
run: observe Codex Responses fetch with phase timings

devxoul:
## Problem

Issue #394 (agent took ~9 min to reply to a simple instruction) prompted a deep analysis of session traces. Aggregating stalls across a full day on the affected agent revealed a pattern that rules out per-request flakiness:

- 21 stalls >60 s in a single day
- 4 stalls >5 min (worst at 9m 58s)
- Stalls cluster in time across distinct sessions: 6 stalls in a 19-minute window, 7 in a 33-minute window, with long zero-stall gaps in between — incident-shaped, not a stable per-request rate
- No correlation with payload size: a 4-token output request stalled for 268 s; a 16,827-token input request stalled for 292 s
- Same model, same account, same container throughout

The `openai-codex-responses` provider in `@mariozechner/pi-ai` does a bare `await fetch(url, { signal })` then a bare `reader.read()` on the SSE body, with no idle timeout and no phase logging. The latest pi-ai release (0.73.1) has the same code as the vendored 0.67.3 — bumping the dependency is a no-op.

We do not yet know **which phase** of the Codex request consumes the time:

- TCP connect / TLS handshake stalled before headers?
- Request sent; response headers never arrived (server-side queue)?
- Headers received; first SSE byte never started flowing (worker assigned but blocked)?
- Mid-stream gap after partial response?

Each cause needs a different fix. Shipping a client-side abort or retry based on a guess could make things worse — if the issue is upstream queueing (the most likely explanation given the clustering pattern), aborting just rejoins the queue.

## What this PR does

Installs a passive `globalThis.fetch` wrapper at the top of `startAgent` (container stage only — host CLI commands go through different entrypoints and are unaffected). For requests matching `chatgpt.com/.../codex/responses` (POST), the wrapper logs one line to stderr per completed request:

```
[codex-fetch] status=200 headers_ms=243 first_byte_ms=812 total_ms=14523 body_bytes=8421 retry_after=null request_id=req_abc123 error=null
```

Fields:

| Field | Meaning |
|---|---|
| `status` | HTTP status code; `null` if fetch itself rejected |
| `headers_ms` | Milliseconds from request POST to response headers received |
| `first_byte_ms` | Milliseconds from request POST to first SSE byte |
| `total_ms` | Milliseconds from request POST to body end (or error) |
| `body_bytes` | Total bytes streamed |
| `retry_after` | Value of the `Retry-After` header if present — explicit throttling signal |
| `request_id` | Value of `x-request-id` — correlatable with provider support |
| `error` | Error message if the stream or fetch rejected |

The wrapper is purely passive: no abort, no retry, no header injection. Bytes flow through a `TransformStream` untouched and are forwarded byte-for-byte to the original consumer. Non-Codex URLs are not wrapped at all — zero overhead. Stream errors and pre-response fetch rejections still log with whatever timing fields are available.

Opt out: `TYPECLAW_CODEX_FETCH_OBSERVER=off`.

## Why diagnose-only

The incident-shaped clustering pattern means something external to the container changes state (upstream queue depth, a worker pool cycling, a rate-limit window). Without knowing which phase stalls, a client-side mitigation risks:

- Aborting a request that was about to succeed (wasting the queue position already held)
- Masking a billing or throttle signal the upstream is deliberately emitting
- Choosing the wrong retry strategy for the wrong failure mode

Once this observer runs through a few stall events, the phase breakdown will either point to a specific mitigation (request-level deadline if it's `headers_ms`, mid-stream keepalive if it's `first_byte_ms` after headers succeed, model fallback profile in `src/agent/model-fallback.ts` if it's account-level) or surface an upstream issue to file at the provider. That targeted mitigation is the next PR.

## Files changed

| File | Change |
|---|---|
| `src/run/codex-fetch-observer.ts` | URL matcher + body-tap via `TransformStream` + second `ReadableStream` wrap so consumer cancel is observable. Preserves `globalThis.fetch.preconnect` if Bun sets it. |
| `src/run/codex-fetch-observer.test.ts` | 11 tests: non-Codex passthrough (asserts sentinel identity, verifying zero wrapping), upstream error propagation, phase timings with deterministic clock and controllable body, Retry-After throttling signal, mid-stream errors, pre-response errors, URL match scope (same host, different path), uninstall restores original, double-install warns, env opt-out, custom host override. |
| `src/run/index.ts` | `installCodexFetchObserver` called at top of `startAgent` (before plugin and session construction); `uninstallCodexFetchObserver` added to the `stop` cleanup chain. |

## Verification

```
bun run typecheck    ✓  (0 errors)
bun run lint         ✓  (60 pre-existing warnings, 0 errors, 0 new)
bun run format:check ✓
bun run test         ✓  (5629/5629 pass)
```

## Review Notes

The key read is `src/run/codex-fetch-observer.ts`: verify the `TransformStream` tap does not buffer — chunks must be enqueued synchronously and the controller must not accumulate. The sentinel-identity assertion in the non-Codex passthrough test is the strongest guarantee that unmatched URLs are completely unwrapped.

Once this lands, grep for `[codex-fetch]` in `typeclaw logs -f` on the affected agent. The `headers_ms` vs `first_byte_ms` vs `total_ms` split across stall events will determine whether the next PR is a deadline, a keepalive, or a provider bug report.

This PR relates to #394 but does not fix it — the fix will follow once there is phase-level evidence to act on.

### DIFF
--- a/src/run/codex-fetch-observer.test.ts
+++ b/src/run/codex-fetch-observer.test.ts
@@ -1,0 +1,345 @@
+import { afterEach, beforeEach, describe, expect, test } from 'bun:test'
+
+import { installCodexFetchObserver } from './codex-fetch-observer'
+
+type LogEntry = { level: 'info' | 'warn'; msg: string }
+
+function captureLogger(): { logger: { info: (m: string) => void; warn: (m: string) => void }; entries: LogEntry[] } {
+  const entries: LogEntry[] = []
+  return {
+    entries,
+    logger: {
+      info: (m) => entries.push({ level: 'info', msg: m }),
+      warn: (m) => entries.push({ level: 'warn', msg: m }),
+    },
+  }
+}
+
+// Programmable clock so phase timings are deterministic. `now()` returns
+// whatever the test most recently set; production reads timings via the
+// `now` option, so we never depend on real wall-clock in assertions.
+function makeClock() {
+  let ms = 0
+  return {
+    now: () => ms,
+    set: (v: number) => {
+      ms = v
+    },
+  }
+}
+
+// Body controller pair: lets the test push chunks/close/error into a Response
+// body at the moments it chooses, after walking the clock to the right value.
+// More predictable than a self-scheduling fake because each event awaits a
+// microtask flush before the next test step assertion.
+function makeControllableBody(): {
+  body: ReadableStream<Uint8Array>
+  push: (data: Uint8Array) => Promise<void>
+  close: () => Promise<void>
+  error: (err: Error) => Promise<void>
+} {
+  let controllerRef: ReadableStreamDefaultController<Uint8Array> | null = null
+  let resolveReady!: (c: ReadableStreamDefaultController<Uint8Array>) => void
+  const ready = new Promise<ReadableStreamDefaultController<Uint8Array>>((resolve) => {
+    resolveReady = resolve
+  })
+  const body = new ReadableStream<Uint8Array>({
+    start(controller) {
+      controllerRef = controller
+      resolveReady(controller)
+    },
+  })
+  const flush = () => new Promise<void>((r) => queueMicrotask(r))
+  return {
+    body,
+    push: async (data) => {
+      const c = controllerRef ?? (await ready)
+      c.enqueue(data)
+      await flush()
+      await flush()
+    },
+    close: async () => {
+      const c = controllerRef ?? (await ready)
+      c.close()
+      await flush()
+    },
+    error: async (err) => {
+      const c = controllerRef ?? (await ready)
+      c.error(err)
+      await flush()
+    },
+  }
+}
+
+// Drain a readable stream to completion, ignoring chunks. Lets us trigger the
+// observer's final log without caring about what the SSE consumer would do.
+async function drainQuiet(stream: ReadableStream<Uint8Array> | null): Promise<void> {
+  if (stream === null) return
+  const reader = stream.getReader()
+  try {
+    while (true) {
+      const { done } = await reader.read()
+      if (done) break
+    }
+  } catch {
+    // expected for error-stream tests
+  }
+}
+
+const originalFetch = globalThis.fetch
+
+describe('installCodexFetchObserver', () => {
+  beforeEach(() => {
+    // Tests can mutate globalThis.fetch; reset to the platform default each
+    // time so a misbehaving test doesn't leak into the next.
+    globalThis.fetch = originalFetch
+  })
+
+  afterEach(() => {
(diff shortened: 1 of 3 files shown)


devxoul:
@typeey yo review this

devxoul:
Update: the observer caught a live stall while you were waiting for this review. Two consecutive `[codex-fetch]` lines from the container just now:

```
[codex-fetch] status=null headers_ms=null first_byte_ms=null total_ms=268275 body_bytes=0      retry_after=null request_id=null error="The operation timed out."
[codex-fetch] status=200  headers_ms=857  first_byte_ms=863  total_ms=4303   body_bytes=212679 retry_after=null request_id=null error=null
```

Decoded:

- The first request **hung for 268.275 s before any response headers arrived** (`status=null`, `headers_ms=null`, `first_byte_ms=null`, `body_bytes=0`). It died with `"The operation timed out."` — Bun-runtime / OS socket-idle timeout finally killing the connection.
- `pi-coding-agent`'s built-in retry layer (`_isRetryableError` matches `"timeout"` / `"terminated"` / `"fetch failed"`) then retried the same request, which succeeded normally in 4.3 s.

So the user-visible 4.5-minute wait was: ~268 s of dead pre-headers TCP, then a 4.3 s healthy retry. The retry path already exists; it just kicks in too late because nothing shortens the wait.

**This rules out the SSE-body-stall hypothesis from #394.** The stall is **pre-response-headers**: the request reaches Cloudflare/Codex, sits on the open TCP socket, no bytes flow either way (confirmed independently via `/proc/net/tcp` inside the container during the stall — `tx_queue=0 rx_queue=0`, zero retransmits, healthy TCP keepalive), and Codex eventually never produces headers on that particular request. The TCP connection itself is fine; the upstream worker assignment is what fails.

**Next step (separate PR):** wrap the existing fetch in an `AbortSignal.timeout(N)` (or `AbortSignal.any([options.signal, AbortSignal.timeout(N)])` to preserve the pre-existing abort signal). Aborting on timeout produces an error that `_isRetryableError` already classifies as retryable, so `pi-coding-agent` retries automatically — exactly the path we just observed working, but starting after `N` seconds instead of ~268 s.

Open question for the follow-up: what is `N`?

- Healthy first-byte times in production look like ~860–1000 ms on this account.
- A 30 s ceiling would catch the stall cliff (~268 s) early without false-positiving normal latency.
- A 60 s ceiling is more conservative; still ~4× faster recovery than today.

I will keep this PR diagnose-only as it is, ship it, and open a follow-up that adds the timeout on top of the observer. That follow-up should be informed by a few more `[codex-fetch]` data points, which the merged version of this PR will provide.

PR #408, opened 2026-05-27
run: cap Codex SSE pre-headers + inter-chunk idle waits

devxoul:
## Problem

On 2026-05-27 the `[codex-fetch]` observer added in #406 caught a live production stall:

```
[codex-fetch] status=null headers_ms=null first_byte_ms=null total_ms=268275 body_bytes=0 retry_after=null request_id=null error=fetch failed
```

The request hung for 268.275 s with no headers and no bytes, dying only when Bun's underlying OS socket timeout fired. The existing pi-coding-agent retry then succeeded in 4.3 s. The full ~4.5-minute user-visible wait was almost entirely waiting for Bun to give up so the retry could fire.

This matches the pattern documented in #394: the `openai-codex-responses` provider in `@mariozechner/pi-ai` does a bare `await fetch(url, { signal })` followed by a bare `reader.read()` on the SSE body, with no TTFB timeout and no inter-chunk idle deadline. The observer (#406) diagnosed which phase stalls; this PR adds the timers to trigger the retry sooner.

This PR does not fix the root cause — that is upstream and beyond our reach. It mitigates the user-visible wait by making the existing retry path fire in ~15 s instead of ~268 s.

## What this PR does

Extends `src/run/codex-fetch-observer.ts` with two timers.

### 1. TTFB timeout (default 15 s)

An `AbortController` fires if response headers don't arrive within `ttfbMs`. Composed with the caller's signal via `AbortSignal.any` so user-initiated aborts still win. The thrown error contains `"timed out"` in its message, which pi-coding-agent's `_isRetryableError` already classifies as retryable — no new retry plumbing is needed.

The 15 s default is empirically chosen:

- Healthy first-byte p50 on the observed account: ~860 ms
- Pathological-but-healthy upper bound (TLS handshake + cache-miss prefill + Cloudflare PoP routing): ~7 s
- 15 s is ~2x that bound
- False-positive cost: one extra retry (~5 s)
- False-negative cost: full Bun socket deadline (~268 s)

Aggressive wins here.

### 2. Sliding inter-chunk idle timeout (default 300 s)

Wraps the SSE body reader in a `Promise.race` between `reader.read()` and an idle abort promise. The deadline resets on every chunk. If no bytes arrive within the window the stream errors with `"idle for Nms"`, also classified retryable.

The 300 s default matches `openai/codex` Rust CLI's `DEFAULT_STREAM_IDLE_TIMEOUT_MS` (see `codex-rs/codex-api/src/sse/responses.rs:399-434`, commit `7d47056`). Aligning with the official client is the most defensible posture for an unofficial endpoint. The Rust CLI has no TTFB split (openai/codex issue #23807 argues that is a bug); we add one here.

### New log field

The observer log line gains a `cause=` field:

```
[codex-fetch] status=null headers_ms=null first_byte_ms=null total_ms=14823 body_bytes=0 retry_after=null request_id=null error=fetch timed out after 15000ms cause=ttfb_timeout
```

Values: `ttfb_timeout`, `idle_timeout`, or `null`.

### Env overrides

| Variable | Default | Effect |
|---|---|---|
| `TYPECLAW_CODEX_TIMEOUTS=off` | — | Disable both timers; keep observer running |
| `TYPECLAW_CODEX_TTFB_MS=<int>` | `15000` | Override the TTFB deadline |
| `TYPECLAW_CODEX_IDLE_MS=<int>` | `300000` | Override the idle deadline; `0` disables idle only |

## Why the existing retry path handles our injected error correctly

The recovery chain, audited:

1. Our TTFB timer calls `controller.abort(new Error("fetch timed out after 15000ms"))`.
2. `AbortSignal.any` propagates that reason as the `fetch()` rejection.
3. The `openai-codex-responses` provider (`pi-ai/dist/providers/openai-codex-responses.js:170-185`) catches the fetch throw, checks `!agentSignal.aborted`, sets `stopReason: "error"` and `errorMessage = error.message`.
4. pi-coding-agent's `_isRetryableError` (`agent-session.js:1931`) matches `/timed? out|timeout|terminated/` on `message.errorMessage` — our message hits the first branch.
5. The retry fires immediately.

The same chain applies for `"idle for Nms"` on the idle timeout path.

**Why composing with the caller signal matters.** pi-coding-agent's `AgentSession.abort()` aborts the agent signal. If we replaced the caller signal rather than composing with it, hitting Escape would not cancel the in-flight Codex POST. The test `TTFB composes with caller-provided AbortSignal (caller abort still wins)` proves caller abort propagates and the wrapper does not swallow it. This was also empirically verified with `bun -e`: `AbortSignal.any([userSignal, ttfbController.signal])` propagates the controller's abort reason all the way out of `fetch()`.

## Files changed

| File | Change |
|---|---|
| `src/run/codex-fetch-observer.ts` | +146/-15. New `ttfbMs`/`idleMs`/`scheduler` options. TTFB timer + `AbortSignal.any` composition + custom error. Sliding idle timer via `Promise.race`. New `cause=` field on the log line. |
| `src/run/codex-fetch-observer.test.ts` | +308/-0. New `makeScheduler()` helper + 7 new test cases: TTFB fires with custom error, TTFB cancelled on headers, caller abort beats TTFB, idle timeout fires, idle timer resets on each chunk, env-disable preserves observer, env override. |

## Verification

```
bun run typecheck    ✓  (0 errors)
bun run lint         ✓  (60 pre-existing warnings, 0 errors, 0 new)
bun run format:check ✓
bun run test         ✓  (5636/5636 pass, +7 new tests over #406's 5629 baseline)
```

## Review notes

- `src/run/codex-fetch-observer.ts` is the primary read. Focus on the `AbortSignal.any` composition (verify the caller signal is never dropped) and the `Promise.race` idle loop (verify chunks reset the deadline and that the abort promise does not hold a reference preventing GC after a clean stream end).
- The `cause=` field is how operators will distinguish a TTFB stall from a mid-stream stall in `typeclaw logs -f` going forward.
- This is a mitigation, not a root-cause fix. The upstream stall behavior (server-side queue depth, Cloudflare PoP routing, worker assignment delays) is beyond our reach. Background context: pi-mono issue #4945 (the stall trace that motivated #406), openai/codex issue #23807 (TTFB split proposal for the Rust CLI).

## Related

- #394 — original 9-minute production hang that prompted investigation
- #406 — observer PR this builds on; introduced `[codex-fetch]` logging and diagnosed the stall phase

### DIFF
--- a/src/run/codex-fetch-observer.test.ts
+++ b/src/run/codex-fetch-observer.test.ts
@@ -1,6 +1,40 @@
 import { afterEach, beforeEach, describe, expect, test } from 'bun:test'
 
-import { installCodexFetchObserver } from './codex-fetch-observer'
+import { installCodexFetchObserver, type TimeoutScheduler } from './codex-fetch-observer'
+
+// Programmable scheduler for tests. Tasks scheduled for time T fire when the
+// test calls `fire(T)`. Out-of-order T values (T must be monotonically
+// non-decreasing across calls) are not supported; tests advance time forward.
+function makeScheduler(): TimeoutScheduler & { fire: (t: number) => Promise<void>; pending: () => number } {
+  type Entry = { id: number; dueAt: number; cb: () => void; cancelled: boolean }
+  let counter = 0
+  let nowMs = 0
+  const entries: Entry[] = []
+  return {
+    set: (delayMs, cb) => {
+      const e: Entry = { id: ++counter, dueAt: nowMs + delayMs, cb, cancelled: false }
+      entries.push(e)
+      return e
+    },
+    clear: (handle) => {
+      const e = handle as Entry
+      e.cancelled = true
+    },
+    fire: async (t) => {
+      nowMs = t
+      const flush = () => new Promise<void>((r) => queueMicrotask(r))
+      while (true) {
+        const due = entries.find((e) => !e.cancelled && e.dueAt <= t)
+        if (!due) break
+        due.cancelled = true
+        due.cb()
+        await flush()
+        await flush()
+      }
+    },
+    pending: () => entries.filter((e) => !e.cancelled).length,
+  }
+}
 
 type LogEntry = { level: 'info' | 'warn'; msg: string }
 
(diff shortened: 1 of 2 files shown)


devxoul:
@typeey yo what do you think of this PR

PR #416, opened 2026-05-27
cli: capture full multi-line PEM when pasting a github app key

devxoul:
## Problem

When an operator pastes a real GitHub App private key into the auth prompt during `typeclaw init` or `typeclaw channel`, only the first line was captured. The prompt used `text()` from `@clack/prompts`, which treats the first newline in the paste as Enter and submits immediately. The captured value was `-----BEGIN RSA PRIVATE KEY-----` — the header line — which the downstream resolver treated as a filesystem path and threw ENOENT.

## What this PR does

**New file `src/cli/prompt-pem.ts`** exports `promptPrivateKeyPem(message)` and a pure helper `readPrivateKeyFromLines(readLine)`.

- Uses `node:readline` to read stdin line by line instead of clack's char-buffered `text()`.
- If the first non-empty line does not start with `-----BEGIN`, returns it as-is (single-line path or escaped PEM).
- If the first line contains both `BEGIN` and `END` markers, treats it as an inline escaped PEM and returns.
- Otherwise enters "block mode" and accumulates lines until a line matches the END marker pattern.
- Handles CRLF (strips trailing `\r`), skips leading blank lines (stray Enter before paste).
- Returns `CANCEL_SYMBOL` on stdin close.

`resolvePrivateKeyInput()` then: real PEM → pass through; escaped single-line with literal `\n` → unescape; otherwise → read from disk.

**`src/cli/channel.ts` and `src/cli/init.ts`** both call `promptPrivateKeyPem()` instead of `text()` and drop the now-dead private resolver helpers and the `readFile` import.

## Tests

New file `src/cli/prompt-pem.test.ts` — 16 tests covering:

- Single-line path passthrough
- Single-line escaped PEM passthrough
- Multi-line PEM block capture (the bug)
- Leading blank lines before paste
- RSA, EC, and generic PKCS#8 `PRIVATE KEY` headers
- CRLF line endings
- Cancellation mid-block and pre-input
- `resolvePrivateKeyInput`: real PEM, escaped PEM, mixed multi-line + literal `\n` (no unescape), file-path read, ENOENT
- `createReadlineLineReader`: end-to-end with Readable stream + queueing

Full suite: 5680 pass, 0 fail. Typecheck clean. Lint: only pre-existing warnings on unrelated files.

## Files changed

| File | Change |
|---|---|
| `src/cli/prompt-pem.ts` | New. Line-buffered PEM prompt + resolver. |
| `src/cli/prompt-pem.test.ts` | New. 16 tests. |
| `src/cli/channel.ts` | +7 −19. Uses `promptPrivateKeyPem`, removes dead resolver helpers. |
| `src/cli/init.ts` | +4 −13. Same. |

## Review notes

- `readPrivateKeyFromLines` in `src/cli/prompt-pem.ts` is the primary read. The state machine has three branches (passthrough, inline-escaped, block mode); the block-terminator regex `/^-----END [A-Z0-9 ]*PRIVATE KEY-----\s*$/` gates multi-line accumulation.
- The leading-blank-line skip handles the common case where the operator hits Enter once before pasting; without it, the first "line" seen is an empty string and the state machine exits early.
- `resolvePrivateKeyInput` deliberately does not unescape when the input spans multiple real lines with literal `\n` — only single-line `\n`-escaped PEMs are unescaped. A test pins this behavior.

### DIFF
--- a/src/cli/channel.ts
+++ b/src/cli/channel.ts
@@ -1,5 +1,4 @@
 import { randomBytes } from 'node:crypto'
-import { readFile } from 'node:fs/promises'
 
 import { cancel, confirm, intro, isCancel, log, note, password, select, spinner, text } from '@clack/prompts'
 import { defineCommand } from 'citty'
@@ -27,6 +26,7 @@ import {
 import { runKakaotalkBootstrap } from '@/init/kakaotalk-auth'
 import { SecretsKakaoCredentialStore } from '@/secrets/kakao-store'
 
+import { CANCEL_SYMBOL, promptPrivateKeyPem } from './prompt-pem'
 import { c, done, errorLine, printDiscordInviteHint, printSlackAppManifestSetup } from './ui'
 
 const CHANNEL_LABELS: Record<ChannelKind, string> = {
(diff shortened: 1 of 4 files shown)


devxoul:
@typeey can you review again

PR #425, opened 2026-05-27
Add typeclaw update command

synchrok:
## Summary
- add a host-stage `typeclaw update` command
- support Bun and npm global self-update commands, with conservative auto-detection
- document the command in the CLI reference

## Tests
- `bun run format`
- `bun run typecheck`
- `bun run lint`
- `bun test --parallel src/update/index.test.ts src/cli/index.test.ts`
- `git diff --check --cached`

## Notes
- Real registry mutation was intentionally not tested; covered via `--dry-run` and planner tests.


devxoul (approved):
LGTM thanks!

PR #447, opened 2026-05-28
channels: structured inbound attachment metadata

devxoul:
## Summary

Screenshots and images sent by users on KakaoTalk, Slack, Discord, and Telegram have been a recurring source of hallucinations: the agent either invented attachment content it could not see, described an image as "empty", or fabricated a platform ref (e.g. pasted a bare KakaoCDN key into the wrong tool field). The root cause was structural — the opaque platform ref (`file_id`, CDN URL, CDN key) was embedded directly in the prompt-visible placeholder text, so the LLM saw it, learned to echo it, and occasionally confabulated variations of it.

This PR fixes the problem at the data model level: attachments are now carried as structured metadata alongside the prompt text, the ref stays out of the LLM context entirely, and the agent uses an in-turn id to request a specific attachment through a router-validated lookup.

### `src/channels/types.ts` — new `InboundAttachment` type

New `InboundAttachment` shape with `id`, `kind`, `ref`, and optional metadata fields (`mimetype`, `filename`, `width`, `height`, `sizeBytes`). Key design contracts:

- `ref` is the opaque platform handle (Slack file id, Discord CDN URL, KakaoCDN key, Telegram `file_id`). It is intentionally **not** rendered into the prompt — keeping it out of the LLM's context is the primary hallucination guard.
- `id` is a 1-based in-turn index assigned by the adapter classifier. Not globally unique; the router qualifies lookups by session + chat + author so cross-message collisions cannot occur.
- `InboundMessage.attachments` and `ChannelHistoryMessage.attachments` carry the array through the pipeline.

### `src/channels/router.ts` — attachment lookup API + backward-compat regex

Two new methods on `ChannelRouter`:

- `lookupInboundAttachment({ adapter, workspace, chat, thread, id })` — walks the live session's `promptQueue` and `contextBuffer` and returns the matching `InboundAttachment`, or `null` if the id is not present in this turn.
- `listInboundAttachmentIds({ adapter, workspace, chat, thread })` — returns the sorted set of ids present in the current turn, used in error messages when the agent passes a nonexistent id.

`stripChannelMediaPlaceholders` updated to match both the new `[<Platform> attachment #N: ...]` format and the legacy `[<Platform> message with ...]` format — no migration needed for persisted history.

### `src/agent/tools/channel-fetch-attachment.ts` — `ref` → `attachment_id`

The tool's input schema replaces `ref: string` with `attachment_id: integer`. The description is rewritten to make explicit that the id must appear in the current inbound message. The handler:

1. Calls `router.lookupInboundAttachment` with the session's `ChannelKey` + the requested id.
2. Returns a structured error if the id is not found, listing all valid ids for this turn (hallucination guard).
3. Returns a separate error for stickers and other non-fetchable kinds with an empty ref (empty-ref guard).
4. Resolves `ref` from the lookup result — never from the prompt.
5. Falls back to `found.filename` when the caller omits the `filename` parameter.

`normalizeRef` retains the `id=` prefix-stripping for backward compat with any persisted refs from old Slack classifiers.

### Adapter classifiers — 4 platforms

Each classifier now populates `InboundMessage.attachments` with typed `InboundAttachment` objects and renders the prompt placeholder as `[<Platform> attachment #N: <kind> <metadata>]` (without the ref):

| File | Platform |
|---|---|
| `src/channels/adapters/kakaotalk-attachment.ts` | KakaoTalk |
| `src/channels/adapters/slack-bot-classify.ts` | Slack |
| `src/channels/adapters/discord-bot-classify.ts` | Discord |
| `src/channels/adapters/telegram-bot-classify.ts` | Telegram |

### `src/skills/typeclaw-channel-kakaotalk/SKILL.md`, `src/skills/typeclaw-channel-telegram-bot/SKILL.md`

Updated to document the `attachment_id` parameter and the new placeholder format, so skills that handle inbound media instruct the agent correctly.

## Backward compatibility

`stripChannelMediaPlaceholders` now matches both placeholder shapes:

```
[KakaoTalk attachment #1: photo 1254×1254]    ← new
[KakaoTalk message with photo 1254x1254 ...]  ← legacy
```

No migration is needed. Existing persisted history with old-style placeholders continues to be stripped correctly in quote anchors.

## Testing

- **Hallucination guard** — `channel_fetch_attachment` tests assert that calling with an id not present in the current turn returns an error listing the valid ids, rather than attempting a fetch. The agent cannot invent an id and proceed.
- **Empty-ref guard** — sticker and no-URL attachment objects with `ref: ''` return a descriptive error directing the agent to acknowledge without promising to view the file.
- **Structured attachment tests** — each adapter classifier test suite covers production of `InboundAttachment` objects with correct `id`, `kind`, `ref`, and metadata fields. Router fake updated in `channel-reply.test.ts`, `channel-send.test.ts`, `skip-response.test.ts`, and `channel-tool-logging.test.ts` for the new `lookupInboundAttachment` / `listInboundAttachmentIds` surface.
- **`stripChannelMediaPlaceholders`** — existing test coverage passes for both old and new placeholder shapes (regex is backward-compatible, no new test file needed).

## Checks

```
bun run lint         ✓  60 pre-existing warnings in dockerfile.ts, 0 errors
bun run format       ✓  clean
bun run format:check ✓  clean
bun run test         5703 pass, 1 fail
```

The single failure is `resolveSelfPackageJsonPath > points at this package manifest` in `src/update/index.test.ts` — a pre-existing environment/path issue where this worktree lives at a path not ending in `typeclaw/package.json`. Unrelated to this PR; passes on CI.

## Review notes

- The load-bearing change is the `ref`-exclusion contract in `InboundAttachment`. If `ref` ever leaks back into the prompt placeholder, the hallucination regresses. The type comment in `types.ts` documents this explicitly.
- `lookupInboundAttachment` searches both `promptQueue` and `contextBuffer` so attachments remain resolvable even after the inbound has been dequeued and committed to context.
- The `attachment_id` minimum is `1` (schema enforced) — the agent cannot pass zero or a negative id.
- `normalizeRef`'s `id=` stripping is intentionally kept for any Slack refs still persisted in the old format.


devxoul:
Thanks for the careful read — all three notes addressed in three small commits on top of the existing branch.

1. **`normalizeRef` shared between the two tools** → e1bb655 extracts `normalizeRef` to `src/agent/tools/normalize-ref.ts` and wires both `channel_fetch_attachment` and `look_at_channel_attachment` through it. Added a focused unit-test file for the helper itself so the symmetry is locked in regardless of which call site evolves.

2. **Restored `id=` prefix test** → ea3063c adds back the dropped test at the new boundary: now the tool first looks up the attachment, then strips `id=` from `found.ref` before handing it to the adapter callback. Asserts on the resolved `{ ref: 'FABC123' }` to pin the contract.

3. **Comment overselling collision protection** → 8c33a72 fixes both the comment AND the underlying behavior. The lookup now walks newest-first through promptQueue then contextBuffer, so when an id collides across messages in the same session, `attachment_id: 1` always resolves to the **current** turn's attachment rather than a stale one from earlier context. The type-contract comment rewritten to describe the actual guarantee (most-recent-wins, scoped to one session+chat+thread).

Verification: `bun run typecheck` / `bun run lint` / `bun run format:check` all clean. `bun run test`: 5709 pass / 1 pre-existing unrelated failure (`resolveSelfPackageJsonPath` — the worktree lives at `/tmp/typeclaw-inbound-attachments` not `…/typeclaw/`).

Issue #452, opened 2026-05-28
Subagent bash allowlist (per-subagent binary/command restriction)


devxoul:
## Motivation

The new `reviewer` subagent (introduced in this work) is a read-only deep-analysis specialist. It is given `bashTool` so it can run `gh pr diff`, `git log`, `git diff`, `git blame`, `gh api -X GET ...`, etc. Today there is no runtime mechanism to constrain bash to those commands — the reviewer's read-only-ness is currently enforced only by its system prompt, the same way `explorer` does it.

This is fine for now, but a defense-in-depth mechanism for per-subagent bash restriction would be valuable as more read-only-with-shell subagents are added.

## Background

From a recent exploration of the codebase:

- pi-coding-agent's bash tool (`@mariozechner/pi-coding-agent/dist/core/tools/bash.js`) exposes `BashToolOptions` with `operations`, `commandPrefix`, and `spawnHook` — **but no allowlist/filter option**.
- TypeClaw's existing bash guards (`secret-exfil-bash`, `git-exfil`) are global denylists, not per-subagent allowlists.
- The `Subagent` / `SubagentShared` type has no per-tool configuration field.
- Plugin `tool.before` hooks are registered globally, not per-subagent.

So adding per-subagent bash restriction is a real architectural question. There is no existing surface to extend.

### Path 1 — Custom `bash` tool inside a subagent plugin
The subagent contributes a `customTool` named `bash` that parses the command line, checks an inline allowlist (e.g. `git`, `gh pr diff`, `gh api -X GET`, …), and delegates to `createLocalBashOperations().exec()` from pi-coding-agent. Pros: zero core changes, isolated. Cons: allowlist lives in plugin code.

### Path 2 — Plugin-side `tool.before` hook gated by `origin.subagent === <name>`
The subagent's plugin registers a global `tool.before` hook that only enforces when the calling subagent matches. Pros: lighter than custom tool. Cons: origin-name string-matching is mildly fragile.

### Path 3 — First-class `bashAllowlist` on `SubagentShared`
Add `bashAllowlist?: string[]` to `SubagentShared`, thread it through the plugin-shim layer (`src/run/index.ts` `PLUGIN_ONLY_KEYS_DROPPED_BY_SHIM`), `createSessionForSubagent`, and inject enforcement into the tool wrapper. Pros: generic, declarative, type-safe. Cons: touches core; only earns its keep when two+ subagents need it.

## Recommendation

[...]

PR #464, opened 2026-05-29
agent: sandbox subagent bash via bwrap and apply to reviewer

devxoul:
## Summary

Part 2 of 2 for [#452](https://github.com/typeclaw/typeclaw/issues/452). Builds on the bwrap-availability infrastructure landed in #463 (bubblewrap install + `--security-opt seccomp=unconfined`). Depends on #463; do not merge before it.

This PR wires per-subagent bwrap sandboxing end-to-end and applies it to the `reviewer` subagent as the first consumer.

### `src/agent/subagents.ts` (+41)

New `SandboxOptions` type and `sandbox` field on `SubagentShared` (plugin-API surface). Fields: `mounts[]` (ro/rw bind pairs), `network` (`'none'` or `'loopback'`), `envPassthrough[]`, `allowlist[]` (command prefix strings), `cwd`.

### `src/agent/sandbox.ts` (new, +197)

`applySubagentSandbox(tool, origin, command, getSubagentByName)` — the bwrap rewriter. Short-circuits for non-bash tools, non-subagent origins, and subagents without a `sandbox` declaration. When triggered:

1. **FORBIDDEN-metachar gate** — rejects shell metacharacters (`;`, `|`, `&`, `` ` ``, `$()`, `${}`) that would bypass prefix-match argv allowlisting via command substitution. Without this gate, prefix-match allowlisting is theater.
2. **Allowlist check** — rejects commands not matching a declared prefix (token-boundary match, not substring).
3. **`ensureBwrapAvailable()`** — fail-closes with `SandboxUnavailableError` if the binary is missing.
4. **bwrap argv construction** — strict rootfs (`--ro-bind /usr`, `--tmpfs /proc`, `--clearenv`); no network unless opted in; declared mounts applied; shell quoting with single-quote edge cases handled.

### `src/agent/sandbox.test.ts` (new, +365)

30 unit tests: no-op cases, allowlist rejection (prefix-substring vs prefix-with-space, whitespace normalization), FORBIDDEN-metachar gate (every dangerous shell char including `echo "$(rm -rf /)"`), bwrap argv shape (mounts, network, envPassthrough, cwd, quoting), and the fail-closed `SandboxUnavailableError` path. `test.skipIf(!bwrapAvailable)` gates bwrap-positive tests; `test.skipIf(bwrapAvailable)` gates the unavailable path — every test runs on at least one side of host/container.

### `src/agent/plugin-tools.ts` (+41)

Calls `applySubagentSandbox` in all three execute pipelines (`wrapSystemTool`, `wrapSystemAgentTool`, `wrapAgentToolAsCustomToolDefinition`) right after `stripGuardAcknowledgements` and right before `tool.execute`. Order is load-bearing: existing `tool.before` hooks (`secret-exfil-bash`, `git-exfil`) see the original command; bwrap rewrites the command between guards and exec.

### `src/agent/index.ts` (+13)

Threads `getSubagentByName` through `WrapSystemToolOptions` and `PluginSessionWiring`.

### `src/run/index.ts` (+6)

`createSessionForSubagent` supplies `getSubagentByName` by reading `pluginRuntime.get().subagents[name]` lazily inside a closure — snapshot is re-read on every call so plugin reload is visible to in-flight tool calls.

### `src/bundled-plugins/reviewer/reviewer.ts` (+53), `skills/code-review.ts` (+9), `src/agent/system-prompt.ts` (+2 / -1)

Applies `sandbox` to the `reviewer` subagent:

- `network: 'none'`
- `mounts`: ro bind of `/agent/.git` at `/work/.git`
- `allowlist`: local git subcommands + pipeline tools (`cat`, `head`, `tail`, `wc`, `sort`, `uniq`, `jq`, `yq`)
- `gh` intentionally excluded (see Security decisions below)

Parent's `system-prompt.ts` reviewer-delegation guidance updated to match. `reviewer.test.ts` drift guard updated. Removes `TODO(#452)`.

## Security decisions

**`gh` excluded from the reviewer allowlist.** A broad `GH_TOKEN` (the default for any developer who has run `gh auth login`) gives `gh api -X GET` read access to every private repo, gist, workflow log, and org the user has — far beyond the PR being reviewed. Prefix-matching on `gh api -X GET` is bypassable: `gh` accepts multiple `--method` flags and last-wins, so a prompt-injected `--method POST` appended after a compliant-looking prefix defeats the guard entirely.

**Parent fetches PR content, not reviewer.** `gh pr diff` / `gh pr view` are called by the parent agent, which embeds the result inline in the reviewer's prompt. Matches the "subagent receives content, does not fetch content" convention already used by `scout`/`explorer`/`operator`.

**FORBIDDEN-metachar gate is defense-in-depth.** Prefix-match allowlisting without the gate is bypassable via `echo "$(rm -rf /)"`. The gate makes allowlist rejection structurally meaningful.

## Testing

- 30 unit tests in `src/agent/sandbox.test.ts` — dual host+container coverage via `test.skipIf`.
- Verified end-to-end in `typeclaw-servant:latest` container with bwrap installed: all 124 reviewer+sandbox+plugin-tools+system-prompt tests pass.

```
bun run lint         ✓  0 errors
bun run format       ✓  clean
bun run test         ✓  5875 pass, 9 skip, pre-existing failures (resolveSelfPackageJsonPath worktree-name, flaky 30s plugin-loader timeouts — verified pre-existing by stashing against branch base)
```

## Review notes

- **`src/agent/sandbox.ts` `applySubagentSandbox`** is the load-bearing new code. The FORBIDDEN-metachar gate and the token-boundary allowlist check are the key correctness properties.
- **Call order in `src/agent/plugin-tools.ts`** is the security invariant: guards first, bwrap second, exec third. Any reordering breaks the defense-in-depth layering.
- **`gh` exclusion** is intentional and documented above. Do not re-add it to the allowlist without a scoped token story.
- **`--tmpfs /proc`** (not `--proc /proc`): required for OrbStack compatibility; `--dev-bind /proc /proc` is rejected because it leaks outer-container `/proc/N/environ` (including `FIREWORKS_API_KEY`) into sandboxed bash. Documented in `docs/content/docs/internals/sandbox.mdx` from #463.


devxoul:
Thanks for the review! The fail-open at `src/agent/sandbox.ts:58` is addressed in `b983241` (pushed after your review, which was against `9d21be4`).

The fix is exactly the split you suggested:

- **Registry miss → throws.** `if (subagent === undefined) throw new SandboxBlockedError("subagent \"…\" is not in the registry; refusing to execute bash for an unresolvable subagent origin")`
- **Found but no `sandbox` declared → returns** (legitimate no-op, unchanged).

So a subagent-origin bash call can no longer bypass the sandbox via a registry lookup failure — it fails closed. The comment block above the function documents the reasoning (typo / registry-rebuild race during plugin reload / missing entry).

`#463` is now merged, so the dependency is satisfied. Could you re-review against `b983241`? Checks are green (typecheck clean, lint 0 errors, 5875 pass — the one `resolveSelfPackageJsonPath` failure is the pre-existing worktree-name artifact noted in the PR description).

devxoul:
History rewritten (force-push) to fix a conceptual confusion at its root rather than in a follow-up commit.

**The bug:** the reviewer mounted the *agent nest* git repo (`/agent/.git` → `/work/.git`) and was given a `git` allowlist + prompt/skill text telling it to inspect "this repo" — but the agent nest is a *different repository* from the GitHub PR being reviewed. For a PR review the reviewer ended up with the agent's own config/memory/session history on disk (and tooling to spelunk it), while the actual PR content only arrived inline. Tooling that lies about its capability.

**The fix (folded into the reviewer commit, not added on top):**
- Removed the `/agent/.git` mount entirely — the reviewer has no repository on disk.
- Removed all `git` from the bash allowlist — pipeline tools only (`cat`/`head`/`tail`/`wc`/`sort`/`uniq`/`jq`/`yq`). No repo ⇒ no git, by design.
- Introduced a small `ReviewTarget` primitive (`kind: 'inline-diff'` today) so the reviewer's capabilities derive from what the parent actually provides. A future kind can stage the target tree and grant repo-scoped read tools without reworking this base.
- Corrected the prompt, code-review skill, and the parent reviewer-delegation guidance to the honest contract: "review the diff/content the parent embedded inline; you have no checkout and no GitHub access."

The bwrap-primitive commits are unchanged. Checks: typecheck/lint/format clean; `bun run test` 5879 pass, 9 skip, 1 pre-existing fail (`resolveSelfPackageJsonPath` worktree artifact).

devxoul:
Closing in favor of a re-architected approach. This PR couples the sandbox primitive too tightly to the `reviewer` subagent. The replacement builds a general-purpose, reusable bash sandbox primitive (with no subagent/origin coupling), designed so that all bash execution can eventually be sandboxed by default. Reviewer remains the intended first consumer, but the primitive's API will not assume it. Re-evaluating the security model from scratch as part of the redesign.

PR #472, opened 2026-05-29
reviewer: stage PR-head tree for deep cross-file review, credential-blind

devxoul:
Stacked on #464 (base: `feat/sandbox-tool`). Merge after #464.

## Problem

The #464 sandboxed reviewer sees only the inline diff plus `/agent/.git` — which is the **agent nest's** own repo, **not the review-target project**. So it cannot trace imports, callers, base classes, or hidden implementations across the PR codebase. PR-surface review is too shallow.

## Approach: parent prepares, reviewer reads, credential-blind

- **Parent stages, reviewer reads.** Before the reviewer LLM starts, its handler fetches the PR head as a GitHub **tarball** (token used only in that fetch) into a per-review temp dir and mounts it **read-only at `/work`** via a per-session sandbox override. The reviewer reads/greps/traces the full tree offline.
- **Token never enters the jail.** The bash sandbox keeps `--clearenv` + no network; `gh` stays out of the allowlist. Deep review is gained **without** reopening the exfil surface — credential and network stay entirely on the parent side.
- **Tarball, not clone.** A tarball extract runs **no** git hooks / filters / LFS smudge / submodule fetch — none of the attacker-controlled code paths a `git clone`/checkout would trigger.
- **Per-session isolation.** Concurrent reviews of different PRs get isolated mounts via `sandboxOverride` (no shared-registry mutation), so they don't race on `workspace/`.

## Commits

1. `agent: structured spawn payload and per-session sandbox override` — `spawn_subagent` gains a typed `payload`; reserved keys spread last (prompt-injection defense); `RunSession.sandbox` + `overlaySandboxOverride` thread a per-spawn sandbox without mutating the shared registry.
2. `reviewer: add credential-free safe tarball staging` — dependency-free tar.gz extractor that **fails closed** on symlinks, hardlinks, absolute paths, and `..` traversal (the reviewer's read/grep/find/ls tools are **not** in the bwrap jail, so a malicious archive symlink could otherwise escape the staged tree); dispose-on-throw, no leak.
3. `reviewer: mount staged PR tree at /work, drop the /agent/.git mount` — handler stages + mounts; `sandbox.test.ts` asserts the staged-mount argv shape (ro-bind at `/work`, clearenv, no share-net).
4. `reviewer: teach the parent-stages / reviewer-reads-/work contract` — system-prompt, code-review skill, github channel skill updated; spawn with `payload.reviewTarget = { kind: 'github-pr', owner, repo, pullNumber, headSha }`.

## Security notes

- **Token isolation:** `GH_TOKEN` is used only in the handler fetch; never written to payload, prompt, staged files, or sandbox env. Reviewer declares no `envPassthrough`.
- **Symlink/traversal gate** is load-bearing because only `bash` is bwrap-jailed; the file tools are not. Safe extraction is the boundary that keeps them in-tree.
- **No shared caches** mounted into the jail.

## Deferred

Opt-in **execution** of repo code (`reviewer.execute`: run tests/build in the same `--clearenv`+`--unshare-net` jail) is intentionally deferred to a later PR. This PR is read-only deep review, which solves the core shallow-review bug without arbitrary-code-execution surface.

## Testing

```
bun run typecheck   0 errors
bun run lint        0 errors (63 pre-existing warnings)
bun run format      clean
bun run test        5904 pass, 10 skip (bwrap-positive tests run container-side via test.skipIf), 1 pre-existing fail (resolveSelfPackageJsonPath worktree-name artifact, documented on #464)
```
28 new tests across spawn-subagent, subagents (overlay + threading), staging (safe extraction + token-non-leak + dispose), reviewer (reviewTarget + sandbox contract), and sandbox (staged-mount argv shape).


devxoul:
Self-review pass (with Oracle) found and fixed a **blocker** plus extractor hardening — pushed in `8b33e67`.

**Blocker — closed the read+exfil hole.** The reviewer still had unsandboxed host file tools (`read`/`grep`/`find`/`ls`) and network tools (`webfetch`/`websearch`), but `applySubagentSandbox` only jails `bash`. Since the reviewer reads attacker-controlled PR content, a prompt injection could `read` a secret *outside* the staged tree (`/agent/.env`, `/proc/self/environ`) and exfiltrate it via `webfetch` — defeating the `--clearenv`/no-network jail. **Fix:** the reviewer's only tools are now sandboxed `bash` + `load_skill`; all filesystem access is funneled through the `/work` jail, so there is no unsandboxed reach and no network sink. Non-PR reviews work from inline content the parent embeds (scoped file access for that path deferred to the execution PR).

**Extractor hardening (staging.ts):**
- compressed + decompressed size caps (zip-bomb / OOM)
- reject truncated entries (declared size past buffer)
- reject raw `..` before top-level stripping (invariant completeness)
- reject GNU base-256 large-size encoding (mis-parse → corruption)

**Verification:** typecheck/lint/format clean; `bun run test` 5907 pass, 10 skip, 1 pre-existing fail (`resolveSelfPackageJsonPath` worktree artifact). New tests cover each gate + the bash-only tool contract.

devxoul:
Closing in favor of a clean rebuild. This PR was built on the pre-correction #464 base (which mounted the agent nest .git as the review target) and carried tarball-staging + bash-only churn from an exploratory iteration. #464 has since been corrected at the root (ReviewTarget primitive, no nest-git confusion). The deep-review capability will be reopened as a fresh, minimal PR adding a `staged-tree` ReviewTarget kind on top of the corrected base.

PR #478, opened 2026-05-29
agent: drop Korean weekday from the per-turn time anchor

devxoul:
## Background

The per-turn `<current-time>` anchor injected into every user turn pre-computes the weekday name so the model can answer "what day is it?" without doing error-prone weekday-from-ISO arithmetic. That rationale is language-agnostic — but the prior implementation emitted *two* names, English and Korean:

```
<current-time>2026-01-15T12:00:00+09:00 (Asia/Seoul, Thursday / 목요일)</current-time>
```

The Korean half was a leftover from an early locale assumption that TypeClaw's users were Korean-first. TypeClaw's users are global. Hardcoding a single second language is not justified by the rationale; reply language belongs to each agent's `SOUL.md`, not a runtime-injected anchor.

## Summary

Drops the Korean weekday name from `renderTurnTimeAnchor`. The anchor now emits one canonical English weekday:

```
<current-time>2026-01-15T12:00:00+09:00 (Asia/Seoul, Thursday)</current-time>
```

`formatLocalWeekday` is simplified to return a plain string instead of an `{ en, ko }` pair.
The `LocalWeekday` type and the `WEEKDAYS_KO` fallback table are removed along with it.
The re-export of the now-deleted type from `src/shared/index.ts` is dropped.

## What this does NOT do

- Does not remove the weekday from the anchor or change its purpose — the pre-computation rationale is unchanged.
- Does not change any agent's reply language — that stays in each agent's `SOUL.md`.
- Does not touch any other locale or I18n surface.

## Review notes

- Load-bearing files: `src/shared/local-time.ts` (the `formatLocalWeekday` function) and `src/agent/system-prompt.ts` (`renderTurnTimeAnchor`). The comment block in both has been updated with the English-only/global-users rationale — verify the two stay consistent.
- The `system-prompt.test.ts` change inverts the Korean-presence contract: the test now asserts that *no* Korean weekday appears in the anchor, not merely that an English one does. Tightened, not just updated.
- The `router.test.ts` delta is one deleted line (the Korean-present assertion); the English-present assertion is kept.
- Pre-flight: `bun run typecheck` and `bun run lint` are clean. One pre-existing failure in `update/index.test.ts` reproduces identically on `origin/main` (a worktree-path artifact) and is unrelated to this change.

### DIFF
--- a/src/agent/system-prompt.test.ts
+++ b/src/agent/system-prompt.test.ts
@@ -5,7 +5,7 @@ import { formatLocalDateTime, resolveLocalTimezoneName } from '@/shared'
 import { renderTurnTimeAnchor } from './system-prompt'
 
 describe('renderTurnTimeAnchor', () => {
-  test('wraps the ISO timestamp, IANA zone, and weekday pair in a single <current-time> tag', () => {
+  test('wraps the ISO timestamp, IANA zone, and weekday in a single <current-time> tag', () => {
     const now = new Date('2026-01-15T12:00:00+09:00')
 
     const anchor = renderTurnTimeAnchor(now)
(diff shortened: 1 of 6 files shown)


devxoul:
@typeey review plz

PR #482, opened 2026-05-29
sandbox: wire per-subagent bash sandbox mechanism + trust-domain guidance

devxoul:
## Background

The bwrap sandbox primitive (`src/sandbox/`) shipped earlier but was never connected: `buildSandboxedCommand` was called only from its own tests, and `docs/internals/sandbox.mdx` flagged the tool-path wiring as "a separate follow-up." So no subagent bash call was sandboxed.

The original intent was to also make the `reviewer` subagent the first consumer. Working through the threat model — a prompt injection in a public PR diff steering the reviewer's bash to pivot into a private repo or secrets — showed that a per-tool sandbox is the **wrong layer** for that separation. Keeping a public-facing review away from private code and credentials is a coarse, whole-container concern, and making one agent safely span both trust domains would require staging, credential-stripped snapshots, output scrubbing, and publication gates: a large, fragile mechanism to defend a configuration users should not create. typeclaw already provides the right boundary — one agent is one container, one `.env`, one `GH_TOKEN`, one `/agent`.

So this PR ships the mechanism and documents the boundary, rather than building the cross-trust-domain machinery.

## Summary

- **Mechanism:** opt-in per-subagent bash sandbox. A subagent declares `sandboxPolicy` on its definition (`SubagentShared.sandboxPolicy`); when present, the `bash` builtin override in `wrapAgentToolAsCustomToolDefinition` rewrites the `command` arg through `buildSandboxedCommand` — **after** `tool.before` guards inspect the raw command, before pi executes it. Threaded through `CreateSessionOptions`, `defaultCreateSessionForSubagent`, and the plugin-subagent path in `startAgent`. Absence of a policy is the trusted default; capability-by-declaration, deliberately decoupled from `visibility`.
- **Guidance (docs):** scope the per-tool sandbox as defense-in-depth *within* one trust domain, and recommend **separate agents for separate trust domains** — do not point one agent at both a private repo and public-facing review work. This closes the public-to-private pivot by construction, and the subtler residual the sandbox cannot (a prompt-injected reviewer smuggling private source into review text posted to a public PR): with the domains in separate agents, that output channel has nothing private to leak.

**Why no bundled consumer yet:** the reviewer's own skills require repo-wide reads (callers, tests, conventions, duplicate-logic grep), so a `workspace/`-only sandbox would make it unable to follow its own methodology, and a full-`/agent` mount would re-expose the secrets the sandbox exists to exclude. Rather than build the staging apparatus to square that circle, the mechanism ships unconsumed for adoption, and the trust boundary is handled operationally.

## What this does NOT do

- Does not give any bundled subagent a `sandboxPolicy` (the reviewer is unchanged).
- Does not build per-session repo staging, snapshot sanitization, output scrubbing, or publication gates.
- Does not change `visibility`, permissions, provenance, or the container flags (`seccomp=unconfined`).

## Review notes

- Load-bearing ordering invariant: in `wrapAgentToolAsCustomToolDefinition` (`src/agent/plugin-tools.ts`), `applyBashSandbox` runs **after** the guard checks — so `secret-exfil-bash` / `git-exfil` inspect the operator-authored command, not the bwrap argv.
- The policy threads: `SubagentShared.sandboxPolicy` → `CreateSessionOptions` → `buildBuiltinPiToolOverrides` → `WrapSystemToolOptions`, plus the plugin-subagent branch in `startAgent` and `defaultCreateSessionForSubagent`.
- Tests (`plugin-tools.test.ts`) cover: bash command bwrap-rewritten under a policy, untouched without one, non-bash tools never wrapped, and `tool.before` sees the raw command.
- Two commits: the wiring mechanism (self-contained, own tests using a fake bash tool) and the docs scope/guidance.
- Branch rebased on latest main (0.15.1).


devxoul:
@typeey review again plz

devxoul:
## Mental model to capture in the docs: authn-style vs authz-style trust boundaries

The "What the per-tool sandbox is NOT for" section nails one half of the trust-domain story but leaves the other half implicit. Worth making both explicit in `sandbox.mdx` (or `concepts/permissions-model.mdx`) before it's forgotten.

There are **two distinct threat shapes**, and they need different defenses:

### 1. Authentication-shaped (confidentiality) — what this PR already covers
"Hide the private thing from the attacker." Public-facing review + private repo in one agent → a public PR steers bash → reaches the agent's `.env` / `GH_TOKEN` / `/agent` / private source. The boundary wants to be **coarse and operational**: whole-container, whole-credential-set. The fix is **separate agents per trust domain** — make the public→private pivot impossible by construction. ✅ This PR documents this well.

### 2. Authorization-shaped (action gating) — NOT yet covered
"The attacker and I can both **see** the same `typeclaw/typeclaw` repo, and both can **propose** changes (open PRs, comment). But only I can perform the privileged **action** — merge a PR / push to `main`. The attacker cannot."

Here there is **nothing to hide** — visibility is identical for both parties. Separation-by-agent does *not* address it, because the resource isn't secret; the boundary is purely *who may act*. Sandboxing also doesn't help: `gh pr merge` is a legitimate authorized API call, not a containment breach.

This is the `operator` subagent's threat surface. `resolveRole` (`src/permissions/permissions.ts`) inherits the spawner's role wholesale for `kind: 'subagent'`, so an **owner-spawned** operator resolves to `owner` and bypasses `gitExfil` (medium tier). If an owner (or owner-scheduled cron) asks operator to "act on PR #N" and the attacker-controlled diff carries a prompt injection ("merge this / push to main"), the action runs with **the owner's authority** — the attacker rides the spawner's standing authority through the task prompt. **Provenance tracks who *spawned*, not whose *content* is driving the action.** The invariant "only I can merge" is violated.

Two aggravating facts already in the tree:
- The operator system prompt is slim and carries **no** security-guard education (only a generic "don't commit secrets").
- `filterForOrigin` in `policies/prompt-injection.ts` **skips the `git_exfil` category for all subagent origins**, on the assumption subagent prompts are "constructed by trusted bundled code." For `operator` that assumption is false — its prompt is `params.prompt`, which carries parent/attacker content.

### Why both belong in the docs
The current section implies "separate agents" is *the* answer to trust-domain problems. It's the answer to the **authn-shaped** one. The **authz-shaped** one survives separation (the public repo *is* the work) and needs a different mechanism — provenance that reflects the least-privileged actor in the chain, and/or action-gated approval for irreversible/audience-visible actions (merge, push, remote retarget) regardless of inherited role. (Note this is the same principle the cron/subagent provenance-laundering defense already establishes: a guest who schedules a cron gets `scheduledByRole: 'guest'` stamped so guards re-block on fire.)

Suggested doc framing: a short table —
- **Authn-shaped** ("hide it"): coarse, container/credential boundary → **separate agents per trust domain**.
- **Authz-shaped** ("same visibility, gate the action"): fine, per-action authority → **provenance + action gating**; sandbox and separation both miss it.

(Design fix for the operator gap is being evaluated separately; this comment is just to pin the *mental model* in the docs.)

devxoul:
Captured in `docs/internals/sandbox.mdx` (commit `b85a6b4`). The "What the per-tool sandbox is NOT for" section now opens with a two-shape table and splits into two subsections:

- **Authentication-shaped** (confidentiality — "hide the private thing") → separate agents per trust domain.
- **Authorization-shaped** (action gating — "same visibility, only I may merge/push") → provenance + action gating; both separation and sandboxing miss it.

The authz subsection documents the `operator` surface explicitly: `spawnedByRole` tracks who *spawned*, not whose *content* drives the action, so an owner-spawned operator acting on an attacker-injected diff runs with owner authority — and it notes the two aggravating facts (slim operator prompt with no guard education; `filterForOrigin` skipping `git_exfil` for subagent origins). It cross-links Permissions for the provenance mechanism and the laundering-defense precedent.

As you noted, this only pins the mental model — the operator design fix is left to be evaluated separately and is called out as tracked elsewhere.

devxoul:
Closing without merging — superseded by an upcoming role-based filesystem-hiding sandbox (tracked in feat/role-based-sandbox).

The new approach hides sensitive paths at the bwrap layer by resolved role rather than per-subagent declared policy; mounts are otherwise unchanged. The authn/authz trust-domain docs from sandbox.mdx have been salvaged and will be carried forward into the replacement PR.

Nothing from this branch needs preserving.

devxoul:
Closing without merging — superseded by an upcoming role-based filesystem-hiding sandbox (tracked in feat/role-based-sandbox).

The new approach hides sensitive paths at the bwrap layer by resolved role rather than per-subagent declared policy; mounts are otherwise unchanged. The authn/authz trust-domain docs from sandbox.mdx have been salvaged and will be carried forward into the replacement PR.

Nothing from this branch needs preserving.

PR #493, opened 2026-05-29
feat(inspect): tag aborted turns distinctly from errors

devxoul:
## Background

`typeclaw inspect` rendered every assistant message carrying an `errorMessage` as a red `error` line — including user-cancelled turns, e.g. `22:29:09  error      Request was aborted.`. Root cause: when an assistant `message_end` is converted to an inspect `error` event (both the replay JSONL path and the live websocket path), the upstream `stopReason` was dropped. `"Request was aborted."` is the verbatim `errorMessage` that pi-coding-agent emits with `stopReason: 'aborted'` on an Escape/cancel — not a provider failure. Without `stopReason`, the renderer could not tell a real error from an abort, so an intentional cancel looked like a crash. `src/agent/provider-error.ts` already treats `aborted` as "not a provider failure" for channels/cron; this brings inspect in line.

## Summary

- Plumb `stopReason`: add an optional `stopReason` to the `error` InspectEvent and thread it through both producers — replay.ts (transcript) and live.ts (websocket tail). The server's message_end payload already carried `stopReason`; live just stops discarding it.
- Render aborts distinctly: `stopReason === 'aborted'` now renders as a dim-yellow `abort` tag instead of a red `error`. Genuine errors that carry a non-abort stopReason get a `(stop=<reason>)` suffix. Why a new `abort` tag rather than only recoloring: the tag is what a reader scans first; keeping the word `error` for cancels is the source of the confusion.
- After: `22:29:09  abort      Request was aborted.` and `22:29:09  error      provider returned 503 (stop=error)`.

## What this does NOT do

- Does not change the underlying `errorMessage` text (produced upstream by pi-coding-agent; inspect only surfaces it).
- Does not add a new filter category — aborts keep `cat: 'error'`, so `--filter !error` still hides them; presentation only.
- Does not touch the pre-existing no-op assistant-text-with-errorMessage branch in live.ts (out of scope).

## Review notes

- Load-bearing change: the `stopReason` field on the `error` variant in src/inspect/types.ts and the matching branches in renderTag/renderBody in src/inspect/render.ts.
- Invariant: an aborted turn renders the `abort` tag and omits the `(stop=)` suffix; a non-abort error keeps the red `error` tag and gains the suffix. Covered by new tests in render.test.ts, replay.test.ts, live.test.ts.

### DIFF
--- a/src/inspect/live.test.ts
+++ b/src/inspect/live.test.ts
@@ -135,6 +135,42 @@ describe('streamLive — live session events', () => {
     expect(assist.model).toBe('m')
   })
 
+  test('aborted message_end surfaces an error event carrying stopReason', async () => {
+    const registry = new LiveSessionRegistry()
+    const session = createFakeAgent()
+    registry.register({ sessionId: 'ses_a', session })
+    const { url } = await startServer({ registry })
+
+    const ctrl = new AbortController()
+    const subscribed = Promise.withResolvers<void>()
+    const gen = streamLive({
+      url,
+      sessionId: 'ses_a',
+      signal: ctrl.signal,
+      onSubscribed: () => subscribed.resolve(),
+    })
+
+    void subscribed.promise.then(() => {
+      session.emit({
+        type: 'message_end',
+        message: {
+          role: 'assistant',
+          content: [],
+          stopReason: 'aborted',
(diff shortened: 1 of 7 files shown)


devxoul:
@typeey why you didn't put the label

PR #495, opened 2026-05-29
fix(security): resolve symlinks in privateSurfaceRead before matching

devxoul:
## Security blocker (found in pre-release review)

`privateSurfaceRead` — the non-bash enforcement point for the role-based filesystem deny-list (#488) — resolved candidate paths **lexically** with `path.resolve`, so it never followed symlinks.

### The bypass

A restricted role (e.g. `guest`) can, via sandboxed bash, plant a symlink inside the guest-visible `public/` dir (#492):

```
public/leak -> ../.env        # or ../memory, ../workspace, ../sessions
```

Then read it back through a **non-bash** tool — `read`, `look_at`, or a `channel_reply` attachment — with `path: "public/leak"`. The guard resolved that lexically to `<agent>/public/leak`, saw it land in the allowed `public/`, and **let it through**. The actual file read followed the symlink to `.env`. The deny-list was bypassed entirely.

The bash sandbox (bwrap masks) caught this; the file-tool enforcement point did not — breaking the "two enforcement points, one deny-list" invariant #488 depends on.

Demonstrated divergence (lexical vs real):
```
lexical resolve -> <agent>/public/env-link   (lands in public/, allowed)
realpath        -> <agent>/.env              (the secret, blocked)
```

## Fix

Resolve the candidate's **real** path before matching: follow symlinks on every existing path component, walking up to the nearest existing ancestor for not-yet-created targets (so `public/leak/x` is caught even if `x` doesn't exist). Both sides are realpath'd because the agent dir itself may sit under a symlink (e.g. `/tmp -> /private/tmp` on macOS) — comparing a real candidate against a lexical deny-list would never match.

Kept **synchronous** via `realpathSync.native` so the security `tool.before` check array stays non-async (one syscall per existing component, negligible at the tool-call boundary). This mirrors the existing `resolveRealIntendedPath` pattern in the guard plugin's `non-workspace-write` policy.

## Tests

- public/ symlink to a hidden **file** → blocked
- read **through** a public/ symlink to a hidden **dir** → blocked
- symlink via a **nested** arg shape (`look_at` `images[].path`) → blocked
- a genuine non-symlink file in `public/` → still allowed

## Release impact

This is one of two blockers gating the sandbox release. The second (false-positive over-blocking) is in a stacked follow-up PR.

## Test plan

`bun run typecheck`, `bun run lint` (0 errors), `bun run format:check` clean. Security + sandbox suites: 516 pass, 0 fail.

> Two pre-existing full-suite failures (`resolveSelfPackageJsonPath`, Xvfb entrypoint-shim) are environment/worktree-path artifacts in files this PR does not touch.

### DIFF
--- a/src/bundled-plugins/security/policies/private-surface-read.test.ts
+++ b/src/bundled-plugins/security/policies/private-surface-read.test.ts
@@ -1,4 +1,8 @@
 import { describe, expect, test } from 'bun:test'
+import { mkdirSync, mkdtempSync, symlinkSync, writeFileSync } from 'node:fs'
+import { realpathSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import path from 'node:path'
 
 import type { HiddenPaths } from '@/sandbox'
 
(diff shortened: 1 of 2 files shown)


devxoul:
@typeey where are the reviews

devxoul:
@typeey can you comment again?

devxoul:
@typeey I cannot see. Please review again

PR #497, opened 2026-05-29
fix(memory): run memory infrastructure as system, not the triggering role

devxoul:
## Regression (found in pre-release review)

The role-based sandbox (#488) hides `sessions/`, `memory/`, `workspace/` from the `guest` role. The memory infrastructure subagents — `memory-logger` and `memory-retrieval` — inherited the **spawning turn's role** via `spawnedByOrigin: <parent origin>`. When a **guest channel turn** triggered them, they resolved to `guest`, and `privateSurfaceRead` then blocked their legitimate access:

- `memory-logger` reading the parent transcript under `sessions/`
- `memory-retrieval` writing `memory/.retrieval-cache/<id>.md`

Result: **guest channel turns silently fell out of memory** — no error surfaced to the operator, the conversation just stopped being logged/retrieved. Silent data loss.

## Root cause

The bug is not "the subagent uses a guarded tool." It is that the runtime **classified TypeClaw-owned memory infrastructure as the guest actor**. Memory logging is infrastructure that fires after *every* turn regardless of who spoke; it acts on the operator's behalf over operator-owned state, not as user-delegated work. The `backup` plugin already recognized this — it spawns its runner with a synthetic TUI origin so it resolves to `owner`.

## Fix: a first-class `system` origin

Rather than copy backup's synthetic-TUI hack (which lies in audit trails — a guest-triggered job recorded as "TUI-originated"), this adds a proper `{ kind: 'system'; component; triggeredBy? }` session origin:

- **Resolves to `owner`** in the permission service.
- **Honest provenance**: `triggeredBy` records what triggered it (e.g. the guest channel turn) for audit, without claiming a TUI session that never existed.
- **Not a role-laundering vector**: the kind is constructed only by runtime/bundled code. Inbound channel/cron content can never produce it (those origins are runtime-built, never from message text), so resolving it to owner opens no new path for untrusted content to gain authority.

`memory-logger` and `memory-retrieval` now spawn with this origin. Two provenance channels are preserved deliberately:

- **`triggeredBy`** on the origin → execution-time audit (and dropped from the persisted `MinimalSessionOrigin` projection so channel author identity never lands in git-backed `sessions/`).
- **`payload.origin`** → content provenance, unchanged, still drives the existing channel-origin retrieval/extraction safeguards.

## Confidentiality note

Role is authorization, not observability. Guest turns remain eligible for memory logging — the operator's agent is allowed to remember its own conversations. The confidentiality boundary lives at **retrieval/response** (the existing channel-origin index-mode + outbound guards), not at logging. This PR does not change those safeguards.

## Commits

1. `feat(agent): add the system session origin kind and its rendering` — defines the kind, threads it through every exhaustive origin switch (render, slim-prompt, persisted projection, inspect label). Role still falls through to guest here.
2. `feat(permissions): resolve the system origin to owner` — the isolated privilege grant, reviewed on its own.
3. `fix(memory): spawn memory subagents with system authority` — wires the two memory subagents to it.

## Test plan

- New tests: system origin → owner even when `triggeredBy` is a guest channel turn; `resolveHiddenPaths` hides nothing for it; rendered provenance shows the real trigger (not a fake TUI); memory-retrieval spawn carries the system origin with the parent as `triggeredBy`.
- `bun run typecheck`, `bun run lint` (0 errors), `bun run format:check` clean. Touched-domain suites: 969 pass, 0 fail.

> Two pre-existing full-suite failures (`resolveSelfPackageJsonPath`, Xvfb entrypoint-shim) are environment/worktree-path artifacts in files this PR does not touch.

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -712,6 +712,7 @@ export function deriveSystemPromptMode(origin: SessionOrigin | undefined): Syste
       return 'full'
     case 'cron':
     case 'subagent':
+    case 'system':
       return 'slim'
     default: {
       const _exhaustive: never = origin
--- a/src/agent/session-meta.ts
+++ b/src/agent/session-meta.ts
@@ -28,10 +28,12 @@ export type MinimalSessionOrigin =
       thread: string | null
     }
   | { kind: 'subagent'; subagent: string; parentSessionId: string }
+  | { kind: 'system'; component: string }
 
 // Reduce a full SessionOrigin to the minimum projection persisted to disk.
-// Drops participant lists, membership counts, recursive provenance, and
-// author identifiers — none of which `typeclaw usage` reads, and all of
+// Drops participant lists, membership counts, recursive provenance (including
+// the system origin's `triggeredBy`, which can carry channel author identity),
+// and author identifiers — none of which `typeclaw usage` reads, and all of
 // which would otherwise land in git history when sessions/ is auto-backed-up.
 // Kept as a separate function so the boundary between "data the LLM sees in
 // the system prompt" (full origin) and "data persisted for usage reporting"
@@ -58,5 +60,7 @@ function minimalOrigin(origin: SessionOrigin): MinimalSessionOrigin {
       }
     case 'subagent':
       return { kind: 'subagent', subagent: origin.subagent, parentSessionId: origin.parentSessionId }
+    case 'system':
+      return { kind: 'system', component: origin.component }
   }
 }
--- a/src/agent/session-origin.test.ts
+++ b/src/agent/session-origin.test.ts
@@ -29,6 +29,17 @@ describe('renderSessionOrigin', () => {
     expect(out).toContain('Stay narrowly within')
   })
 
+  test('system origin names the component and shows honest trigger provenance (not a fake TUI)', () => {
+    const out = renderSessionOrigin({
+      kind: 'system',
+      component: 'memory-logger',
+      triggeredBy: { kind: 'channel', adapter: 'slack-bot', workspace: 'T0', chat: 'C0', thread: null },
+    })
+    expect(out).toContain('`memory-logger` system process')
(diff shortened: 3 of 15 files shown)


devxoul:
Both findings addressed — thanks, the BLOCKER was a real gap.

**[BLOCKER] system origin dropped in the stream parser** — fixed in 418c86a. `SESSION_ORIGIN_KINDS` in `parseSpawnedByOriginJson` now includes `'system'` (with a keep-in-sync-with-the-union note), so a streamed memory/backup spawn no longer loses its owner-resolving origin and silently falls back to guest. Added a regression test through `createSubagentConsumer` that publishes a `new-session` message with `spawnedByOriginJson` = a system origin and asserts it survives parse and reaches the spawn's `createSessionForSubagent`. (Confirmed it fails without the allowlist change.)

Also added (e18e0a5) `'system'` to `ORIGIN_KINDS` + the two `usage/report.ts` switches so `typeclaw usage` labels system sessions instead of bucketing them as `unknown`.

**[CONCERN] trust model not explicit at the plugin boundary** — documented in 14af253 on `SpawnSubagentOptions.spawnedByOrigin`. Rather than enforce (which can't be done without breaking the in-process plugin trust model), the doc makes the model explicit: passing `{ kind: 'system' }` resolves to owner, and that is **not** an escalation path because a plugin is already a full-trust in-process module with no sandbox — minting a system origin grants it nothing it lacks. The narrower anti-forgery guarantee is unaffected: inbound channel/cron *content* can never construct a system origin (those origins are built by the runtime from the transport, never from message text), so a content-driven turn cannot reach owner. Bundled infra (memory, backup) uses `system`; third-party plugins should pass `event.origin`.

If you'd prefer hard enforcement (e.g. restricting `system` to a host-only spawn path that the public `spawnSubagent` can't reach), I'm happy to do that as a follow-up — but it's a larger surface change and I didn't want to expand this regression-fix PR's scope without your call.

PR #502, opened 2026-05-30
feat(permissions): move fail-safe floor to undefined origin; split session.control from channel.respond

devxoul:
## Background

Today the `guest` role is the security floor by virtue of being empty (`permissions: []`). Every defensive `permissions.has(maybeUndefinedOrigin, ...)` check across the codebase — security bypass tiers, `fs.see.*`, subagent spawn — relies on "resolves to guest" meaning "holds nothing". `resolveRole()` returns `'guest'` for an undefined origin, an unmatched channel author, and cron/subagent origins with an unresolved role.

That coupling blocks a feature we want: scoping crowd-facing agents so unmatched authors fall to `guest` while the operator can still grant `guest` a narrow capability (e.g. `channel.respond`, so strangers get a masked, `public/`-only turn). The moment `guest` is non-empty, the implicit floor collapses — an undefined origin would inherit whatever `guest` carries.

This PR decouples the two so `guest` can become a normal grantable role without weakening the floor. It is the foundation for the follow-ups (dropping the `member:*` default; an agent-callable role-grant tool).

## Summary

**1. The floor moves from the `guest` role to the undefined origin.** `has()` now returns `false` for `origin === undefined` before role resolution. The floor no longer depends on `guest` being empty, so every `has(maybeUndefined, ...)` safety default stays closed regardless of what `guest` is later granted. `resolveRole`/`describe` still report `'guest'` for an undefined origin, so audit and display surfaces are unchanged — only the authorization answer is forced closed.

Why here and not "early-return in `resolveRole`": `resolveRole(undefined)` legitimately returns `'guest'` for display, and the `system` origin legitimately resolves to `owner`. The authorization gate is `has()`, so that is the correct and minimal place to pin the floor.

**2. A new `session.control` permission splits `/stop` from `channel.respond`.** The `/stop` command (both the text-prefix path and the native Discord/Slack slash path) previously gated on `channel.respond`. Once an operator can grant `guest` `channel.respond` (so strangers can drive masked turns), that alias would also hand strangers the ability to abort other speakers' in-flight turns. `session.control` is granted to owner/trusted/member but **not** guest, keeping session lifecycle control member-and-up while respond can be opened to guests independently.

## What this does NOT do

- Does **not** grant `guest` any permission. `guest` stays empty here; this only makes the floor independent of that fact.
- Does **not** drop the `member:*` init default or add a grant tool. Those are follow-up PRs.
- Does **not** change masking, cron/subagent dispatch, or the `rolePromotion` guard.

## Review notes

- **Load-bearing:** the `origin === undefined` short-circuit in `has()` (`src/permissions/permissions.ts`). It is the new floor; removing it silently re-opens every `has(undefined, ...)` default. Stale comments in `builtins.ts` that described "empty-permission guest is the fail-safe floor" are corrected to point at the undefined-origin guard.
- **Invariant to verify:** `session.control` is on owner/trusted/member and absent from guest (pinned by a new test in `builtins.test.ts`), and both `/stop` paths in the router gate on it (pinned by three new tests in `router.test.ts`, including the guest-shaped "respond but no control → /stop refused" case).
- **Note:** `roles.permissions` is `restart-required` per `FIELD_EFFECTS`; this PR only adds built-in defaults, no reloadable-classification change.

### DIFF
--- a/src/channels/router.test.ts
+++ b/src/channels/router.test.ts
@@ -5124,6 +5124,67 @@ describe('ChannelRouter channel.respond gate', () => {
     expect(sessions[0]!.prompts).toHaveLength(1)
   })
 
+  test('respond-capable author WITHOUT session.control cannot /stop via text prefix', async () => {
+    const dir = await tempDir()
+    const logs: string[] = []
+    // guest-shaped: channel.respond granted (can drive turns) but no
+    // session.control, so /stop must be refused.
+    const permissions = buildPermissions({
+      alice: ['channel.respond', 'session.control'],
+      stranger: ['channel.respond'],
+    })
+    const { router, sessions } = makeRouter(dir, { permissions, logs })
+
+    // given: alice (full control) starts a live session
+    await router.route(inbound({ authorId: 'alice', externalMessageId: 'm-alice' }))
+    await router.__testing!.flushDebounce(KEY)
+    expect(sessions).toHaveLength(1)
+
+    // when: stranger (respond but no control) types /stop
+    await router.route(inbound({ authorId: 'stranger', text: '/stop', externalMessageId: 'm-stop' }))
+    await new Promise((r) => setTimeout(r, 10))
+
+    // then: the command is refused and the session is not aborted
+    expect(sessions[0]!.aborted).toBe(0)
+    expect(logs.some((l) => l.includes('session.control') && l.includes('author=stranger'))).toBe(true)
+  })
+
+  test('author WITH session.control can /stop via text prefix', async () => {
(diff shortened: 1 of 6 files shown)


devxoul:
Addressed in f51fbf9. Added a direct `permissions.has(undefined, ...)` test that asserts the fail-safe floor holds **even after `guest` is granted that permission** — i.e. `has(undefined, 'channel.respond') === false` while a matched guest origin returns `true`. This locks the undefined-origin floor invariant independently of router or builtin-role behavior, exactly as suggested.

PR #507, opened 2026-05-30
feat(agent): add grant_role tool for conversational role management

devxoul:
> Stacked on #502 (base: `feat/scoped-permissions`). Review #502 first.

## Background

Onboarding someone to a TypeClaw agent today means hand-editing `typeclaw.json#roles` or running `typeclaw role claim` once per person from the CLI. For the crowd-facing scenarios this work targets — a community Discord, a maintainer bot — that is too clunky. We want the operator to manage access conversationally: "respond to author U_WIFE" → grant them member; "let anyone in this channel message you" → grant `guest` `channel.respond`.

The hard part is doing this safely. A role grant is a confused-deputy target. A guest can post prompt-injection into a group channel where a trusted human is engaged; the trusted turn carries the guest's buffered messages in its context, and the model cannot reliably tell the trusted principal's instructions from the guest's quoted text. If `grant_role` were callable from that turn, the injection could steer it into granting the guest capabilities — using the trusted human's authority. "You can only grant what you hold" bounds the grant to the trusted user's power, but that power is exactly what the attacker wants.

## Summary

The gate confines all grants to **single-principal turns**, where no third-party content enters the turn:

- **origin** must be TUI or a 1:1 DM (`isDmChannelOrigin`); group/open channels are refused outright. A DM is two participants (principal + bot), structurally like the TUI — no buffered third-party messages, so no injection surface.
- **caller** must resolve to `owner` or `trusted`.
- **tier ceiling**: a granter cannot assign or empower a role above its own (owner ≥ trusted ≥ member ≥ guest).
- **permission grants** additionally require "grant only what you hold", and `security.bypass.*` is **never** grantable via the tool — it disables guards and stays a deliberate hand-edit gated by the `rolePromotion` guard.

The tool calls `grantRole` / `grantRolePermission` directly (atomic `typeclaw.json` write + `replaceRoles` hot-reload) — the same out-of-band path the role-claim flow uses, which `rolePromotion` intentionally does not cover (its docstring says so). `grantRolePermission` materializes a built-in role's effective default permission set before appending, so granting one capability to a role with no explicit `permissions[]` does not silently narrow it.

This design follows an adversarial review that rejected an earlier "owner+trusted can grant from any channel" version as exploitable via the chain above.

## What this does NOT do

- **TUI grants.** The server's TUI `createSession` does not thread a `PermissionService` (the known-hardening item in `sandbox.mdx`), so `grant_role` is **DM-only** for now. Threading the server permission service is a separate follow-up.
- **Grant `security.bypass.*`.** Refused by the tool; remains a hand-edit.
- **Drop the `member:*` init default.** That is a separate PR.
- **Make permission grants take effect live.** `roles.permissions` is `restart-required`; the tool result says so.

## Review notes

- **Load-bearing:** the gate in `src/agent/tools/grant-role.ts` — specifically `isSinglePrincipalOrigin` (the confused-deputy defense), the tier ceiling, the `security.bypass.*` refusal, and the "grant only what you hold" check. Gate tests in `grant-role.test.ts` cover each branch, including the group-channel rejection.
- **Invariant to verify:** `grantRolePermission` must not narrow a built-in role. `grant.test.ts` pins that granting one permission to `member`/`guest` preserves the full default set.
- **Two commits:** the primitives (`grantRolePermission`, `isDmChannelOrigin`) land first with their own tests; the tool + wiring second.

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -9,7 +9,7 @@ import { loadMemory } from '@/bundled-plugins/memory/load-memory'
 import type { ChannelRouter } from '@/channels/router'
 import { getConfig, resolveModel, resolveProfile } from '@/config'
 import { defaultThinkingLevelForRef, providerForModelRef, type KnownModelRef } from '@/config/providers'
-import type { PermissionService } from '@/permissions'
+import type { PermissionService, RolesConfig } from '@/permissions'
 import type {
   BuiltinToolRef,
   HookBus,
@@ -50,6 +50,7 @@ import { createChannelFetchAttachmentTool } from './tools/channel-fetch-attachme
 import { createChannelHistoryTool } from './tools/channel-history'
 import { createChannelReplyTool } from './tools/channel-reply'
 import { createChannelSendTool } from './tools/channel-send'
+import { createGrantRoleTool } from './tools/grant-role'
 import { createRestartTool } from './tools/restart'
 import { createSkipResponseTool } from './tools/skip-response'
 import { createSpawnSubagentTool } from './tools/spawn-subagent'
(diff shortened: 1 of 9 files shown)


devxoul:
Good catch — addressed in 6681a81. The hot-reload did reapply stale roles: `grantRole` writes `typeclaw.json` directly, but the reload read `getConfig().roles` (an in-memory snapshot taken before the write), so `replaceRoles` reinstated the pre-grant table.

Fix: renamed the option `rolesProvider` → `reloadRoles` with a contract that it must re-read **from disk**, and wired production to a reload-then-read (`reloadConfig` + `getConfig`) that mirrors the config reloadable's order. Added a regression test that spies on `replaceRoles` and asserts it receives the freshly-written rule, not a stale snapshot.

Note: the role-claim controller has the same `getConfig`-snapshot shape, but it's pre-existing (not introduced here) and the operator is watching the CLI through that flow, so I left it out of scope and will track it separately rather than widen this PR.

PR #510, opened 2026-05-30
feat: auto-reload config after `typeclaw role claim`

devxoul:
## What

After `typeclaw role claim` succeeds, the CLI now triggers a `config`-scoped hot reload of the running agent, so the freshly-granted `roles.<role>.match` rule takes effect immediately — no manual `typeclaw reload` or container restart required.

## Why

`role claim` persists the new match rule to `typeclaw.json` (via `grantRole`), but the running agent kept serving its in-memory config until the next explicit reload/restart. Users had to remember to run `typeclaw reload` for the claim to actually grant anything. This closes that gap so claiming a role "just works".

## How

- New `reloadAfterClaim` helper (`src/role-claim/reload-after-claim.ts`) wraps the existing `requestReload` WS client with `scope: 'config'`.
- `src/cli/role.ts` calls it after the claim completes, reusing the same agent `url` already built for the claim session — no new connection-discovery code.
- **Best-effort by contract:** the role is already on disk when the claim completes, so a reload failure is surfaced to the user (with a hint to run `typeclaw reload`) but does **not** fail the claim.
- Logic lives in the domain layer (`src/role-claim/`) so it is unit-tested directly; the CLI stays a thin shell per the repo's testing philosophy.

The `config` reloadable (`createConfigReloadable`) already runs `reloadConfig()` then `permissions.replaceRoles(getConfig().roles)` when `roles.match` changes, so the new rule is applied to the live `PermissionService` end-to-end.

## Testing

- `src/role-claim/reload-after-claim.test.ts`:
  - reloads with `scope: 'config'` on success
  - treats a reload failure as non-fatal and surfaces the reason
  - stringifies non-Error throwables
- Full suite green locally: `bun run lint`, `oxfmt .`, `tsgo --noEmit`, `bun test --parallel` (1187 pass / 0 fail).

### DIFF
--- a/src/cli/role.ts
+++ b/src/cli/role.ts
@@ -3,7 +3,7 @@ import { defineCommand } from 'citty'
 
 import { requireContainerRunning, resolveHostPort, resolveTuiToken } from '@/container'
 import { findAgentDir } from '@/init'
-import { runClaimSession } from '@/role-claim'
+import { reloadAfterClaim, runClaimSession } from '@/role-claim'
 
 import { c, errorLine } from './ui'
 
@@ -76,6 +76,15 @@ const claimSub = defineCommand({
 
     if (result.kind === 'completed') {
       s.stop(c.green(`Paired as ${result.payload.role}.`))
+      s.start('Reloading config so the new match rule takes effect...')
+      const reloaded = await reloadAfterClaim({ url })
+      if (reloaded.ok) {
+        s.stop(c.green('Config reloaded.'))
+      } else {
+        // The role is already persisted; a reload failure is non-fatal.
+        s.stop(c.yellow(`Config reload failed: ${reloaded.reason}`))
+        console.log(c.dim('Run `typeclaw reload` manually to apply the new match rule.'))
+      }
       outro(`Match rule added: ${c.bold(result.payload.matchRule)}`)
       return
     }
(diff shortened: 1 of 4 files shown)


devxoul:
Thanks @typeey — both findings addressed:

1. **Missing `reloadAfterClaim` import in `src/cli/role.ts`** — already fixed in `38e685d` (the import line is now `import { reloadAfterClaim, runClaimSession } from '@/role-claim'`). Your review landed against the earlier `b0e8170`; the current branch head builds cleanly (`bun run typecheck` passes).

2. **`reloadAfterClaim` reported success even on `ok: false`** — valid catch, fixed in `4a1c9f2`. `requestReload` resolves with `ReloadResult[]` even when the server reports a per-scope failure, so the old `return { ok: true, results }` masked failed reloads. It now inspects the results and surfaces any failed scope (and an empty result set) as `{ ok: false, reason }`, so the CLI's existing "Config reload failed" branch actually fires and tells the user to run `typeclaw reload` manually. Added tests for the failed-scope and empty-result paths.

devxoul:
Correction to my note above: the reload-failure fix landed as `53576cd` (not `4a1c9f2` — typo). Remote branch head is now `53576cd`, both findings addressed, all checks green.

PR #515, opened 2026-05-30
docs: sync permissions, CLI, and channel docs with recent PRs

devxoul:
## Background

A run of feature and fix PRs merged since the last docs sync without touching the docs site, so typeclaw.dev drifted from the runtime. The pages a user lands on — reference/permissions, concepts/permissions-model, reference/cli, reference/channel-adapters, and guides/getting-started — all described behavior that no longer matched the code.

While auditing those pages I also found a correctness bug in reference/permissions.mdx: the security-guard tier tables placed `gitExfil`, `rolePromotion`, and `cronPromotion` under the `high` tier. All three are `medium` per their `GUARD_*_SEVERITY` constants in `src/bundled-plugins/security/policies/*`. A reader copying the tier model to scope role bypass logic would mis-classify which guards operators need elevated permissions to disable.

PRs covered: #502 (session.control split + undefined-origin fail-safe floor), #507 / #513 (grant_role tool), #508 (chat agents scoped-by-default, no `member: ["*"]` seeding), #510 (role claim auto-reload), #491 (GitHub App opened-PR-as-review), #492 (scaffolded `public/` directory).

## Summary

Every claim was cross-checked against source, not inferred from PR titles.

**permissions.mdx** — Added `session.control` and `subagent.spawn.operator` to the role and core-permission tables. Expanded the `subagent.*` wildcard shorthand into real individual perms (only owner's `security.bypass.*` sentinel actually expands in source). Fixed the tier tables: `high` now holds exactly `outboundSecret`, `systemPromptLeak`, and `gitRemoteTainted`; the other three move to `medium`. Added a "Runtime role grants" section documenting the `grant_role` tool's origin/role/tier gating and the grant-only-what-you-hold invariant.

**permissions-model.mdx** — Documented the undefined-origin fail-safe floor (guest is now grantable, so the floor moved off the role enum onto the no-actor input path), the `session.control` vs `channel.respond` split, the `grant_role` flow, and the muted-until-claimed chat-agent default.

**cli.mdx** — Noted that `typeclaw role claim` auto-reloads config on success (best-effort) and cross-linked `grant_role`.

**channel-adapters.mdx** — Documented GitHub App mode treating an opened PR as a review request (an App cannot sit in `requested_reviewers`, so the adapter treats `pull_request.opened` as the trigger) versus PAT mode which waits for `review_requested`. Added team-reviewer support note.

**getting-started.mdx** — Corrected the scaffolded folder list: added `public/`, `.agents/skills/`, `mounts/`, `packages/`; removed `memory/`, which `typeclaw init` does not scaffold.

## Preview

Docs-only Markdown edits — table rows, prose paragraphs, one section addition. No rendered UI changes beyond the page content itself.

## What this does NOT do

- No source or runtime changes of any kind.
- Does not touch `docs/content/docs/internals/permissions.mdx` — that page already lists the high-tier inhabitants correctly. This PR is scoped to the reference/concept/guide pages a user lands on; a follow-up can align the internals role-matrix table.
- Does not re-document `public/` egress in the sandbox page — already covered by #492's `docs(sandbox)` commit.

## Review notes

The load-bearing change is the high-vs-medium tier reclassification in `reference/permissions.mdx`. Verify against the `GUARD_*_SEVERITY` constants: `high` must be exactly `{outboundSecret, systemPromptLeak, gitRemoteTainted}`. The three moved guards (`gitExfil`, `rolePromotion`, `cronPromotion`) belong in `medium`.

Role-matrix rows were rebuilt from `src/permissions/builtins.ts` `BUILTIN_ROLES`. Confirm `member` now lists `session.control` and `trusted` now lists `subagent.spawn.operator`.

Checks: `oxfmt --check` (whole repo clean), `oxlint` (no new warnings; 64 pre-existing unrelated), `tsgo --noEmit` (green).


devxoul:
Thanks @typeey. Addressed in c4099f9:

**Role precedence (fixed).** `permissions-model.mdx` now spells out the real walk order instead of "first matching role wins": `owner` → `trusted` → custom roles (reverse declaration order, later wins) → `member` → `guest`, with the built-in tower pinned ahead of user rules to stop a broad `member.match` from demoting the owner's own session. Matches `buildRoleTable` in `src/permissions/permissions.ts`.

**`grant_role` permission grants — keeping "restart-required".** I re-checked the code and the doc is correct as written. `FIELD_EFFECTS['roles.permissions']` is `restart-required` (`src/config/config.ts:558`, comment at :551: "permissions lists are restart-required (changing what a role can DO)"). The `replaceRoles()` call after a permission grant rebuilds the *match* table (so a same-session subsequent match-grant reads a consistent table), but `grant.ts:78-80` and `grant-role.ts:29-31` both state the permission grant itself does not take effect until restart, and the tool's user-facing message says "run `typeclaw restart`". The split is intentional: `roles.match` is `applied` (hot), `roles.permissions` is `restart-required`. Happy to add a sentence clarifying that the match table reloads but the new permission waits for restart, if you think the distinction is worth calling out in the docs.

PR #520, opened 2026-05-31
feat(slack): !cmd thread fallback for native slash commands

devxoul:
## Background

Slack blocks native slash commands inside threads — the platform returns "/stop is not supported in threads. Sorry!" and never fires a slash_commands event for thread invocations.
`/stop` was therefore unusable from inside a thread.
Worse, the AMBIGUOUS reply copy told users to reply "/stop" inside the specific thread — advice Slack makes impossible to follow.

## Summary

Recognize a leading "!" in ordinary messages (which do fire in threads) and route a recognized command through the same executeCommand path as native slashes.
The guard is strict: only a first token matching a known command name (currently "stop", from SLACK_SLASH_COMMAND_NAMES) is treated as a command — casual messages like "!nice work" pass through as ordinary agent input.
Pattern borrowed from NousResearch/hermes-agent's `!` prefix (gateway/platforms/slack.py).

## Design notes

**Exact thread targeting.**
A thread message carries `thread_ts`, so the "!cmd" path targets the exact thread session.
The key is constructed with `key.thread = thread_ts` rather than null.
The native slash path sets thread to null and relies on a workspace+chat fallback — retiring the ambiguous-match case for thread invocations.
The AMBIGUOUS copy now advises "!stop".

**Interception point.**
Interception lives in `handleMessageEvent` before classifyInbound.
A command is control traffic — neither dropped nor agent-routed — so it must short-circuit before classify.
The classifyInbound contract (drop or route) stays clean.

**Security parity.**
Bypassing classifyInbound also bypasses its self_author and no_user drop guards, so those are explicitly replicated.
Never execute a command from the bot's own message (prevents echo loops) or a userless event.
`invokerId` is forwarded to executeCommand, which gates on session.control, so a stranger in a thread cannot abort someone else's turn.
Dedupe is marked on execution so a Slack redelivery or double-delivery cannot fire the command twice.

**Shared reply mapper.**
`commandResultReply` now backs both the native-slash ack payload and the "!cmd" thread postMessage reply.
The two delivery paths cannot drift; the ternary chain in createSlashCommandHandler is de-duplicated.

**Scope.**
Covers every name in SLACK_SLASH_COMMAND_NAMES (not just "stop"), so the feature scales as that set grows with no separate allow-list to drift.

## What this does NOT do

- Does not change native slash behavior for non-thread invocations.
- Does not add a separate allow-list distinct from SLACK_SLASH_COMMAND_NAMES.
- Does not change classifyInbound's contract or the engage/suppress logic downstream.

## Review notes

Load-bearing changes:

- `parseThreadCommand` (slack-bot-slash-commands.ts) — parses the "!" prefix against the known-command set; unknown "!" prefixes return "ignore" and never execute.
- `createThreadCommandHandler` factory (slack-bot.ts) — execute-then-reply cycle, including guard replication.
- Guard block in `handleMessageEvent` (slack-bot.ts) — placed before classifyInbound; dedupe marked on "executed".
- SLACK_SLASH_REPLY_AMBIGUOUS update — the old copy was actionably wrong.

## Testing

87/87 tests pass across the two Slack test files (11 new parseThreadCommand cases, 7 new createThreadCommandHandler cases).
typecheck clean; lint 0 errors; format:check clean.

Two pre-existing failures in the full suite, both unrelated to src/channels/:

- resolveSelfPackageJsonPath — asserts path ends with "typeclaw/package.json"; fails only when the checkout dir name differs from "typeclaw".
- createCloudflareNamedProvider restarts-with-backoff — timing-sensitive tunnel flake.

### DIFF
--- a/src/channels/adapters/slack-bot-slash-commands.test.ts
+++ b/src/channels/adapters/slack-bot-slash-commands.test.ts
@@ -2,7 +2,12 @@ import { describe, expect, test } from 'bun:test'
 
 import type { SlackSocketModeSlashCommandArgs } from 'agent-messenger/slackbot'
 
-import { buildSlashAckPayload, parseSlashCommand } from './slack-bot-slash-commands'
+import {
+  buildSlashAckPayload,
+  parseSlashCommand,
+  parseThreadCommand,
+  type ThreadCommandInput,
+} from './slack-bot-slash-commands'
 
 type Body = SlackSocketModeSlashCommandArgs['body']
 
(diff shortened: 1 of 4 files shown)


devxoul:
Addressed the dedupe race (request-changes).

**Fix (`6636310`):** The thread-command path marked dedupe only *after* awaiting `executeCommand` + the reply post, so two duplicate Slack deliveries could both clear `dedupe.check()` and both execute the same control command across the check→execute window.

Now `createThreadCommandHandler` takes a `reserve` callback it invokes **synchronously between parse and execute, before the first await**. The adapter implements it as an atomic check-then-mark over the live event. The race loser returns a new `duplicate` outcome and aborts without touching the router — marking is no longer deferred to post-execution.

**Tests added:**
- Two concurrent duplicate deliveries through a real `createSlackDedupe` (gated router) → asserts **exactly one** execution.
- Ordering test → `reserve` fires **before** the router await.
- Lost-race short-circuit → `duplicate` outcome, no router call, no reply.

Also de-duplicated the `ThreadCommandInput` type (was declared in two modules).

Local gate: typecheck clean, lint 0 errors, format clean, 90/90 slack tests + 1120 channels tests green.

PR #524, opened 2026-05-31
Add MCP dispatcher tools and per-server prompt catalog

devxoul:
Stacked on #523 — review/merge in order.

## Background

PR #523 can connect to MCP servers; this PR exposes them to the model without bloating the prompt. This is where the token-economy design pays off.

## Summary

**Three fixed dispatcher tools** replace N×M individual tool registrations:
- `mcp_list_tools(server)` — enumerate a connected server's tools on demand.
- `mcp_describe(server, tool)` — fetch the full input JSON Schema for one tool on demand.
- `mcp_call(server, tool, args)` — invoke a tool.

All three handle unknown server/tool gracefully (helpful error strings, never throw) and surface MCP `isError` results as text rather than exceptions.

**Per-server catalog in the system prompt.** One line per connected server rendered into the STABLE (cacheable) region of the prompt — after origin/role, before git-nudge and memory — so the cache-suffix contract is preserved. When there are no connected servers the block is omitted entirely, keeping byte-identical prompts for agents that don't use MCP.

**Boot wiring in `src/run/index.ts`.** Manager created from `config.mcpServers`, `connectAll()` awaited, threaded into TUI/server/channel/cron/command sessions, `closeAll()` on shutdown.

**Why a dispatcher and not per-tool native registration.** `pi-coding-agent` 0.67.3's `setActiveToolsByName` only filters pre-registered tools; it cannot hot-add definitions mid-session (verified by reading the installed package). Individual MCP tools can never become native tools after the session starts, so the generic dispatcher is not a workaround — it's the only design that fits the runtime.

**Subagents do not get the dispatcher.** Their tool set is intentionally narrowed; MCP access is an operator-level concern.

## What this does NOT do

No code-execution/sandbox invocation path ("code mode") — deferred as a possible future for workflow-heavy servers. No MCP resources or prompts yet (tools only). The cache-suffix guard test (`scripts/dump-system-prompt.test.ts`) passes unchanged because the catalog is empty when no servers are connected (the placeholder dumper has none).

## Review notes

Two load-bearing things:

1. **Catalog placement preserves the cache-suffix contract.** A direct `composeSystemPrompt` test asserts the catalog sits after origin and before memory. Reordering the section breaks that test.
2. **Boot wiring in `src/run/index.ts`.** Verify the manager lifecycle: create → `connectAll` → thread into sessions → `closeAll` on stop. Confirm the manager is NOT threaded into subagent sessions.

Pre-existing flaky tests unrelated to this change: a schema-drift guard (also fails on clean main) and `resolveSelfPackageJsonPath` (worktree-path sensitive).

### DIFF
--- a/src/agent/index.test.ts
+++ b/src/agent/index.test.ts
@@ -274,6 +274,25 @@ describe('createResourceLoader', () => {
     expect(prompt).not.toContain('Session started at')
   })
 
+  test('composeSystemPrompt places MCP catalog after origin and before memory', () => {
+    const catalog = '## MCP servers\n\n- files (2 tools): Filesystem tools'
+    const prompt = composeSystemPrompt({
+      self: '# Identity\n\nfoo',
+      origin: { kind: 'tui', sessionId: 'ses_test' },
+      mcpCatalog: catalog,
+      gitNudge: '',
+      memorySection: '## Memory\n\nmemory-marker',
+    })
+
+    const originIndex = prompt.indexOf('## Session origin')
+    const catalogIndex = prompt.indexOf(catalog)
+    const memoryIndex = prompt.indexOf('memory-marker')
+
+    expect(originIndex).toBeGreaterThan(-1)
+    expect(catalogIndex).toBeGreaterThan(originIndex)
+    expect(memoryIndex).toBeGreaterThan(catalogIndex)
+  })
+
   test('memory section is NOT visible to plugin session.prompt hooks (intentional: memory injection is core-owned and runs after all plugin hooks)', async () => {
     // The security plugin's applyPromptInjectionDefense scans `event.prompt`
     // for attack patterns during the session.prompt hook chain. After this PR
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -2,13 +2,21 @@ import { existsSync } from 'node:fs'
 import { dirname, join } from 'node:path'
 import { fileURLToPath } from 'node:url'
 
-import { createAgentSession, DefaultResourceLoader, SessionManager } from '@mariozechner/pi-coding-agent'
+import {
+  createAgentSession,
+  DefaultResourceLoader,
+  defineTool as definePiTool,
+  SessionManager,
+} from '@mariozechner/pi-coding-agent'
 import type { AgentSession, ToolDefinition } from '@mariozechner/pi-coding-agent'
 
 import { loadMemory } from '@/bundled-plugins/memory/load-memory'
 import type { ChannelRouter } from '@/channels/router'
 import { getConfig, resolveModel, resolveProfile } from '@/config'
 import { defaultThinkingLevelForRef, providerForModelRef, type KnownModelRef } from '@/config/providers'
+import { renderMcpCatalog } from '@/mcp/catalog'
+import type { McpManager } from '@/mcp/manager'
+import { createMcpDispatcherTools, MCP_DISPATCHER_TOOL_NAMES } from '@/mcp/tools'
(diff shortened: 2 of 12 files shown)


devxoul:
@typeey review again

devxoul:
Good catch — fixed. Replaced the case-sensitive `startsWith` check with an `isHttpProtocol(u)` helper that checks `new URL(u).protocol` (the WHATWG parser normalizes the scheme to lowercase), so `HTTPS://…` and `HTTP://…` now validate correctly. Added a regression test for uppercase schemes. The fix is on the config branch and cascaded through the stack.

devxoul:
Good catch — fixed. Replaced the case-sensitive `startsWith` check with an `isHttpProtocol(u)` helper that checks `new URL(u).protocol` (the WHATWG parser normalizes the scheme to lowercase), so `HTTPS://…` and `HTTP://…` now validate correctly. Added a regression test for uppercase schemes. The fix is on the config branch and cascaded through the stack.

devxoul:
Good catch — fixed. Replaced the case-sensitive `startsWith` check with an `isHttpProtocol(u)` helper that checks `new URL(u).protocol` (the WHATWG parser normalizes the scheme to lowercase), so `HTTPS://…` and `HTTP://…` now validate correctly. Added a regression test for uppercase schemes. The fix is on the config branch and cascaded through the stack.

devxoul:
Good catch — fixed. Replaced the case-sensitive `startsWith` check with an `isHttpProtocol(u)` helper that checks `new URL(u).protocol` (the WHATWG parser normalizes the scheme to lowercase), so `HTTPS://…` and `HTTP://…` now validate correctly. Added a regression test for uppercase schemes. The fix is on the config branch and cascaded through the stack.

devxoul:
Good catch — fixed. Replaced the case-sensitive `startsWith` check with an `isHttpProtocol(u)` helper that checks `new URL(u).protocol` (the WHATWG parser normalizes the scheme to lowercase), so `HTTPS://…` and `HTTP://…` now validate correctly. Added a regression test for uppercase schemes. The fix is on the config branch and cascaded through the stack.

devxoul:
Good catch — fixed. Replaced the case-sensitive `startsWith` check with an `isHttpProtocol(u)` helper that checks `new URL(u).protocol` (the WHATWG parser normalizes the scheme to lowercase), so `HTTPS://…` and `HTTP://…` now validate correctly. Added a regression test for uppercase schemes. The fix is on the config branch and cascaded through the stack.

devxoul:
Good catch — fixed. Replaced the case-sensitive `startsWith` check with an `isHttpProtocol(u)` helper that checks `new URL(u).protocol` (the WHATWG parser normalizes the scheme to lowercase), so `HTTPS://…` and `HTTP://…` now validate correctly. Added a regression test for uppercase schemes. The fix is on the config branch and cascaded through the stack.

devxoul:
Good catch — fixed. Replaced the case-sensitive `startsWith` check with an `isHttpProtocol(u)` helper that checks `new URL(u).protocol` (the WHATWG parser normalizes the scheme to lowercase), so `HTTPS://…` and `HTTP://…` now validate correctly. Added a regression test for uppercase schemes. The fix is on the config branch and cascaded through the stack.

devxoul:
Good catch — fixed. Replaced the case-sensitive `startsWith` check with an `isHttpProtocol(u)` helper that checks `new URL(u).protocol` (the WHATWG parser normalizes the scheme to lowercase), so `HTTPS://…` and `HTTP://…` now validate correctly. Added a regression test for uppercase schemes. The fix is on the config branch and cascaded through the stack.

PR #526, opened 2026-05-31
fix(github): seed GH_TOKEN for single-owner multi-repo Apps

devxoul:
## Background

A running agent with GitHub **App** credentials in `secrets.json` failed to authenticate `gh` for ad-hoc calls — `gh auth status` reported "no hosts logged in" — and resorted to ~90s of trial-and-error (inspecting `secrets.json`, fumbling `node` heredocs, tripping over the PKCS#1 private key) before manually minting an installation token to post PR reviews.

Root cause: the adapter seeds the process-wide `GH_TOKEN` only when **exactly one repo** is configured (or PAT auth), keying off *repo count* via `ghTokenSeedRepo`. But a GitHub App installs **per owner**, so the correct token is unambiguous whenever every configured repo shares a single owner — no matter how many repos are listed. The old gate conflated "multiple repos" with "multiple installations," leaving `GH_TOKEN` empty for the common single-owner/multi-repo setup (e.g. several `typeclaw/*` repos under one installation). The agent then had a perfectly resolvable token it was never handed.

## Summary

Key the seed off **owner count, not repo count**:

- **PAT** → seed context-free (unchanged).
- **App, all repos under one owner** → seed using an owner-scoped context (`{ owner }`), which `AppAuthStrategy` already resolves via `GET /orgs/{owner}/installation`.
- **App, no repos configured** → seed context-free; `AppAuthStrategy` discovers the sole installation.
- **App, repos spanning >1 owner** → genuinely ambiguous; keep the skip + log. The adapter's own API calls still resolve a repo-scoped token per call via `authToken`, so outbound posting is unaffected.

Why owner-scoped and not repo-scoped for the seed: a single `GH_TOKEN` must represent one installation; the owner is the installation boundary, so `{ owner }` is the smallest unambiguous key that works for any number of repos under that owner.

This complements #521 (multi-installation support on the read/outbound path) by fixing the *seed* path it left conservative.

## What this does NOT do

- Does not change PAT behavior or the outbound/read token resolution.
- Does not attempt to seed a global token for genuinely multi-owner Apps — that case remains a deliberate skip with operator guidance.
- Does not add a `gh` wrapper, credential helper, or any new CLI surface.

## Review notes

- Load-bearing change: `ghTokenSeedContext(authType, repos)` (replaces `ghTokenSeedRepo`) — returns `undefined` (context-free seed), `{ owner }` (owner-scoped seed), or `null` (skip). The `null`-vs-`undefined` distinction is the crux; the closure narrows `null` out before calling `auth.token`.
- Invariant to verify: single-owner/multi-repo now seeds (`GET /orgs/{owner}/installation`); multi-owner still skips and logs. Both are covered by new lifecycle tests, plus a sole-installation context-free case.

### DIFF
--- a/src/channels/adapters/github/index.ts
+++ b/src/channels/adapters/github/index.ts
@@ -176,17 +176,18 @@ export function createGithubAdapter(options: GithubAdapterOptions): GithubAdapte
         throw err
       }
       started = true
-      // GH_TOKEN is a single process-wide env var the container's `gh` CLI
-      // reads, but a GitHub App spanning multiple owners has no single correct
-      // token. Seed/refresh it only when exactly one repo is configured (PAT,
-      // or App with one unambiguous installation). With multiple repos we skip
-      // the global seed: ad-hoc `gh` calls must target a specific repo, and the
-      // adapter's own API calls always resolve a repo-scoped token via authToken.
-      const ghTokenRepo = ghTokenSeedRepo(options.configRef().repos ?? [])
-      const seedGhToken = async (): Promise<void> => {
-        process.env.GH_TOKEN = await auth.token(ghTokenRepo === null ? undefined : { repoSlug: ghTokenRepo })
-      }
-      if (ghTokenRepo !== null || options.secrets.auth.type === 'pat') {
+      // GitHub Apps install per owner, so the single process-wide GH_TOKEN is
+      // unambiguous whenever every configured repo shares one owner (one
+      // installation) — regardless of repo count. Only repos spanning >1 owner
+      // are ambiguous; that case skips the global seed (authToken still resolves
+      // a repo-scoped token per call). Keying off owner count, not repo count,
+      // is the fix for single-owner/multi-repo Apps that the old gate skipped.
+      const ghTokenContext = ghTokenSeedContext(options.secrets.auth.type, options.configRef().repos ?? [])
+      if (ghTokenContext !== null) {
+        const seedContext = ghTokenContext
+        const seedGhToken = async (): Promise<void> => {
+          process.env.GH_TOKEN = await auth.token(seedContext)
+        }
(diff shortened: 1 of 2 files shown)


devxoul:
Both addressed in 66446a8.

**[BLOCKER] user-owned repos** — Fixed. Seeding no longer uses an owner-scoped context. `ghTokenSeedContext` is replaced by `ghTokenSeedDecision`, which seeds from a configured **repo slug** via `repos/{owner}/{repo}/installation` — confirmed against the GitHub REST docs as the only installation-lookup that works for **both** org- and user-owned repos (`orgs/{owner}/installation` 404s on personal accounts). When all configured repos share one owner, we pick a deterministic (sorted) slug from that owner; the resulting installation token already covers every repo under that installation. Added a user-owned/personal-account regression test that 404s `orgs/{owner}/installation` and asserts the seed path never touches the `orgs/` endpoint.

**no-repos case** — Fixed to your preference. The empty-repos branch now **skips** the global seed (reason `no-repos`) rather than seeding context-free via sole-installation discovery, restoring the original "don't seed unless we can prove the installation" contract. Added a dedicated test asserting no `access_tokens` POST happens and the skip is logged.

Seeding decision table now:

| Auth | Repos | Owners | Result |
|---|---|---|---|
| PAT | any | — | seed (context-free) |
| App | 0 | 0 | **skip** (no-repos) |
| App | 1+ | 1 | seed via `{ repoSlug }` (org or user) |
| App | 2+ | 2+ | **skip** (multiple-owners) |

The multi-owner skip stays: a single static `GH_TOKEN` can't span two installations, and a `gh` credential helper would be shadowed by `GH_TOKEN` (and `gh api` ignores helpers entirely). Filed #527 to track a proper per-owner credential mechanism for the multi-owner case.

PR #529, opened 2026-06-01
feat(channels): add /help command listing available commands without an LLM call

devxoul:
## Background

Channel users had no way to discover what commands an agent accepts. The only
command was `/stop`, and you had to already know it existed. There was also no
in-band, zero-cost way to ask — typing anything that wasn't `/stop` fell through
to the agent and burned an LLM turn just to answer "what can I type here?".

This adds a `/help` command that lists the available commands and replies
directly, without ever reaching the agent (no LLM call).

## Summary

`/help` is registered as a normal channel command alongside `/stop`, so it works
on every surface that already routes commands: text-prefix `/help`, Slack native
slash command, Slack `!help` thread fallback, and Discord slash interaction.

Two design decisions a reviewer would otherwise have to ask about:

- **Generated, not hardcoded help text.** `/help` enumerates the registry via
  new `list()`/`get()` metadata accessors and formats from each command's
  `description`. Adding a future command makes it appear in `/help`
  automatically — the listing can't drift from the actual command set.

- **Session-less + ungated, modeled explicitly rather than special-cased.**
  `/stop` requires a live turn and `session.control` permission. `/help` needs
  neither. Instead of branching on the command *name* in the dispatcher, commands
  now declare `permission` and `requiresLiveSession`, and both dispatch paths
  (`route()` text-prefix and `executeCommand`) honor those declarations. `/help`
  is ungated because it only reveals command names that are already public via
  the platform slash-command pickers; gating it behind `session.control` would
  make it useless in the exact shared channels where discovery matters most.

The registry generalization (metadata + a `{ reply }` handler payload) lands as a
separate first commit with no behavior change, so the `/help` commit is purely
the feature.

## What this does NOT do

- Does not add a TUI-local `/help`. The TUI only intercepts `/quit`/`/exit`
  client-side; `/help` typed in the TUI would reach the server like any prompt.
  Out of scope here.
- Does not introduce per-command help visibility (e.g. hiding admin-only
  commands from the listing). With only `stop` + `help` today there's nothing to
  hide; that can be modeled later if a sensitive command is added.
- Does not rename the Discord/Slack `STOP_REPLY_*` constants now that the reply
  path is shared — they remain accurate for `/stop`, and `/help` uses the
  handler-provided `reply`.

## Review notes

- Load-bearing change: the two dispatch sites in `src/channels/router.ts`
  (`route()` text-prefix block and `executeCommand`) now branch on
  `commandInfo.permission` / `commandInfo.requiresLiveSession`. Verify `/stop`
  still gates on `session.control` and still returns `no-live-session` on a cold
  channel — covered by the existing tests, which were updated only to reflect
  that `/stop` now also carries a confirmation `reply`.
- The `/help` handler forward-references `commands` (it reads the registry it
  belongs to). This is safe because the handler only runs on invocation, long
  after construction; the array is declared first, then handed to
  `createCommandRegistry`.
- Slack manifest sync: `SLACK_APP_MANIFEST.features.slash_commands` and the
  runtime `SLACK_SLASH_COMMAND_NAMES` allow-list are now asserted equal by a new
  drift test in `ui.test.ts`, making good on the comment that already claimed
  this invariant.

### DIFF
--- a/src/channels/adapters/discord-bot.test.ts
+++ b/src/channels/adapters/discord-bot.test.ts
@@ -842,7 +842,7 @@ describe('createInteractionHandler', () => {
   type CapturedCall = { url: string; init: RequestInit }
   type RouterCall = { key: ChannelKey; name: string; invokerId: string }
   type RouterResult =
-    | { kind: 'handled'; name: string }
+    | { kind: 'handled'; name: string; reply?: string }
     | { kind: 'no-live-session' }
     | { kind: 'permission-denied' }
     | { kind: 'unknown-command'; name: string }
@@ -912,6 +912,18 @@ describe('createInteractionHandler', () => {
     expect(fetchCalls[0]!.url).toBe('https://discord.com/api/v10/interactions/i-1/tok-abc/callback')
     const body = JSON.parse(fetchCalls[0]!.init.body as string)
     expect(body.data.content).toContain('Stopped')
+    expect(body.data.flags).toBe(64)
+  })
+
+  test('/help interaction acks with the handler-provided command list', async () => {
+    const helpText = 'Available commands:\n/help — List available commands\n/stop — Abort the current turn'
+    const { handler, fetchCalls } = setup(async () => ({ kind: 'handled', name: 'help', reply: helpText }))
+
+    await handler(interaction({ data: { name: 'help', type: DISCORD_SLASH_COMMAND_TYPE_CHAT_INPUT } }))
+
+    expect(fetchCalls).toHaveLength(1)
+    const body = JSON.parse(fetchCalls[0]!.init.body as string)
+    expect(body.data.content).toBe(helpText)
     expect(body.data.flags).toBe(64)
   })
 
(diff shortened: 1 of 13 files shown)


devxoul:
Good catch — fixed in e6cb582.

I went with the narrow-exception option (not moving all command parsing above the gate, which would broaden the channel.respond boundary and risk `/stop`). A known command that is **both** `permission: 'none'` **and** `requiresLiveSession: false` now runs in a pre-gate fast path, mirroring the native-slash behavior. Requiring both flags prevents a future "public but live-session-aware" command from silently bypassing the gate.

Everything else is unchanged: unknown commands, `/stop`, `//escaped` text, and plain messages still hit the channel.respond gate exactly as before, and the bypassed `/help` is not logged as a denied inbound. Extracted `runChannelCommand` so the pre-gate and post-gate paths share one execute→reply shape.

Added regression tests for all five cases: non-respond user gets `/help`, non-respond `/stop` still denied, non-respond `/foo` still denied (not treated as unknown-command), `//help` not executed, and an authorized `/help` produces exactly one reply.

PR #530, opened 2026-06-01
feat(reviewer): sharpen code-review skill against false positives and vague findings

devxoul:
## Background

The `code-review` skill (loaded by the read-only reviewer subagent) teaches the model what to flag and what to ignore on a code change. Two recurring failure modes weren't covered:

1. **False positives the author already owns.** The reviewer would restate the PR description back at the author, or re-raise a gap the author had already marked with a `TODO`/`FIXME` or scoped out in the PR body. The parent agent then has to filter that noise before posting.
2. **Findings too vague to triage.** A finding anchored to a line says *where* a defect is but not *how far it reaches*, and findings that lean on code outside the diff often cited nothing the author could click through to.

There were also two code-smell classes the language-neutral base prompt can't catch: temporary scaffolding (`wip`/`fixup!` commits, debug logging) that escaped into the change, and test suites that look broad but only vary the input while asserting the same thing.

## Summary

Four additions to `code-review.ts`, all layered on top of the reviewer's existing neutral output contract (no change to severity/verdict/`<review>`):

- **Change hygiene** (new "What to look for" item): flag stray `wip`/`fixup!`/`squash!` commits, debug logging, and leftover scaffolding — and name the commit a stray should fold into, rather than "this looks temporary".
- **Duplicate-test detection** (folded into Test coverage): cases that vary only the input while asserting the same thing exercise one code path, not many — read the gap as untested, not covered.
- **Author-owned false positives** (two new "What NOT to find" items): don't restate the change description; don't re-raise a self-flagged `TODO`/`FIXME` or scoped-out gap unless you have *new* information (and say what).
- **Finding precision** (two new line-anchor rules): state the blast radius (which callers break, which inputs trigger it) so a `concern` vs `blocker` call is groundable, and pin cross-file evidence to a `path:line` / commit permalink so it survives later edits.

Kept deliberately language-agnostic — no syntax-specific lint rules, matching the skill's existing cross-language framing.

## What this does NOT do

- No change to the reviewer subagent prompt, tool set, or output contract.
- No new skill; this only extends the existing `code-review` skill body.
- No language-specific or linter-territory rules (those stay out by design).

## Review notes

The load-bearing change is the skill prose in `src/bundled-plugins/reviewer/skills/code-review.ts`. Three new drift-guard tests in `reviewer.test.ts` assert the new phrasing survives future edits, matching the existing drift-guard convention in that file.


devxoul:
Addressed in d2ecf40. Re-anchored the duplicate-test heuristic on the *inputs being indistinguishable to the code under test* (same branch / validation rule / boundary) rather than identical assertion shape, required the finding to name the path the cases collapse onto, and added an explicit carve-out: table-driven and parametrized tests that share one assertion while each input hits a distinct branch are coverage, not duplication, and must not be flagged.

PR #531, opened 2026-06-01
fix(github): inject per-repo gh token for multi-owner App auth

devxoul:
## Background

Follow-up to #526. That PR seeded the process-wide `GH_TOKEN` for single-owner/multi-repo GitHub Apps but **deliberately skipped the multi-owner case** (`ghTokenSeedDecision` → `skip: 'multiple-owners'`): `gh` reads a single `GH_TOKEN`, while a GitHub App installs **per owner**, so when configured repos span more than one owner there is no single correct token to seed. A running agent with multi-owner App credentials then can't run ad-hoc `gh` (e.g. post a PR review) against an arbitrary owner — the exact failure #526 set out to fix, narrowed to the multi-owner slice it left conservative.

## Approach

Govern `gh` at the **bash tool boundary** with a new bundled plugin (`github-cli-auth`) instead of trying to seed one global token. Its `tool.before` hook inspects each `gh` invocation and branches on the current `GH_TOKEN` prefix (officially documented, reliable):

| Token | Prefix | Behavior |
|---|---|---|
| Classic PAT | `ghp_` | Cross-owner — pass through unchanged |
| Fine-grained PAT | `github_pat_` | Single-owner, **not** re-mintable per repo, owner not queryable at runtime → enforce an explicit repo so `gh` fails honestly cross-owner; leave the seeded token in place |
| App installation | `ghs_` / unknown | Resolve `owner/repo` from `-R/--repo` or a `gh api /repos/{owner}/{repo}/...` path, **mint that repo's installation token**, and rewrite the command to `GH_TOKEN=<token> gh ...` |

A repo-targeting App-auth call with no resolvable repo is **blocked** with actionable guidance. Repo-less subcommands (`gh auth status`, `gh --version`, `gh api graphql`, `gh api /user`, …) pass through.

## How the token is minted

Reuses the github adapter's existing `AppAuthStrategy` (the same `repos/{owner}/{repo}/installation` lookup #526's reviewer locked in — the only one that works for both org- and user-owned repos) via a small **in-process `GithubTokenBridge`**:

- the github adapter registers a per-repo resolver on **App-auth start** (and unregisters on stop/rollback);
- plugins receive only the **read-side** through `PluginContext.github`;
- when the adapter isn't live the hook gets a typed `unavailable` result and **blocks cleanly** rather than crashing.

This keeps the plugin↔channel subsystems isolated (no router widening, no global singleton, no duplicated auth) — the bridge is the single thin seam.

## Why command-rewrite instead of env injection

A `tool.before` hook **cannot** set the spawned process's env (the bash sandbox does `--clearenv` then sets only `PATH`/`HOME`/`LANG`). So injection is an inline shell assignment baked into the command string. The token is **shell-quoted**, and `github-cli-auth` is ordered **after `security`** so security's `tool.before` scans the original command before the token is ever added.

## What this does NOT do

- No change to classic-PAT behavior or to the existing `GH_TOKEN` seed path from #526.
- No new CLI subcommand; the parser/classifier are pure, unit-tested helpers.
- Does not attempt to make `git push/clone` multi-owner — this is scoped to `gh`, which was the actual #526 symptom.

## Commits

1. `feat(channels): add GithubTokenBridge service`
2. `feat(plugin): expose GithubTokenBridge read-side via ctx.github`
3. `feat(github): register a per-repo token resolver on App-auth start`
4. `feat(github-cli-auth): add pure gh-command parser and token classifier`
5. `feat(github-cli-auth): wire the hook and register the bundled plugin`

## Tests

- `github-token-bridge.test.ts` — unavailable-when-unregistered, throw→unavailable, register/unregister lifecycle, stale-unregister safety.
- `gh-command.test.ts` — repo-less pass-through, `-R`/`--repo`/`--repo=` extraction, `gh api /repos/...` path extraction, missing-repo block, non-`gh` fast-path, leading env-assignment.
- `inject.test.ts` — token classification by prefix, shell-quoting.
- `index.test.ts` — App mint→rewrite, App no-repo→block, adapter-down→block, classic-PAT pass-through, fine-grained-PAT no-mint, non-bash ignored.

`bun run typecheck`, `bun run lint` (0 errors), `bun run format:check`, and the full suite pass (one unrelated pre-existing failure in `update/index.test.ts` is a worktree-directory-name artifact, not a regression).


devxoul:
Thanks @typeey — both findings addressed in 253bd58, f497dbe, 15c2314.

**[HIGH] token in the command string → fixed.** The hook no longer rewrites the command. It now sets a TypeClaw-internal env overlay (`TYPECLAW_INTERNAL_BASH_ENV`) that the bash wrapper delivers to the spawned process *as env*, never in the command string:
- non-sandboxed (the primary owner/TUI path): the bash builtin is built via pi's `createBashTool` with a `spawnHook` that merges a per-call overlay into the spawn env. The overlay is carried through `AsyncLocalStorage`, so concurrent bash calls don't clobber each other (no `process.env` mutation).
- sandboxed: `applyBashSandbox` forwards the overlay to `buildSandboxedCommand` as `env.set`, so bwrap re-adds it via `--setenv` after `--clearenv`.
- the wrapper strips any client-supplied overlay *before* `tool.before` runs (only trusted hooks may set it) and extracts+deletes it before the loop guard and pi's execute, so the token never reaches the command string, logs, or loop-detection state. Tests assert the command is unchanged and the overlay key is absent after execution.

**Compound commands → fixed.** `analyzeGhCommand` now inspects *every* `gh` invocation across `&&`, `||`, `;`, and `|` (the tokenizer splits operators so `true; gh ...` boundaries are detected). Single-owner → inject that owner's token; a repo-targeting `gh` with no resolvable repo → block; invocations spanning >1 owner → block (one `GH_TOKEN` can't authenticate them all). Added coverage for `echo ok && gh ...`, `true; gh ...`, `cmd | gh ...`, multi-`gh` same-owner, multi-owner, and missing-repo-in-chain.

Since injection is now env-based, there is no `GH_TOKEN=... gh` rewrite left to mis-scope.

devxoul:
Thanks @typeey — both round-2 findings fixed in bdbe47d.

**[HIGH] repo-scoped commands in the repo-less allowlist.** Correct — `label`, `ruleset` (and `secret`/`variable`/`cache`) are repo-scoped and were wrongly passing through. The allowlist is now the conservative set of *genuinely* account/global subcommands (`auth, config, extension, alias, completion, gpg-key, ssh-key, status, org, gist, codespace, search, preview, accessibility, attestation`), verified against gh source (commands wired through `cmdutil.EnableRepoOverride` are repo-scoped). Everything else — including `label`/`ruleset`/`secret`/`variable`/`cache` — now falls through to block-unless-explicit-`-R`. Added tests: `gh label list` / `gh ruleset list` / `gh secret set` / `gh variable list` / `gh cache list` block; the same with `-R owner/repo` inject; the repo-less set passes through.

**[MEDIUM] `gh api` arg scanning.** Fixed — `extractRepoFromApiPath` now parses only the endpoint arg (the first positional after `api`, skipping value-taking flags like `-f/-F/-H/-X/-q/--jq/...`), so a `-f query=/repos/evil/repo` field or a `--jq` expression can no longer masquerade as the target. Added tests for the field-value and `--jq` cases plus the legitimate `gh api /repos/acme/widgets/pulls -f body=/repos/evil/repo` extracting only `acme/widgets`.

PR #532, opened 2026-06-01
Add enabled flag to MCP server config

devxoul:
Stacked on #524 — review/merge in order. (Supersedes #525, which GitHub auto-closed after a stack rebase; the enabled-flag change is unchanged.)

## Background

During review of the MCP stack we wanted a way to keep a server's config in `typeclaw.json` while turning it off — e.g. a server that hangs on connect, or one you're temporarily not using — without deleting and re-adding its block.

## Summary

Adds an `enabled` boolean to each MCP server entry, defaulting to `true`.

- Positive polarity, default-true — omitting the field keeps the server on, matching typeclaw's existing convention (cron job `enabled`, docker feature toggles). A `disabled` flag would be the only negative-polarity toggle in the schema.
- A disabled server is completely invisible: the manager filters disabled servers out at construction (`servers.filter((s) => s.enabled)`), so they are never connected, never spawned, and never appear in the system-prompt catalog or the dispatcher's server lookup — one filter point covers both connection and discovery.
- No `FIELD_EFFECTS` change: `enabled` is a subfield of the already-`restart-required` `mcpServers` path.

## What this does NOT do

No hot-reload of the flag (restart-required like the rest of mcpServers). No per-tool enable/disable — server granularity only.

## Review notes

The load-bearing change is the single manager-level filter — verify the test asserting a disabled server is never connected (fake connect not invoked) and is absent from listServers()/getConnection(). This also doubles as the mitigation for the connect-hang concern raised in review: disable a misbehaving server instead of removing it.

### DIFF
--- a/src/config/config.test.ts
+++ b/src/config/config.test.ts
@@ -188,6 +188,7 @@ describe('mcpServerSchema', () => {
     })
     expect(parsed).toEqual({
       name: 'filesystem',
+      enabled: true,
       command: 'bunx',
       args: ['@modelcontextprotocol/server-filesystem'],
       env: {},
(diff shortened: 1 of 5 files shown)


devxoul:
@typeey review again

PR #536, opened 2026-06-01
feat(sandbox): ship jq in the container baseline so it is sandbox-visible

devxoul:
## What

Adds `jq` to `BASELINE_APT_PACKAGES` in `src/init/dockerfile.ts` so it lands in the container base image (and, by extension, the per-agent image).

## Why

The per-tool bwrap sandbox that wraps agent bash calls only `--ro-bind`s `/usr` + `/bin` from the container (see `src/sandbox/build.ts`). A binary that is **not in the base image is invisible inside the sandbox** — there is no per-call install path. `jq` was not installed in the container, so any sandboxed `... | jq` pipeline failed with "command not found".

The `explorer`, `reviewer`, and `scout` subagent prompts already advertise `jq` as an available read-only pipeline tool (`curl ... | jq`, `cat foo.json | jq`), so it ships unconditionally in baseline rather than behind a `docker.file` toggle — consistent with the existing rationale for `bubblewrap`/`util-linux`/`iptables`.

## Notes

- No checked-in base Dockerfile to sync: CI regenerates it from this source via `scripts/emit-base-dockerfile.ts`. Both `buildBaseDockerfile()` and the dev inline path compose from the same `BASELINE_APT_PACKAGES` constant, so they can't drift.
- Existing entrypoint-shim tests asserting the Stop-hook script does **not** contain `jq ` are unaffected (the shim never carried the apt list).

## Testing

- `bun run typecheck` — clean
- `bun run lint` — 0 errors (64 pre-existing warnings, unrelated)
- `bun run format` — no changes
- `bun test src/init/dockerfile.test.ts --parallel` — 207 pass / 0 fail (both baseline-list drift-guard assertions updated)

### DIFF
--- a/src/init/dockerfile.test.ts
+++ b/src/init/dockerfile.test.ts
@@ -112,7 +112,7 @@ describe('buildDockerfile feature toggles', () => {
         }),
       ),
     )
-    expect(pkgs).toEqual(['git', 'ca-certificates', 'curl', 'gnupg', 'iptables', 'util-linux', 'bubblewrap'])
+    expect(pkgs).toEqual(['git', 'ca-certificates', 'curl', 'gnupg', 'iptables', 'util-linux', 'bubblewrap', 'jq'])
   })
 
   test('xvfb: true (the default) adds xvfb to the toggle apt package list, after the baseline packages', () => {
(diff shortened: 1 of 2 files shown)


devxoul:
@typeey can you review again

PR #537, opened 2026-06-01
feat(permissions): allow grant_role in trusted-only group channels

devxoul:
## Background

The `grant_role` guard confined role/permission grants to single-principal turns (TUI / 1:1 DM) to close the confused-deputy surface: a group turn buffers in other authors' messages, so a guest could prompt-inject a trusted turn into rewriting the access-control table.

That refused legitimate cases: an operator alone with the bot in a group-capable channel, and small trusted-only operational channels. A channel with no untrusted humans is injection-equivalent to a DM — no untrusted human can buffer a message into the turn — so it is safe to allow.

## Summary

Relax the `grant_role` origin guard so it accepts a group/open channel in two new cases, in addition to the existing TUI and 1:1 DM:

1. **Single-human group** — the platform proves the channel has exactly one human member.
2. **All-humans-trusted group** — the platform proves every human member resolves to trusted or owner.

Two layers of implementation:

**Carry member identities (commits 2–3).** The membership pipeline was count-only (no identities), which cannot prove *who* is in the room. Added an optional field `humanMemberIds?: readonly string[]`, populated ONLY when an adapter enumerated and classified the complete membership in the same pass that produced the human count, guaranteeing the lengths match by construction.
- **Slack / Discord** collect non-bot member ids on their existing ≤-cap enumeration path. Discord drops to counts-only if any member is unidentifiable (never silently dropped).
- **GitHub** emits collaborator ids serialised as strings to match the inbound authorId namespace; counts-only when the page is truncated or any collaborator is unidentifiable.
- **Telegram / KakaoTalk** cannot enumerate members, so they never set the field.

**The guard (commit 4).** The new helper `isAllHumansTrustedGroupChannelOrigin` allows a grant only when ALL hold (fail-closed otherwise): channel not a DM, membership present + non-truncated + fresh, the member-id list present, length equal to the human count, non-empty, at most 20 members (to bound per-member role-resolution work on a tool call), and every id resolving to trusted or owner. The single-human branch remains for adapters that cannot enumerate identities. The per-turn caller-role check, tier ceiling, and grant-only-what-you-hold are all unchanged.

## Preview

N/A — no visual output.

## What this does NOT do

- Does not use `participants` (observed speakers) for authorization — it cannot prove the absence of a silent untrusted lurker. Only an authoritative complete membership enumeration counts.
- Does not relax the per-turn caller check: the acting human must still resolve to owner/trusted.
- Does not enable the multi-human relaxation for Telegram/KakaoTalk (no identity enumeration) — they stay limited to single-human.
- Stale / truncated / missing membership, count/length mismatch, and rooms over the 20-human cap all refuse.

## Review notes

Both relaxations were designed against Oracle review. Key safety points: a bot misclassified as human resolves to guest → fails the check (fail-closed); a human misclassified as bot is defended by the length-equals-count invariant plus same-pass derivation; Discord guild-member overcount only causes false refusals (fail-closed). No user-facing opt-in — the guard is strictly stricter than "single caller is trusted" and fully fail-closed.

Load-bearing invariant: the member-id list length cannot drift from the human count because the field is populated only in the same pass that increments the count. Any adapter that cannot guarantee this omits the field and falls through to the count-only single-human path.

**Tests:** grant-role: single-human allowed; all-trusted multi-human allowed; one-guest-human refused; length-mismatch refused; missing-ids refused; stale refused; truncated refused; over-cap refused; untrusted caller refused. Adapter resolvers: Slack/Discord enumerate paths now assert `humanMemberIds`.

**Checks:** typecheck clean, lint 0 errors, format:check clean, full test suite green (one unrelated environmental failure in `src/update/index.test.ts` on worktree path; passes on a normal checkout/CI).

### DIFF
--- a/src/agent/tools/grant-role.test.ts
+++ b/src/agent/tools/grant-role.test.ts
@@ -298,6 +298,56 @@ describe('grant_role gate — all-humans-trusted group channel', () => {
   })
 })
 
+describe('grant_role gate — peer bots', () => {
+  test('single-human group with a peer bot present is refused', async () => {
+    const dir = freshAgentDir()
+    // humans===1 but bots===2 (agent + one peer bot), so a peer bot's messages
+    // can still be buffered into the turn — not DM-equivalent.
+    const origin = trustedGroupWithMembership({ humans: 1, bots: 2, truncated: false })
+    const details = await run(makeTool(dir, origin), { role: 'guest', permission: 'subagent.spawn' })
+    expect(details.ok).toBe(false)
+    if (!details.ok) expect(details.error).toContain('another bot')
+    expect(readRoles(dir).guest?.permissions).toBeUndefined()
+  })
+
+  test('all-trusted group with a peer bot present is refused', async () => {
+    const dir = freshAgentDir()
+    const origin = trustedGroupWithMembership({
+      humans: 2,
+      bots: 2,
+      truncated: false,
+      humanMemberIds: ['U_TRENT', 'U_OWNER'],
+    })
+    const details = await run(makeTool(dir, origin), { role: 'guest', permission: 'subagent.spawn' })
+    expect(details.ok).toBe(false)
+    expect(readRoles(dir).guest?.permissions).toBeUndefined()
+  })
+
+  test('github channel is refused (collaborator membership is not a no-peer-bot proof)', async () => {
+    const dir = freshAgentDir()
+    const permissions = createPermissionService({
+      roles: { trusted: { match: [{ kind: 'channel', platform: 'github' }] } } as unknown as RolesConfig,
+    })
+    const tool = createGrantRoleTool({
+      agentDir: dir,
+      getOrigin: () => ({
+        kind: 'channel',
+        adapter: 'github',
+        workspace: 'acme/repo',
+        chat: '1',
+        thread: null,
+        lastInboundAuthorId: 'U_TRENT',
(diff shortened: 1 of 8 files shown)


devxoul:
Thanks — both findings were valid and are addressed in 0e457bb.

**[HIGH] peer bots.** You're right: the single-human (and all-humans-trusted) branches checked humans but never bots, so a channel with one human + peer bots still buffered third-party bot messages into the turn. Fixed with a shared `provesOnlyAgentBotPresent` precondition required by both branches: a non-DM channel must have a fresh, non-truncated membership with `bots === 1`. Since the agent's own bot is always a channel member and never an inbound author (adapters drop self-authored messages), a complete (`!truncated`) read with exactly one bot proves the only non-human present is the agent itself — matching the DM baseline `{ humans: 1, bots: 1 }`. A peer bot pushes `bots >= 2` (or trips the human checks if misclassified as human) — fail-closed either way.

**GitHub caveat.** `membership.bots === 1` is *not* a valid no-peer-bot proof for GitHub: its membership is the repo collaborator list, a different population from the authors who can comment into a PR/issue turn (and the agent App is usually not a collaborator). GitHub is therefore excluded from the relaxation entirely and its resolver no longer emits `humanMemberIds`; GitHub grants stay on the TUI/DM path.

**[MEDIUM] tests.** Added the adversarial cases: single-human-with-peer-bot refused, all-trusted-with-peer-bot refused, and github refused — locking in the authorization boundary against regression.

PR #542, opened 2026-06-01
feat(channels): make stickiness group-aware so the bot observes in busy rooms

devxoul:
## Background

When the agent was engaged in a **group chat**, it replied to every single message from anyone it had recently replied to. The mechanism is sticky credit: after the bot replies, `grantStickyForReplyTargets` grants each reply-target author a time-windowed credit (default 15 min). On that author's *next* message, `decideEngagement` consumes the credit and engages — regardless of content — then the bot replies and a fresh credit is granted, looping for the whole window.

That tight back-and-forth is exactly what you want in a 1:1 DM. In a multi-human room it makes the bot answer chatter that isn't addressed to it, which reads as spammy and crowds out the humans.

## Summary

Make sticky credit **context-aware** instead of adding a new gate.

- `decideEngagement` still **consumes** sticky credit whenever it's present (it stays one-shot — important so a later membership change can't resurrect stale conversational credit), but it only lets the credit **force `engage`** outside a multi-human group. "Multi-human group" = `isMultiHumanGroup(isDm, effectiveHumans)` → not a DM and `effectiveHumans > 1`.
- In a group, sticky alone no longer wakes the bot. The author re-engages through the **same symmetric triggers humans and peer bots already share** — mention / reply / alias. DMs and solo-human channels keep the chatty continuation.
- **Why this and not a new suppressor / a `peerBotTriggers` knob:** the subsystem's pinned philosophy is "fix over/under-engagement with `trigger` + `alias`, not new gates; no bot-only knobs." This change narrows an *existing* permissive rule in the one context where it misfires, and keys off a condition that treats humans and peer bots identically. It is not a bot-specific gate.
- On the turns the bot **is** woken in a multi-human group, `composeTurnPrompt` appends a group-chat nudge so the model self-selects: reply only when addressed/relevant, else `NO_REPLY` / `skip_response`. It reuses the existing fenced `**[SYSTEM MESSAGE — not from a human]**` convention (cache-neutral user-turn suffix, not the system prompt) and is suppressed when the loop guard already fired, so a turn never stacks two silence notices.
- Router and engagement share **one** `isMultiHumanGroup` definition, so sticky suppression and the prompt nudge can never disagree.

## What this does NOT do

- Does **not** change DM or solo-human behavior — sticky back-and-forth there is unchanged.
- Does **not** add any config field; this is the new default (per product decision).
- Does **not** firewall peer bots behind extra triggers — bot-to-bot remains first-class via the same triggers as humans.
- Does **not** touch the system prompt or the prompt-prefix cache contract (verified by `scripts/dump-system-prompt.test.ts`).
- Does **not** alter the peer-bot loop guard's accounting (observed, sticky-suppressed messages were already non-engaging and don't increment it).

## Review notes

- Load-bearing change: the sticky condition in `decideEngagement` (`… && ledger.consume(...) && !multiHumanGroup`). Note the ordering — `consume()` runs whenever stickiness is on and a credit exists, *then* `multiHumanGroup` decides engage-vs-observe, so the credit is always spent. Tests `sticky credit does NOT engage in a multi-human group, but IS consumed` and `a consumed-in-group credit does not resurrect a later follow-up` pin this.
- One existing test (`sticky credit still bypasses high membership counts`) encoded the old behavior; it's been reframed to the solo-human case and the group case moved into the new `(group-aware sticky)` block.
- The router regression `observed peer-bot messages do not increment the guard` now asserts on the loop-guard-specific text rather than the shared `[SYSTEM MESSAGE]` marker, because the group nudge legitimately shares that marker.
- Docs updated: `docs/content/docs/internals/engagement.mdx` (engage ladder rule #2, new "Group-chat nudge" section, rules-of-thumb clarification that narrowing an existing symmetric rule is allowed).

### DIFF
--- a/docs/content/docs/internals/engagement.mdx
+++ b/docs/content/docs/internals/engagement.mdx
@@ -24,7 +24,7 @@ Every inbound that survives the `channel.respond` permission gate is tagged with
 `decideEngagement` is a priority ladder. The first matching rule wins; falling off the bottom yields `observe`.
 
 1. **Explicit triggers** (config `trigger`): `dm`, `mention` (platform `<@id>`), `reply` to a bot message. Any one engages.
-2. **Sticky credit**: if `stickiness` is on and the author holds an unexpired credit in the `StickyLedger`, engage and consume it. Credits are granted to the authors a bot reply targets (`grantStickyForReplyTargets`), so a back-and-forth stays engaged without re-mentioning.
+2. **Sticky credit**: if `stickiness` is on and the author holds an unexpired credit in the `StickyLedger`, **consume** it — but it only forces `engage` outside a multi-human group (`isMultiHumanGroup`: not a DM and `effectiveHumans > 1`). Credits are granted to the authors a bot reply targets (`grantStickyForReplyTargets`), so a 1:1 / DM / solo-human back-and-forth stays engaged without re-mentioning. In a multi-human group the credit is still spent (it stays one-shot, so a later membership change can't resurrect stale conversational credit) but does **not** by itself wake the bot — the author must re-address it via mention / reply / alias. Sticky is a dyadic continuation affordance, not a multi-human group auto-reply mandate.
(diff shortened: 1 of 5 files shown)


devxoul:
@typeey do without subagent

PR #547, opened 2026-06-01
feat(github): drop the decoy reviewer after a PR review (App auth)

devxoul:
## Background

Under GitHub App auth, the agent is woken to review a PR by requesting a **decoy user account** (named after the App slug) as a reviewer. A GitHub App can't itself be a `requested_reviewer`, so the decoy is the only way to direct a review request at the App. When the agent posts its formal review, GitHub auto-adds the App account to the PR's reviewers list — recording the reviewed state under the App identity — but the **decoy** account stays pinned in the requested-reviewers list as a perpetual "review requested", as if the review never happened. Every reviewed PR accumulates a stale decoy entry. This PR does the natural "I'm done, drop me" cleanup a human reviewer gets for free.

## Summary

Adds a new step 6 to the PR review flow in the `typeclaw-channel-github` skill: after the review is confirmed posted (step 5), issue a DELETE against the PR's `requested_reviewers` for the decoy login before closing the turn (now step 7). The decoy login is derived from `selfLogin` by stripping the trailing `[bot]` suffix — matching the derivation in `resolveDecoyReviewerLogin` in `inbound.ts`.

```sh
gh api -X DELETE /repos/owner/repo/pulls/<N>/requested_reviewers -f 'reviewers[]=<decoy-login>'
```

Scope is deliberately narrow:

- **App auth, explicit review-request path (path A) only.** PAT auth has no decoy — the bot is a real user GitHub keeps listed, and self-removal is not wanted. Plain-language asks (path B) and team requests never place a decoy on the requested-reviewers list, so the DELETE would be a no-op or 404.

The removal is safe under the adapter's existing self-loop guard: the `DELETE` is issued authenticated as the App, so the `review_request_removed` webhook GitHub emits carries the bot actor as its `sender`, which `classifyReviewRequest` drops. The cleanup never wakes a fresh session. The "Self-loop safety" section in the skill is updated to document that the guard covers both add and remove.

A failed DELETE is non-fatal — the review already landed. No retry loop, no PR-thread noise.

## Preview

_No runtime behavior change for PAT auth or plain-language review requests. Under App auth + explicit request, the decoy entry disappears from the requested-reviewers list once the agent's review posts._

## What this does NOT do

- Does not change PAT-mode behavior. PAT agents reviewing PRs are unaffected.
- Does not touch `resolveDecoyReviewerLogin`, `inbound.ts`, or any production TypeScript. This is a skill-only change (SKILL.md).
- Does not retry or surface a failed DELETE to the PR thread.
- Does not affect plain-language review asks (path B) or team-requested reviews.

## Review notes

The load-bearing addition is step 6 in the PR review flow and the corresponding update to the "Self-loop safety" section at the bottom of the skill. Verify: (1) the DELETE call uses the correct decoy-login derivation (strip `[bot]` from `selfLogin`), (2) the scope guards (PAT skip, path B skip, team-request skip) are stated precisely, and (3) the self-loop explanation correctly describes the `sender.login` match that keeps the cleanup silent.

**Verification**

- `bun run typecheck` — clean
- `bun run lint` — 0 errors (64 pre-existing warnings in unrelated test files)
- `bun test src/channels/adapters/github/` — 192 pass / 0 fail


devxoul:
@typeey don't lie

devxoul:
@typeey review again

devxoul:
@typeey what's wrong?

PR #550, opened 2026-06-01
feat(permissions): cap subagent_output/cancel to the requester's role

devxoul:
## Background

In a GitHub channel session we hit `subagent.output denied: insufficient permissions` when the agent tried to read the result of a background `reviewer` subagent it had spawned moments earlier. The spawn succeeded; the read did not.

Root cause is a provenance gap, not a missing permission string. `subagent_output`/`subagent_cancel` gate on `permissions.has(getOrigin(), …)`, and in a channel session `getOrigin()` is the *current turn's* origin. For a reminder-only turn (the subagent-completion `<system-reminder>`), the router restores `lastInboundAuthorId` to the last human speaker. On GitHub App-auth that speaker resolves to `guest`, and `guest` holds no subagent permissions — so the agent was blocked from reading its own subagent.

The operator's decision was **not** to make the parent always able to read its own subagent (that reopens a hole in shared channels), but the inverse: **cap these tools to the requester's role.** A requester must not reach a subagent spawned by a higher role than they currently resolve to. A `guest` turn cannot see or cancel a `member`/`trusted`/`owner`-spawned subagent — even if `guest` has been granted `subagent.output`/`subagent.cancel`.

## Summary

Two checks now apply to `subagent_output` and `subagent_cancel`: the existing "holds the permission" check, **and** a new "requester role ≥ spawner role" cap. The cap is enforced via a shared `denySubagentAccess` helper so both tools stay identical.

- **Why a severity rank and not permission-set superset?** A high-trust custom role with intentionally narrow permissions would fail superset comparison and be unable to reach a `member`-spawned subagent it should reach. A fixed tower avoids that surprise: `owner > trusted > (custom) > member > guest`, with every configured custom role at one tier between trusted and member (matching the existing resolver contract that custom roles slot there).
- **Why fail closed?** This is security provenance. A missing `spawnedByRole` (e.g. a spawn with no permission service) or an unknown role on either side denies rather than allows.
- **No `parentSessionId` exemption.** A channel session contains turns from different authors/roles; trusting "same session" as a privilege boundary would reopen exactly the guest-reading-member-spawned hole. The cap is strict.
- **Provenance source.** `spawn_subagent` already resolves the spawner's role; it now also persists it on the `LiveSubagent` registry entry so the tools can compare.

## What this does NOT do

- Does not change the "agent reads its own background reviewer" UX gap on its own. Under the strict cap, a reminder-only turn whose last speaker is a guest will still be denied — that is the intended, operator-requested behavior. If "agent-owned background tasks" is wanted later, it needs a separate stored-principal model, not a session-identity bypass.
- Does not alter `subagent.spawn` gating or the role tower itself.
- Does not touch the channel router's reminder-turn author restoration.

## Review notes

- Load-bearing invariant: `denySubagentAccess` (`src/agent/tools/subagent-access.ts`) must compose **both** the `has(...)` check and the severity cap, and must treat `compareRoleSeverity === undefined` as deny. If a future edit lets `undefined` fall through to allow, the privilege-escalation hole reopens.
- `compareRoleSeverity` returns `undefined` for unknown roles by design; the only caller treats that as deny.
- The two tools resolve the live entry and surface `Unknown task_id` *before* the cap, so a too-low requester cannot probe task existence via the denial message.
- `noopPermissionService` and the inline `PermissionService` stubs across the suite gained the new interface member; the grant-all stubs return `1` (requester always wins) to preserve their "allow everything" intent.

### DIFF
--- a/src/agent/live-subagents.ts
+++ b/src/agent/live-subagents.ts
@@ -19,6 +19,10 @@ export type LiveSubagent = {
   sessionId: string
   subagentName: string
   parentSessionId?: string
+  // Role that resolved at spawn time, captured for the provenance cap on
+  // subagent_output/subagent_cancel. Absent when no permission service was
+  // active at spawn, in which case the cap fails closed.
+  spawnedByRole?: string
   startedAt: number
   status: SubagentStatus
   completion?: SubagentCompletion
--- a/src/agent/tools/spawn-subagent.test.ts
+++ b/src/agent/tools/spawn-subagent.test.ts
@@ -299,6 +299,7 @@ describe('createSpawnSubagentTool — permissions gating', () => {
     return {
       has: (_origin, permission) => allowed.has(permission),
       resolveRole: () => 'tester',
+      compareRoleSeverity: () => undefined,
       describe: () => ({ role: 'tester', permissions: [...allowed] }),
       replaceRoles: () => {},
     }
@@ -466,6 +467,7 @@ describe('createSpawnSubagentTool — role inheritance', () => {
     return {
       has: () => true,
       resolveRole: () => role,
+      compareRoleSeverity: () => undefined,
       describe: () => ({ role, permissions: ['subagent.spawn'] }),
       replaceRoles: () => {},
     }
--- a/src/agent/tools/spawn-subagent.ts
+++ b/src/agent/tools/spawn-subagent.ts
@@ -129,6 +129,7 @@ export function createSpawnSubagentTool(options: CreateSpawnSubagentToolOptions)
         sessionId: resolvedHandle.sessionId ?? '<pending>',
         subagentName,
         parentSessionId,
+        ...(spawnedByRole !== undefined ? { spawnedByRole } : {}),
         startedAt,
         status: 'running' as const,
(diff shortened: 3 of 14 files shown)


devxoul:
Fixed in e0cdc75. Both tools now route through a single `authorizeLiveSubagentAccess` helper that closes the existence oracle:

- The task-independent base-permission check runs **before** any registry lookup, so a caller lacking `subagent.output`/`subagent.cancel` gets the same `denied: insufficient permissions` whether or not the task exists.
- For a caller that holds the base permission but is **not** authorized for the task, an absent task ID, a present-but-capped task, and a present task with missing provenance all collapse to one identical denial: `denied: unknown task_id or insufficient role`. A lower-role caller can no longer distinguish "task exists (capped)" from "task absent".
- Only `owner` (which outranks every possible spawner) still receives a truthful `Unknown task_id` for genuine typos, and bypasses the cap.

Added behavior tests on both tools asserting non-distinguishability across absent/capped/missing-provenance for a low-role caller, plus the owner truthful-unknown path. Same applied to `subagent_cancel` (abort is not invoked on a capped deny).

PR #551, opened 2026-06-01
fix(channels): scope Discord membership to the channel, not the whole guild

devxoul:
## Background

PR #537 relaxed the `grant_role` guard to allow role grants in a **single-human** or **all-humans-trusted** group channel, proven from a membership snapshot. The guard helper `provesOnlyAgentBotPresent` requires `membership.bots === 1` — the only non-human in the room is the agent's own bot — so no peer bot can buffer a prompt-injection message into the turn.

That relaxation never fires on Discord in a common real setup. The Discord membership resolver enumerated **guild-wide** membership via `GET /guilds/{id}/members` and never filtered by who can actually view the target channel. In any guild that hosts **more than one bot**, the guild-wide count reports `bots > 1` even when the *specific channel* is just operator + agent bot — so the guard false-refuses with "this is a mixed-speaker surface".

### Confirmed in production

A private channel with `@everyone` denied `VIEW_CHANNEL` and only the agent bot allowed:

- **Guild membership** (what the resolver returned): `1 human + 3 bots` → `bots !== 1` → refused.
- **Channel membership** (the real authorization surface): `1 human (owner, bypasses overwrites) + 1 bot (agent)` → the two peer bots are denied `VIEW_CHANNEL` and cannot post → should allow.

The peer bots are not a prompt-injection surface for that channel, so counting them is wrong.

## Summary

Filter the guild member list down to members who can `VIEW_CHANNEL` the target channel, then count humans/bots over that visible set. Visibility follows Discord's documented permission-overwrite resolution order:

1. base = `@everyone` role perms `OR` all of the member's role perms
2. `owner_id` or `ADMINISTRATOR` short-circuits to visible
3. `@everyone` channel overwrite (clear deny bits, then set allow bits)
4. all matching role overwrites combined (OR every deny, OR every allow; apply denies then allows)
5. member-specific overwrite (clear deny, set allow)
6. visible iff `VIEW_CHANNEL` is set

All bit math is **BigInt** because Discord serialises permission sets as decimal strings beyond `Number.MAX_SAFE_INTEGER`.

This matches **Slack**, which already scopes membership to the channel via `conversations.members`. The guard's `bots === 1` invariant is unchanged — only the data feeding it is corrected.

Data fetched on the enumeration path: `GET /channels/{chat}` (overwrites + channel type) and `GET /guilds/{id}` (`owner_id` + `roles[].permissions`), in parallel, in addition to the existing preview + members calls.

## Fail-closed

Any inability to compute channel visibility derives from history instead (`truncated: true`), keeping every security consumer fail-closed:

- channel or guild fetch failure
- a member referencing a role absent from the guild role map (`memberCanViewChannel` returns `null`)
- a **thread** channel (its visibility comes from the parent channel + an explicit member list, which this resolver does not model — refusing to treat it as world-visible)

`humanMemberIds` stays channel-scoped, and its `length === humans` invariant is preserved by construction; an unidentifiable visible human drops it to counts-only.

## What this does NOT do

- Does not change the `grant_role` guard or its `bots === 1` / `humanMemberIds` contract.
- Does not change Slack, Telegram, KakaoTalk, or GitHub resolvers.
- Does not merge parent-category overwrites for normal channels (synced categories already copy their overwrites onto the channel; unsynced channels intentionally diverge).
- Does not model private-thread membership — threads fall back to history.

## Origin prose

Exact Discord counts are now channel-scoped, so the origin membership summary's "count of guild members" caveat is replaced with "counts only members who can view this channel". The caveat is gated to the **exact** branch; truncated (history-derived) summaries drop it, since recent-speaker counts are not a channel-membership claim.

## Tests

New `createDiscordMembershipResolver` cases:
- **the exact production shape**: `@everyone` denied `VIEW_CHANNEL`, only the agent bot allowed → `{humans:1, bots:1}` (the single-human relaxation can fire)
- `ADMINISTRATOR` role bypasses a channel deny overwrite
- public channel (no deny) → channel count equals guild count
- fail-closed: unknown-role member → history fallback
- fail-closed: thread channel → history fallback
- fail-closed: channel fetch 403 → history fallback
- unidentifiable visible human → counts-only (no `humanMemberIds`)

Existing enumeration tests updated to supply the channel + guild objects and assert the two new scoped fetches. Origin summary tests updated for the new caveat wording.

## Checks

`typecheck` clean, `lint` 0 errors, `format:check` clean. Full suite green except the one pre-existing environmental failure in `src/update/index.test.ts` (`resolveSelfPackageJsonPath`) on worktree paths — passes on a normal checkout/CI, same as noted in #537.

### DIFF
--- a/src/agent/session-origin.test.ts
+++ b/src/agent/session-origin.test.ts
@@ -469,7 +469,7 @@ describe('renderSessionOrigin', () => {
     expect(out).not.toContain('guild members')
   })
 
-  test('channel origin renders Discord guild-level caveat for exact counts', () => {
+  test('channel origin renders Discord channel-scoped caveat for exact counts', () => {
     const now = 100_000
     const out = renderSessionOrigin(
       {
@@ -484,7 +484,7 @@ describe('renderSessionOrigin', () => {
     )
 
     expect(out).toContain('This channel has 47 members: 12 humans, 35 bots.')
-    expect(out).toContain('this is the count of guild members')
+    expect(out).toContain('counts only members who can view this channel')
   })
 
   test('channel origin renders large-channel truncated summary', () => {
@@ -501,7 +501,7 @@ describe('renderSessionOrigin', () => {
     expect(out).toContain('exceeds the 50-member cap')
   })
 
-  test('channel origin adds Discord caveat to truncated summaries', () => {
+  test('channel origin omits the channel-scoped caveat on truncated (history-derived) summaries', () => {
     const out = renderSessionOrigin({
       kind: 'channel',
       adapter: 'discord-bot',
@@ -512,7 +512,7 @@ describe('renderSessionOrigin', () => {
     })
 
     expect(out).toContain('approximately 200 members')
-    expect(out).toContain('private channels with permission overwrites')
+    expect(out).not.toContain('counts only members who can view this channel')
   })
 
   test('channel origin omits member summary when membership is missing', () => {
--- a/src/agent/session-origin.ts
+++ b/src/agent/session-origin.ts
@@ -400,15 +400,19 @@ function renderMembershipSummary(
   if (membership === undefined) return null
 
   const total = membership.humans + membership.bots
-  const caveat =
-    origin.adapter === 'discord-bot' && origin.workspace !== '@dm'
-      ? ' (Note: this is the count of guild members; private channels with permission overwrites may have fewer actual viewers.)'
-      : ''
+  // Exact Discord counts are channel-scoped (filtered by who can VIEW_CHANNEL),
+  // so the count is the channel's room, not the guild. The truncated branch is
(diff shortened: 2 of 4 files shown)


devxoul:
Good catch on the thread path — addressed in 2bea35e.

`scopeMembersToChannel` now scopes on `key.thread ?? key.chat`, mirroring the history callback's channel-id resolution. Today Discord always stores a thread's own snowflake in `chat` with `thread` null, so the misscoping was latent, but the resolver now correctly evaluates the thread channel (not the parent) if a future caller passes `chat=parent, thread=threadId`.

Added a regression test (`scopes visibility to key.thread when set, not the parent key.chat`) asserting the channel object fetched is the thread id, and that the parent is *not* fetched, when `thread` is set.

PR #557, opened 2026-06-01
feat(channels): drop the engage :eyes: reaction after the agent replies

devxoul:
## Background

#549 made the runtime drop an `:eyes:` reaction on the triggering item the moment it engages, replacing the old "On it" ack comment. But the reaction is only half a status indicator: nothing ever removed it. Every issue/PR/comment the agent answered stayed marked `:eyes:` forever, so `:eyes:` came to mean "was engaged at some point" rather than "currently being looked at" — the signal a maintainer actually wants is "looking" → (reply) → "done looking".

This PR closes that loop: once the agent has actually replied in a turn, the engage `:eyes:` is removed.

## Summary

The hard constraint is GitHub's API: there is no delete-by-content endpoint, so removal requires the numeric reaction id, which is only available in the `POST` response body at add time. The add and the reply are far apart in the turn lifecycle, and the add is fire-and-forget.

Key decisions:

- **Carry the id, not re-list.** The add now parses the reaction id from the `POST` body and returns it encoded in an opaque per-instance `ReactionRef` on the success result (consistent with the existing opaque-ref pattern). Removal decodes it and issues `DELETE`. We never re-list reactions to find our own.
- **Separate remove type, not an `action` flag.** Add and remove need different data (target ref vs. instance id), so `RemoveReactionRequest`/`RemoveReactionCallback` are distinct. Additive reactions stay backward-compatible and adapters without removal stay no-ops. The removal ref uses an `op:'remove'` discriminator so it can never be decoded as an add target ref (and vice versa).
- **Remove on reply, at the per-turn boundary.** The trigger is "this turn produced a successful channel send", evaluated in the per-turn finally inside `drain()` — not the outer multi-turn finally. This means multi-part replies don't drop `:eyes:` after chunk 1, and skip/no-reply/reminder-only turns keep it.
- **Ordered off the in-flight add.** A fast reply can resolve before the engage `POST` does. Removal is chained off the add promise (threaded through the queued inbound so it binds to the exact answered inbound), so it can never race ahead and no-op against a not-yet-created reaction.

## What this does NOT do

- Does not add reaction support to Slack/Discord. They register no removal callback and stamp no `reactionRef`, so they remain silent no-ops — GitHub is still the only adapter with reactions.
- Does not touch reactions created by the explicit `channel_react` tool. Only the engage-added reaction is auto-removed.
- Does not block the reply on reaction I/O. Removal is best-effort fire-and-forget; failures only log.

## Review notes

- Load-bearing change is in `drain()`: `sentReplyThisTurn = live.successfulChannelSends > successfulSendsBeforePrompt` gates removal, and `dropEngageReactionAfterReply` chains off the snapshotted engage add-promise. Verify the gate excludes skip/no-reply turns and that the chaining preserves add→remove ordering (covered by the "orders removal after the in-flight add resolves" test).
- `not-found` (reaction already gone) and `unsupported` are treated as non-noisy on the remove path. A `200` "already existed" `POST` with no usable id yields a success with no removal ref, so removal cleanly no-ops rather than guessing.


devxoul:
Good catch — fixed in a6bb470. The turn now collects an engage-add promise per inbound in the coalesced batch (`currentTurnEngageReactions`, via `batch.flatMap`) and removes every one after the reply, instead of keying off only `batch[batch.length - 1]`. Added a router test that batches two inbounds with distinct `:eyes:` instances and asserts both are removed; it fails against the old last-only logic.

PR #558, opened 2026-06-01
docs(recipes): add Recipes section with Code reviewer page

devxoul:
## Background

The docs had no goal-oriented walkthrough surface for topics that stitch multiple features together — they lived entirely in reference pages and individual guide entries. A user who wants to set up a complete workflow like PR code review has to find and mentally assemble three or four separate pages on their own.

The second gap: MDX pages that include architectural diagrams had no native rendering path. Any Mermaid block was either skipped or copy-pasted as an image, making diagrams a manual, drift-prone artifact.

## Summary

**Commit 1 — server-side Mermaid rendering.**
Add `beautiful-mermaid` and a server-rendered `Mermaid` MDX component, wired via `remarkMdxMermaid` from `fumadocs-core`. Any `` ```mermaid `` fenced block in MDX is converted to a themed SVG at build time — no client runtime needed. Colors are wired to the Fumadocs CSS custom properties (`--color-fd-background`, `--color-fd-foreground`, `--color-fd-primary`, etc.) so the diagram automatically respects light and dark themes. If `renderMermaidSVG` throws on a malformed graph, the component falls back to rendering the source as a code block — fail-open for content, never a build error.

Key decisions:
- **Why server-render, not a client component?** No hydration overhead, no layout shift, no client JS bundle contribution. The diagram is static SVG by the time it reaches the browser.
- **Why `beautiful-mermaid` and not `mermaid` + a rehype plugin?** `beautiful-mermaid` exposes a pure synchronous `renderMermaidSVG` call that works in Node at build time without a headless browser. The official `mermaid` package requires a DOM.
- **Why inline CSS vars as theme values?** Fumadocs does not expose Mermaid theme tokens via its theme config; passing the CSS var strings lets the SVG's `style` attributes inherit the correct values at render time without any build-time color extraction.

**Commit 2 — Recipes section + Code reviewer recipe.**
Add a new top-level Recipes section (placed between Guides and Concepts in `content/docs/meta.json`) for goal-oriented, link-heavy walkthroughs that assemble multiple existing features. The first recipe documents the live production setup for PR code review.

## What's in the Code reviewer recipe

The recipe documents the exact setup `@typeey` — TypeClaw's own mascot agent — uses on `typeclaw/typeclaw`: request it as a reviewer on a pull request, it spawns the bundled `reviewer` subagent to do line-anchored analysis, and posts a formal GitHub review back. Nothing human-triggered after the initial wire-up.

Covered in order:

1. **Wire the GitHub channel** via `typeclaw channel add github` — the CLI wizard handles auth (PAT or App), tunnel provisioning, webhook installation, and repo scoping.
2. **Scope events** to the review-request path via `channels.github.eventAllowlist`.
3. **The decoy-reviewer pattern** — a GitHub App cannot be requested as a reviewer directly. A machine account named after the App slug stands in; requesting that account fires the `review_requested` webhook with the App's auth. Explained in prose, a summary table, and a how-it-works diagram.
4. **Output paths** — conversational comments go through `channel_reply`; formal reviews go through `gh api .../pulls/N/reviews`. The recipe shows the single-quoted body quoting required to avoid shell interpolation.
5. **Gotchas** — `gh` may be unauthenticated inside the container by default; the `GH_TOKEN` workaround is called out explicitly.
6. **Role gating** — `grant_role` can restrict reviewer requests to trusted or owner-tier authors.

## Diagram

The how-it-works flow in the recipe (rendered as SVG at build time via the new Mermaid component):

```
Request the decoy reviewer → GitHub channel adapter → reviewer subagent → Formal review posted to PR
```

Left-to-right `graph LR`, themed to match the docs light/dark palette.

## What this does NOT do

- Does not change any runtime behavior, CLI commands, or agent logic.
- Does not add any new channel-adapter configuration options — the recipe documents existing knobs only.
- Does not add a client-side Mermaid fallback; SVG renders at build time or falls back to the raw source code block.
- Does not gate the Recipes section behind any feature flag — it ships as a publicly visible docs section.

## Review notes

- The `Mermaid` component uses `dangerouslySetInnerHTML` to inject the SVG string. The input is the chart source from the author's own MDX, not from user input, so there is no XSS surface here. `renderMermaidSVG` produces a self-contained SVG with no `<script>` tags.
- `remarkMdxMermaid` is added to `source.config.ts` `remarkPlugins` — it runs before the MDX component registry, so the `Mermaid` export in `mdx-components.tsx` is what the remark plugin's output resolves to.
- `meta.json` ordering: `"guides"` → `"recipes"` → `"concepts"` → `"reference"`. The Recipes section appears between Guides and Concepts in the sidebar intentionally — recipes are more task-specific than guides but less structural than concepts.

## Testing

- `bun run lint` — passes (only pre-existing `page.tsx` warnings, not introduced here).
- `bun run format:check` — clean.
- `bun run build` — TypeScript clean; all 46 static pages generated. Server-side Mermaid render runs at build time with no errors.
- Dev server — diagram and page verified visually; light/dark theme switching tracked correctly via the CSS var wiring.

### DIFF
--- a/docs/bun.lock
+++ b/docs/bun.lock
@@ -6,6 +6,7 @@
       "name": "@typeclaw/docs",
       "dependencies": {
         "@types/mdx": "^2.0.13",
+        "beautiful-mermaid": "^1.1.3",
         "fumadocs-core": "^16.5.0",
         "fumadocs-mdx": "^14.2.6",
         "fumadocs-ui": "^16.5.0",
@@ -411,6 +412,8 @@
 
     "baseline-browser-mapping": ["baseline-browser-mapping@2.10.31", "", { "bin": { "baseline-browser-mapping": "dist/cli.cjs" } }, "sha512-MujYO3eP72uvmSE0i4wltsodRfIpZATP3jvzRNRGGxgzId7aVocVJJV3nf01qnzzKFGxQVC9bpWxl5cjxTr/7Q=="],
 
+    "beautiful-mermaid": ["beautiful-mermaid@1.1.3", "", { "dependencies": { "elkjs": "^0.11.0", "entities": "^7.0.1" } }, "sha512-TItrtrAyHp1vwFfFVYauWGrquouk/6SS21Aq3RsxindSYZODcN4xYrPZD6BiZRU+o5mKJzDPz9MUSMvELdylyg=="],
+
     "caniuse-lite": ["caniuse-lite@1.0.30001793", "", {}, "sha512-iwSsYWaCOoh26cV8NwNRViHlrfUvYsHDfRVcbtmw0Kg6PJIZZXwMkj1442FYLBGkeUf1juAsU3DTfxW579mrPA=="],
 
     "ccount": ["ccount@2.0.1", "", {}, "sha512-eyrF0jiFpY+3drT6383f1qhkbGsLSifNAjA61IUjZjmLCWjItY6LB9ft9YhoDgwfmclB2zhu51Lc7+95b8NRAg=="],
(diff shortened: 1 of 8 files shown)


devxoul:
Fixed in 8d38a4e — thanks @typeey. Added `pull_request.review_request_removed` back into the narrowed event allowlist, with a note explaining why to keep it (so the agent still sees when a review request is withdrawn). The full default allowlist already included it; the omission was only in the "narrow it to just the review path" example.

PR #567, opened 2026-06-01
Add allowApprove flag to github channel config

devxoul:
## Background

When the github channel reviews a PR, the agent currently maps the reviewer subagent's `approve` verdict straight to a formal GitHub `APPROVE` review. There was no way for an operator to allow the agent to *review* PRs (post line-anchored findings) while withholding the authority to *approve* them — a common policy split for a bot account that should comment but never be a green-checkmark approver.

This adds `channels.github.allowApprove` (boolean, **default `true`**, so existing behavior is unchanged). When set to `false`, the agent reviews and comments as before but downgrades any `approve` verdict to a `COMMENT` review — findings still posted, no formal approval.

## Summary

The gate is delivered as an **operator-policy note appended to inbound message text** by the github adapter, and consumed by the `typeclaw-channel-github` skill's verdict→event mapping.

**Why inbound text and not a bash-layer block (the obvious choice):** approval is posted via `gh api --input /tmp/review.json`, so the `event: "APPROVE"` value lives in a temp file. The `github-cli-auth` interceptor only sees the command string (not the file) and early-returns for classic PATs, so a reliable hard block at the bash layer isn't feasible without reading arbitrary temp files. The skill is the single chokepoint every review passes through, so steering the verdict→event mapping there is the reliable seam.

**Why ride the note on every inbound (only when disabled):** the agent is asked to review two ways — a `review_requested` webhook (clean synthetic inbound) *and* plain-language asks ("@bot review #123") that arrive on arbitrary issue/PR/discussion comments. A note attached only to `review_requested` would silently leave plain-language approvals ungated. Appending one line to every github inbound — gated entirely behind the non-default `false` — closes that hole with zero cost on the default (enabled) path.

## What this does NOT do

- Does not change the default: approval stays enabled unless an operator explicitly sets `allowApprove: false`.
- Does not hard-block approval at the API/bash layer — enforcement is via the skill honoring the policy note (see the bash-layer rationale above).
- Does not touch `request-changes` or `comment` verdicts; those never approve regardless.

## Review notes

- Load-bearing change: `withApprovalPolicy` in `src/channels/adapters/github/inbound.ts` (applied once at the `route()` seam) and the policy section added to the skill's verdict→event mapping.
- Invariant to verify: with `allowApprove` defaulted/true, inbound text is byte-identical to before (covered by the "does not append when allowed / when omitted" tests).
- `lifecycle.test.ts` / `manager.test.ts` carry one-line fixture updates only — the new schema field is required, so those `GithubAdapterConfig` literals must include it to compile.

### DIFF
--- a/src/channels/adapters/github/inbound.test.ts
+++ b/src/channels/adapters/github/inbound.test.ts
@@ -8,6 +8,7 @@ import {
   classifyGithubInbound,
   createGithubWebhookHandler,
   type GithubWebhookHandlerOptions,
+  PR_APPROVAL_DISABLED_NOTE,
   verifySignature,
 } from './inbound'
 import { decodeGithubReactionRef } from './reactions'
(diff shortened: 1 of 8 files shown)


devxoul:
Fixed in 1ad4dab. The zero-actionable-findings `approve` branch now submits a formal `COMMENT` review (summary as the review body, no `comments[]`) when approval is disabled, instead of a top-level issue comment — keeping the review metadata and matching the actionable-findings path.

Also restructured the config in fe8b57e: the gate now lives at `channels.github.review.approve` (nested under a `review` object) so future review knobs (`requestChanges`, etc.) have a home. Nothing is shipped yet, so no migration was needed.

devxoul:
@typeey please approve

PR #583, opened 2026-06-02
reviewer: bound spawn with a timeout so a stalled review can't strand the parent

devxoul:
## Problem

A `reviewer` subagent spawned to review a GitHub PR can stall mid-turn and **strand the parent forever**, with no error surfaced anywhere.

Observed in production: a `@bot review this PR` request spawned the reviewer in the background. Mid-review the model called a tool by a wrong name, got a `Tool ... not found` result, and then **wedged** — the subagent's final transcript entry is the errored `toolResult`, with no terminal assistant message and no `<review>` block.

Because `createReviewerSubagent()` declared **no `timeoutMs`**, the spawn's `completion` promise never settled (pi's `session.prompt` takes no `AbortSignal`, so a half-open stream stays alive). Per `src/agent/tools/spawn-subagent.ts`, the `subagent.completed` broadcast only fires from `completion.then(...)` — so it **never fired**. The parent channel session was never woken via the completion bridge, and **no review was ever posted**. The PR just went silent after the "reviewing now" status comment.

## Fix

Declare a wall-clock ceiling on the reviewer spawn (`REVIEWER_SPAWN_TIMEOUT_MS = 600_000`, 10 min). When exceeded, `awaitWithSubagentTimeout` settles the spawn with a `SubagentTimeoutError`, which surfaces to the parent as a **FAILED** completion reminder (`renderSubagentCompletionReminder`) instead of hanging silently. This is the same liveness mechanism the `memory-retrieval` subagent already uses (`timeoutMs: 30_000`).

This is **liveness for the parent, not hard cancellation** — the underlying LLM stream may keep running until the OS reaps it; the ceiling's job is to release the parent (and the coalescing key) so the request fails loudly.

## Sizing

10 min is generous for a thorough `deep`-model review of a large diff plus a few web lookups, well above the typical sub-minute review, well below "wait forever".

## Scope

This is the **liveness** half of the fix. The naming-slip *trigger* (model called `web_search`/`web_fetch` instead of `websearch`/`webfetch`) is addressed separately by a "did you mean?" nudge on unknown tool calls — intentionally a different PR.

## Tests

- `declares a spawn timeout so a stalled review fails the parent loudly instead of hanging forever`
- `spawn timeout is generous enough for a deep-model review but not unbounded`

Both assert behavior (a bounded, positive ceiling) rather than the literal constant, so they survive a value change but fail if the timeout is removed — guarding the exact regression this PR fixes.

## Checks

`bun run typecheck`, `bun run lint`, `bun run format:check` all pass. Reviewer suite: 50 pass. (Full-suite has one pre-existing, unrelated failure — `resolveSelfPackageJsonPath` asserts the cwd ends in `typeclaw/`, which fails only because the build ran from a `/tmp` worktree; it passes in isolation and is untouched by this change.)

### DIFF
--- a/src/agent/subagents.test.ts
+++ b/src/agent/subagents.test.ts
@@ -828,6 +828,72 @@ describe('startSubagent', () => {
     }
   })
 
+  test('timeoutMs settles completion with ok=false when prompt wedges (parent gets woken, not stranded)', async () => {
+    // given: a session whose prompt never resolves, and a subagent with a tiny timeout
+    let abortCount = 0
+    const session = {
+      prompt: () => new Promise<void>(() => {}),
+      dispose: () => {},
+      subscribe: () => () => {},
+      abort: async () => {
+        abortCount += 1
+      },
+    } as unknown as AgentSession
+    const registry = { greeter: { systemPrompt: 'X', timeoutMs: 20 } satisfies Subagent }
+
+    // when
+    const { completion } = startSubagent('greeter', {
+      registry,
+      createSessionForSubagent: async () => session,
+      agentDir: '/agent',
+      userPrompt: 'q',
+      taskId: 'bg_timeout',
+    })
+    const result = await completion
+
+    // then: the spawn fails (so spawn_subagent fires the FAILED broadcast) and the session is aborted
+    expect(result.ok).toBe(false)
+    if (!result.ok) {
+      expect(result.error).toContain('timed out')
+    }
+    expect(abortCount).toBe(1)
+  })
+
+  test('without timeoutMs a slow prompt keeps completion pending (legacy unbounded behavior preserved)', async () => {
+    // given: a session whose prompt has not resolved yet, and no timeout declared
+    let resolvePrompt: () => void = () => {}
+    const session = {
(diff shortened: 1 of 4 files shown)


devxoul:
You're right — the declared `timeoutMs` wasn't enforced on the `spawn_subagent` → `startSubagent` path (the one the GitHub review actually uses). `awaitWithSubagentTimeout` only wrapped the `SubagentConsumer` path. Fixed in aae7acd: `startSubagent` now races `invokeSubagent` against `registry[name].timeoutMs` and settles its `completion` promise with `{ ok: false, error: 'timed out' }` on expiry — which is exactly what drives `spawn_subagent`'s `subagent.completed` broadcast, so the parent gets a FAILED reminder instead of hanging forever. It also fires the session `abort` handle best-effort (pi's `session.prompt` takes no AbortSignal, so the LLM stream may run until the OS reaps it). Added two tests: one proving a wedged prompt + tiny `timeoutMs` settles `completion` with `ok=false` and aborts the session, and one proving the unbounded (no-`timeoutMs`) path stays pending as before.

PR #592, opened 2026-06-02
channels: forbid self-authored PR reviews in github skill

devxoul:
## Problem

A single PR review request can produce **two contradictory reviews** on the same PR. Observed in production on #583: the bot posted an `APPROVE` review at 17:08:25, then a top-level "I found a blocker" comment at 17:09:10 — 45s later, flatly contradicting its own approval.

### Root cause

The github channel skill delegates PR analysis to the `reviewer` subagent (it runs on the `deep` model and returns a line-anchored `<review>` block). The intended flow is **spawn → wait → translate → post**. But the skill also told the parent to "keep working" after spawning, with **no rule forbidding the parent from analyzing the same PR itself** while the reviewer ran.

In the incident, the parent (default model) did exactly that:

1. Spawned the `reviewer` in the background (correct).
2. While it ran, the parent ran `gh pr view` / `gh pr diff`, read the changed files, and formed its **own** opinion.
3. After two `subagent_output` polls still showed "Running…", the parent **posted its own `APPROVE` review** and ended the turn with `skip_response({ reason: "review posted via gh api" })`.
4. The reviewer finished a minute later; the completion `<system-reminder>` re-woke the parent.
5. The parent fetched the reviewer's actual output — a `blocker` finding — and **posted it too**, contradicting the approval it had already shipped.

Two independent gaps:

- **No exclusivity rule.** Nothing said "the reviewer's `<review>` is the *only* source of the verdict — don't review the PR yourself." "Keep working" read as license to do the reviewer's job on the parent model.
- **No idempotency on the completion turn.** The post-completion reminder fires unconditionally and invites a post, with nothing reconciling "I already shipped a verdict for this PR."

## Fix

Skill-level guardrails in `src/skills/typeclaw-channel-github/SKILL.md` (docs/skill only):

1. **Exclusivity** (top of PR review flow): the reviewer's `<review>` block is the only source of the verdict; the parent does **no analysis** of the PR between spawn and result — no `gh pr diff`, no reading changed files, no drafting a verdict.
2. **End the turn after spawning**: clarifies that "stay responsive" means *other* chats, not continuing to work *this* PR; no busy-wait polling.
3. **One verdict per PR per request**: if the completion reminder arrives and a verdict already went out this engagement, `skip_response({ reason: "review already posted for this PR" })` instead of posting a second, possibly-contradicting review.

## Scope

This is the **skill-wording** half — the direct cause and the glaring loophole. A durable **mechanical** idempotency guard (suppress the completion-reminder re-wake when the parent already self-resolved the review while its spawned reviewer was still in-flight; lands around `injectSubagentCompletionReminder` in `src/channels/router.ts`) is a separate follow-up, since it needs new per-engagement state and its own test coverage.

## Checks

`bun run format:check`, `bun run lint` (0 errors), `bun run typecheck` all pass. Full suite: 6799 pass, 1 skip, 1 fail — the fail is the pre-existing, unrelated `resolveSelfPackageJsonPath > points at this package manifest` (asserts cwd ends in `typeclaw/`, fails only because the build ran from a `/tmp` worktree; passes in the real checkout, untouched by this change).


devxoul:
Good catch — fixed in 703acf1. The guard was in step 7, downstream of the read/translate/post steps, so on a completion reminder the agent could reach the posting path before it ever fired. Moved it to the top of step 3 (the completion-path entry): the "have I already shipped a verdict this engagement?" check now runs *before* any `subagent_output` fetch, translation, or post, and short-circuits to `skip_response` when true. Step 7 is back to just the `skip_response`-vs-trace-reply mechanics.

devxoul:
@typeey approve the pr please

Issue #291, opened 2026-05-21
Self-restart handoff: agent says 'I'm back' after restarting its own container


devxoul:
> **UPDATE (priority correction):** This issue was originally written TUI-first, and the "Scoping concerns" below excluded channel sessions from restart-resume. That scoping was wrong. **TUI is not the priority surface — channels are the real workload, and channel sessions are an equal-priority target for restart-resume.** A self-restart that originated from a channel session SHOULD reopen that session, fire the post-restart turn, and resume any pending todos in that conversation. The "no stray channel posts" concern still holds for *non-originating* channels and for cron/subagent/system origins (those remain no-ops); it does not justify excluding the originating channel.
>
> Implemented for the channel case in #598 (restart-handoff v2 carries the origin; channel boot reopens the originating session and resumes). The acceptance criteria below are amended accordingly (see the struck items).

## Goal

Make self-initiated restart feel continuous — for the originating session, whether that's a TUI session **or a channel conversation**:

```
H: restart
Agent: Okay, restarting now.
(container exits, new container boots, ~1-2s)
Agent: I'm back. Where were we?
```

Today the `restart` tool (`src/agent/tools/restart.ts`) exits the container cleanly and hostd brings up a fresh one, but the user must manually re-run `typeclaw tui` and the new agent has no awareness that it just restarted itself. The originator/sibling distinction was anticipated at the broadcast layer (lines 22-29 of `restart.ts`) but the proactive new-container greeting was never wired.

Out of scope: mid-LLM-call resumability. AGENTS.md is explicit that "agentic work is not resumable mid-LLM-call" and the in-memory Stream cannot survive a process death. This issue only covers the clean-turn-boundary case where the agent itself initiated the restart.

## What already exists

- `restart` tool publishes a `container-restarting` broadcast with `originatingSessionId` and `restartedAt` before `setTimeout(exit, 500ms)`.
- Hostd `restart` RPC, ACK-then-execute, container relaunch — all working.
- Session JSONL transcripts on disk under `sessions/`, force-committed by the backup plugin.
- `SessionManager` already loads JSONL files for the explorer plugin.

## What's missing

Three pieces, in increasing difficulty:

### 1. Durable handoff state across the restart

The Stream is in-memory and dies with the process, so the new container has no signal that it was just restarted by itself vs. cold-started. Options:

[...]

PR #598, opened 2026-06-03
channels: resume pending todos after a channel-origin restart

devxoul:
## Background

A user in a Discord channel told the agent to add a todo ("greet after restart"), then said "restart". The agent wrote the todo and called the `restart` tool; the container restarted — and the agent came back silent. The pending todo was on disk but was never resumed, and no post-restart turn fired.

Container logs showed the restart completing normally. The todo entry (`typeclaw.restart-self`) was present in memory. No follow-up turn ever appeared in the channel.

## Root cause

The restart-resume + todo-continuation machinery was TUI-only by construction.

`buildRestartHandoffWiring` in `src/agent/index.ts` wrote no restart handoff for non-TUI origins. This was a deliberate prior scoping decision (issue #291) to avoid the agent posting an unprompted "I'm back" message into a channel. Only the websocket `open` handler in `src/server/index.ts` consumed the handoff and drove resume.

Channel sessions are created lazily on inbound (`ensureLive`). After a reboot no live channel session exists, no inbound arrives, the handoff is never consumed, and the channel todo-continuation path — which only runs at the tail of a live `drain()` loop — never fires.

## What changed

This PR is 6 commits; the loop-guard abort commit is included because it was on the branch and is orthogonal.

**`agent: add restart-handoff v2 schema with kind-aware consume`**

Handoff carries an `origin` discriminator — `{kind:'tui'}` or `{kind:'channel', key}` — so the next boot routes resume to the right subsystem. v1 files (TUI-only by construction) read forward as a tui origin.

`consumeRestartHandoff` gains an `accept` predicate and **restores** a handoff it declines. The TUI websocket path and the channel-startup path can each claim only their own kind without the first reader deleting the other's file.

**`agent: thread restart handoff origin through the restart tool`**

`createRestartTool` and `requestContainerRestart` accept a `handoffOrigin` and gate the v2 handoff write on it. Callers that omit it (cron/subagent/system, in-memory, ad-hoc test construction) write no handoff, exactly as before.

**`agent: emit a restart handoff for persisted channel sessions`**

`buildRestartHandoffWiring` now returns origin metadata for persisted TUI and channel sessions (still nothing for cron/subagent/system or in-memory sessions). This intentionally extends issue #291's scope — see _Behavior change_ below.

**`server: claim only TUI restart handoffs on websocket open`**

The websocket open handler consumes with a tui-only predicate, leaving a channel handoff on disk for the channel startup path rather than deleting it on the first TUI reconnect.

**`channels: resume a channel-origin restart at boot`**

New `ChannelRouter.resumeRestartHandoff`, driven once from `src/run/index.ts` after `channelManager.start()`. It reopens the exact originating session — repairing the persisted `channels/sessions.json` mapping to bypass stale-rollover (a restart easily outlasts the 5-min freshness TTL, which would otherwise spawn a fresh session and lose the `typeclaw.restart-self` entry) — arms the one-shot restart-kick suppressor, and wakes the session via a fenced SYSTEM-MESSAGE nudge. On a stale/missing mapping, unconfigured adapter, or reopen failure it logs and skips; the todo still resumes on the next real inbound.

**`agent: abort the turn when the loop guard blocks, instead of only refusing`** (orthogonal, included on this branch)

Loop-guard `block` verdicts previously surfaced as `isError` tool results that the model just retried. This fires `session.agent.abort()` at each block branch. Unrelated to the restart fix.

## Behavior change

This **partially reverses issue #291**: channel-origin restarts now resume. The agent reopens the session, fires the post-restart turn, and resumes pending todos — including the in-channel greeting the user explicitly triggered.

The decision to include the in-channel greeting for channel origins was explicit: a user who told the agent to restart and expects a follow-up deserves the resume. TUI behavior is unchanged; cron and subagent origins still write no handoff.

## Testing

New test coverage added in this PR:

- `src/agent/restart-handoff/index.test.ts` — v2 schema, v1 forward-read, `accept` predicate (claim / restore / re-claim across TUI and channel kinds)
- `src/agent/restart-handoff-wiring.test.ts` — `buildRestartHandoffWiring` branch coverage: tui / channel / cron / subagent / system / in-memory / no-file / no-agentDir
- `src/agent/tools/restart.test.ts` — restart tool writes tui + channel handoffs and skips without `handoffOrigin`; `requestContainerRestart` field-gating
- `src/channels/router.test.ts` — `resumeRestartHandoff` reopens the exact session and wakes it; skips on stale mapping; is a no-op for tui handoffs
- `src/agent/plugin-tools.test.ts` — loop-guard abort fires exactly once on block (plugin and system paths); does not fire before the threshold

Gates: `bun run typecheck` clean, `bun run lint` 0 errors, `bun run format:check` clean, `bun run test` 6980 pass / 1 skip.

**One pre-existing test failure** in `src/update/index.test.ts` — `resolveSelfPackageJsonPath > points at this package manifest` — asserts the repo directory basename is `typeclaw`, which fails only when running from a worktree checked out under a differently-named directory. It passes on a standard `typeclaw` checkout and will pass in CI. It is not a regression from this change.

## Risk / graceful degradation

If `resumeRestartHandoff` fails at any step (stale mapping, unconfigured adapter, reopen error) it logs and returns without throwing, so `typeclaw run` startup is unaffected. The todo entry remains on disk and resumes on the next real inbound from that channel. There is no new failure mode that leaves the agent in a broken state.

The `accept` predicate + restore-on-decline logic means a TUI reconnect racing a channel boot cannot permanently consume the other kind's handoff — the file is atomically restored if declined.

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -46,6 +46,7 @@ import {
   zodToToolParameters,
 } from './plugin-tools'
 import { createReloadTool } from './reload-tool'
+import type { RestartHandoffOrigin } from './restart-handoff'
 import { loadSelf } from './self'
 import { SESSION_META_CUSTOM_TYPE, sessionMetaPayload } from './session-meta'
 import { renderSessionOrigin, type SessionOrigin, type SessionRoleContext } from './session-origin'
@@ -446,22 +447,39 @@ export async function createSessionWithDispose(options: CreateSessionOptions = {
 
 // Decides whether the restart tool should write the cross-restart handoff
 // file (`<agentDir>/.typeclaw/restart-pending.json`) and supplies the agentDir
-// + session file path it needs to do so. Returns an empty object — meaning
-// "no handoff" — for any session whose origin is not TUI, so a channel-
-// originated or cron-originated `restart` call cannot accidentally produce an
-// "I'm back" greeting in the next container's first TUI session. See
-// issue #291's scoping concerns. Also returns empty when the session is not
-// persisted to disk (in-memory sessions have no file the next container could
-// reopen).
+// + session file path + origin metadata it needs to do so. Returns an empty
+// object — meaning "no handoff" — for cron/subagent/system origins (no
+// attended session the next boot could resume) and for in-memory sessions
+// (no file to reopen).
+//
+// TUI and channel origins both resume: a TUI restart reattaches to the
+// reconnecting client (websocket open handler), a channel restart reopens the
+// originating chat session on the channel router's boot path. The `origin`
+// discriminator in the handoff is what routes the next boot to the correct
+// subsystem.
 export function buildRestartHandoffWiring(
   options: { origin?: SessionOrigin; plugins?: { agentDir: string } },
   sessionManager: SessionManager,
-): { agentDir?: string; originatingSessionFile?: string } {
-  if (options.origin?.kind !== 'tui') return {}
+): { agentDir?: string; originatingSessionFile?: string; handoffOrigin?: RestartHandoffOrigin } {
+  const origin = options.origin
+  if (origin === undefined) return {}
+  const handoffOrigin = restartHandoffOriginFor(origin)
+  if (handoffOrigin === null) return {}
   const agentDir = options.plugins?.agentDir
   const sessionFile = sessionManager.getSessionFile()
   if (agentDir === undefined || sessionFile === undefined) return {}
-  return { agentDir, originatingSessionFile: sessionFile }
+  return { agentDir, originatingSessionFile: sessionFile, handoffOrigin }
+}
+
+function restartHandoffOriginFor(origin: SessionOrigin): RestartHandoffOrigin | null {
+  if (origin.kind === 'tui') return { kind: 'tui' }
+  if (origin.kind === 'channel') {
+    return {
+      kind: 'channel',
+      key: { adapter: origin.adapter, workspace: origin.workspace, chat: origin.chat, thread: origin.thread },
+    }
+  }
+  return null
 }
 
 // Subscribes the given session to the in-process broadcast that the `restart`
--- a/src/agent/restart-handoff-wiring.test.ts
+++ b/src/agent/restart-handoff-wiring.test.ts
@@ -1,0 +1,74 @@
+import { describe, expect, test } from 'bun:test'
+
+import type { SessionManager } from '@mariozechner/pi-coding-agent'
+
+import { buildRestartHandoffWiring } from './index'
+import type { SessionOrigin } from './session-origin'
+
+function fakeSessionManager(sessionFile: string | undefined): SessionManager {
+  return { getSessionFile: () => sessionFile } as unknown as SessionManager
+}
+
+const tuiOrigin: SessionOrigin = { kind: 'tui', sessionId: 'ses-1' }
+const channelOrigin: SessionOrigin = {
+  kind: 'channel',
+  adapter: 'discord-bot',
+  workspace: 'g1',
+  chat: 'c1',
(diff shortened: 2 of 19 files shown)


devxoul:
Thanks — all three addressed. Pushed `5fc40ad`, `ab1e777`, `082f974`.

**1. Mapping mutated before reopen confirmed (`router.ts`)** — `resumeRestartHandoff` no longer pre-mutates `channels/sessions.json`. `ensureLive` gained an optional `resumeTarget` that forces rehydration of the exact `(sessionId, sessionFile)` and bypasses stale-rollover **in-memory only**; the mapping is persisted solely through `ensureLive`'s existing success path. On reopen failure the durable record is left byte-for-byte unchanged (no repointed `sessionFile`, no refreshed `lastInboundAt`), so the next inbound still stale-rolls into a clean session. New test asserts the mapping is untouched after a throwing factory (`leaves the durable mapping untouched when reopen fails`).

**2. Fire-and-forget boot resume races inbounds (`run/index.ts`)** — now `await`ed. The boundary is deliberate: `resumeRestartHandoff` enqueues the synthetic wake (`pendingSystemReminders`) **before** it resolves and drains the LLM turn fire-and-forget, so awaiting blocks boot only on the reopen + wake-enqueue, not on the model turn. That orders the wake ahead of any real same-channel inbound the server accepts once boot continues. (The same-key `creating`/`liveSessions` dedupe means a racing inbound for that key awaits the same reopen rather than starting a competing one.)

**3. consume deletes-then-restores race (`restart-handoff/index.ts`)** — switched to peek-then-conditional-delete. The file is read and validated (parse, TTL, `accept`) first and only deleted when actually claimed; an unclaimed handoff is never removed, so there's no deletion window for the rightful consumer. The claiming delete is non-forced so two same-kind consumers still honor it exactly once (the loser's unlink throws ENOENT → returns null). Updated the test accordingly (`...LEAVES the file untouched`).

Gates: typecheck/lint/format clean, full suite 6982 pass / 1 skip (the lone failure is the pre-existing worktree-basename test, green on a normal checkout + CI).

devxoul:
@typeey that's blocking?

devxoul:
Good catch — fixed in `f9ddf7c`. You're right that awaiting the resume only ordered the wake ahead of inbounds arriving *after* the await; an inbound landing in the window between `channelManager.start()` and the resume was unguarded.

Closed the race by **reserving the channel key before adapters start**, two-phase:

1. **Before `start()`** — `reserveRestartHandoff(handoff)` claims the handoff and seeds a per-key entry in the router's existing `creating` dedupe map. It touches no resolvers/outbound (none registered yet) — it just installs the gate.
2. **`start()`** — adapters connect and register outbound callbacks.
3. **After `start()`** — `reservation.resume()` reopens the exact session via `ensureLive(resumeTarget)` and enqueues the synthetic wake.

Because the key is reserved before any adapter can receive, an inbound that races in **coalesces onto the reservation** (its `ensureLive` awaits the seeded `creating` promise) instead of stale-rolling the mapping or creating a rival session. `route()` flips `reservation.sawInbound` when it coalesces, and `resume()` then **skips the synthetic wake** — the real inbound is the wake, so no duplicate/spurious "I'm back" turn. Deadlock avoided by having `resume()` drop its own `creating` seed before calling `ensureLive` (otherwise it'd await the gate it's about to resolve).

New tests in `router.test.ts`:
- a racing inbound + reserved resume produce exactly ONE session (coalesced, no rival create);
- the synthetic wake is skipped when an inbound coalesced during boot;
- the wake still fires when no inbound raced.

`src/run/index.ts` now reserves before `start()` and resumes after. Gates green (typecheck/lint/format; channel/run/restart suites all pass in isolation).

PR #599, opened 2026-06-03
channels: resume conversation on /restart command

devxoul:
## Background

Stacked on #598 (`fix/channel-restart-todo-continuation`). Merge after that PR lands.

After #598, a restart fired via the **restart tool** (what the agent calls on its own behalf) resumes the originating channel conversation on the next boot — #598 added the v2 handoff shape and `ChannelRouter.resumeRestartHandoff`. The `/restart` **command** (an operator types `/restart` in-channel or uses a native slash command) still bounced the container with no resume handoff. That conversation's pending todos would resume on the next inbound, but there was no proactive "I'm back."

This is the one gap identified in a TUI-primacy audit of the restart/resume surface. Todo continuation, the restart tool, and cron-restart were already channel-aware; the operator-facing command was the sole exception. Closes part of #291 (channels as equal-priority runtime to TUI).

### Commit 1 — `commands: add wantsLiveSession to resolve a session without requiring one`

`requiresLiveSession` aborts a command when no live session exists. A command that must execute regardless (e.g. `/restart` bounces the container even from a cold channel) but can do **more** when a session is present had no way to express that distinction. The new `wantsLiveSession` flag resolves an existing live session into the handler context when present and is silently ignored when absent. It is also ignored when `requiresLiveSession` is true. The flag is surfaced on `CommandInfo` so both dispatch paths can read it.

### Commit 2 — `channels: resume the originating conversation on a /restart command`

`/restart` now declares `wantsLiveSession`. Both dispatch paths (text-prefix `/restart` and native-slash) attempt a best-effort session resolution for the invoking channel: unambiguous match only, never fails. When a session is live, the handler passes its identity — `sessionId`, transcript file, and a `{kind:'channel', key}` `handoffOrigin` — to a widened `onRestart(ctx?)` callback. `src/run/index.ts` threads `stream`, session id/file, and `handoffOrigin` into `requestContainerRestart`, which appends the `typeclaw.restart-self` entry via the container-restarting broadcast and writes a channel-origin v2 handoff. On next boot, #598's channel resume path reopens that exact conversation.

With **no live session** (cold channel, native slash with no matching session, or ambiguous multi-session state), `ctx` is `undefined` and the container bounces as before — no handoff, todos resume on the next inbound.

## What this does NOT do

- Does not change the restart **tool** path or TUI restart behavior — both are untouched.
- Does not change `/restart` gating: `session.admin` (owner/trusted only) is unchanged.
- Cold-channel and ambiguous-session cases degrade gracefully to a plain bounce — same behavior as before this PR.
- Does not add new retry or backoff logic; the resume path is exactly #598's existing channel-handoff machinery.

## Review notes

- **Load-bearing change:** `onRestart(ctx?)` widening in `src/channels/router.ts` and the corresponding call site in `src/run/index.ts`. Verify that `ctx` is only forwarded when unambiguously resolved (the "never fails" invariant means a multi-session match returns `undefined`, not a guess).
- **`wantsLiveSession` vs `requiresLiveSession` interaction:** when both are set, `requiresLiveSession` wins — the command aborts before `wantsLiveSession` resolution runs. Covered by the registry test.
- **Environmental test note:** `resolveSelfPackageJsonPath > points at this package manifest` asserts the repo dir basename is `typeclaw` — it fails only when running from a differently-named worktree. Pre-existing, unrelated to this PR, passes on a standard checkout and in CI.
- Changed files: `src/commands/index.ts`, `src/commands/index.test.ts`, `src/channels/router.ts`, `src/channels/router.test.ts`, `src/channels/manager.ts`, `src/channels/commands.test.ts`, `src/run/index.ts`.
- Full suite: 6984 pass, 1 skip, 1 environmental fail (above).

### DIFF
--- a/src/channels/commands.test.ts
+++ b/src/channels/commands.test.ts
@@ -10,6 +10,7 @@ const info = (name: string, description: string): CommandInfo => ({
   description,
   permission: 'session.control',
   requiresLiveSession: true,
+  wantsLiveSession: false,
 })
 
 describe('formatChannelCommandHelp', () => {
--- a/src/channels/manager.ts
+++ b/src/channels/manager.ts
@@ -13,7 +13,13 @@ import { createKakaotalkAdapter, type KakaotalkAdapter } from './adapters/kakaot
 import { createSlackBotAdapter, type SlackBotAdapter } from './adapters/slack-bot'
 import { createTelegramBotAdapter, type TelegramBotAdapter } from './adapters/telegram-bot'
 import type { GithubTokenBridge } from './github-token-bridge'
-import { createChannelRouter, type ChannelRouter, type ClaimHandler, type CreateSessionForChannel } from './router'
+import {
+  createChannelRouter,
+  type ChannelRouter,
+  type ClaimHandler,
+  type CreateSessionForChannel,
+  type RestartCommandContext,
+} from './router'
 import {
   ADAPTER_IDS,
   type AdapterId,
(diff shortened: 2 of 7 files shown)


devxoul:
@typeey please approve

PR #600, opened 2026-06-03
sandbox: confine low-trust bash writes to the write/edit allowlist

devxoul:
## Background

The per-tool bash sandbox (`src/sandbox/`, role-derived via `resolveHiddenPaths`) was a **deny-list** that only *hid* private paths from low-trust roles: it bound the whole agent folder read-write, then masked `workspace/`, `memory/`, `sessions/`, `.env`, `secrets.json`. Hiding reads is only half the boundary. Writes were never confined — a low-trust role could `echo … > /agent/anything`, `tee`, or otherwise redirect to any path under `/agent` and sidestep the `nonWorkspaceWrite` guard that bounds the `write`/`edit` tools to `workspace/`, `public/`, `mounts/`, `packages/`, the skill dirs, and the root-file allowlist.

So the file tools enforced a write boundary that `bash` did not. This PR makes `bash` enforce the same boundary at the kernel level (`EROFS`), not by parsing commands.

## Summary

Invert the model for the roles that already get masks (guest/member and any unresolved/unmatched origin): instead of "bind everything rw, then hide some reads," bind the agent folder **read-only**, keep the read-hiding masks, then re-expose **only** the write/edit allowlist read-write. Anything else under `/agent` (`.git/`, `node_modules/`, the agent-dir root itself) is read-only.

Key decisions:

- **Why scope the jail to low-trust roles, not all?** `trusted`/`owner` legitimately run `git commit` (writes `.git/`), `bun install` (writes `node_modules/`, lockfiles), and other whole-project tooling that lives *outside* the user-facing write/edit allowlist. Confining them would break real workflows, so they keep today's unsandboxed behavior. The split falls out naturally: a role that gets no masks gets no jail.
- **Why bind read-only then re-expose, rather than bind only the allowlist?** Tools still need to *read* the whole tree (minus masked paths). RO-root preserves reads; the rw binds carve out writes. Mount order is load-bearing — bwrap applies ops in order and last-op-wins, so the builder emits `--ro-bind /agent` → masks → `--bind` allowlist last.
- **Why reject symlinked zone roots?** An rw bind follows symlinks; a planted `workspace -> /etc` at a zone root would grant outside write access. `resolveWritableZones` binds only paths that exist and are not symlinks. (Symlinks *inside* a real zone are already safe — the kernel resolves them to the read-only parent mount.)
- **Why can't bash create new root files?** A missing `cron.json`/`USER.md` is not pre-created (that would make a read-only command mutate the project as a side effect). bash can *edit* an existing root file; *creating* one goes through the `write`/`edit` tool. Confirmed as the intended limitation.

Split into three atomic commits: (1) additive `writable` policy field in the command builder, (2) the allowlist resolver, (3) the `applyBashSandbox` behavior flip + docs.

## What this does NOT do

- Does **not** confine `trusted`/`owner` bash — they remain unsandboxed by design. A future "dev writable set" (`.git/`, `node_modules/`, lockfiles) would be needed to jail them, and is explicitly out of scope.
- Does **not** change the read-hiding masks or the `privateSurfaceRead` guard.
- Does **not** close the confused-deputy egress concern (a higher-role turn copying private data into `public/`) — that remains an authn-shaped problem solved by per-trust-domain agent separation, as documented in `sandbox.mdx`.
- Does **not** add command-syntax filtering; enforcement is purely kernel-level mount permissions.

## Review notes

- Load-bearing change: `applyBashSandbox` in `src/agent/plugin-tools.ts` — the `--bind` → `--ro-bind` flip plus the `writable` overlay. The early-return on empty masks is what keeps trusted+ unsandboxed; do not "fix" it.
- Invariant to verify: in `src/sandbox/build.ts`, writable binds must render **after** masks (and masks after the ro-root) so last-op-wins makes the allowlist the only rw surface. Covered by `build.test.ts` ordering tests.
- The writable set in `src/sandbox/writable-zones.ts` **must** stay in sync with the `nonWorkspaceWrite` guard allowlist — divergence is a security hole in either direction. Flagged in a code comment.
- **Not verifiable on a macOS dev host** (no `bwrap`): the actual `EROFS`-on-write behavior only exercises inside the Linux container. Unit tests cover argv ordering and zone resolution (existing/symlink/type-mismatch cases); the end-to-end write-denial is asserted by construction. A real-bwrap fixture is in the design notes if a reviewer wants to run it in-container.


devxoul:
Both blockers addressed in 4f1e385.

**Blocker 1 (workspace re-exposed for guest):** You're right — the RW overlay rendered after the masks and last-op-wins re-exposed the real masked dir. Added `subtractMasked` (`src/sandbox/writable-zones.ts`), applied in `applyBashSandbox` before binding: it drops any writable zone that is, or is nested under, a masked path. A guest's masked `workspace/` is no longer re-bound, so its writable set is now just `public/`, `mounts/`. Member (no dir masks) keeps `workspace/`, `public/`, `mounts/`. Covered by new `subtractMasked` tests, including a `workspace-public` vs masked `workspace` prefix-collision case.

**Blocker 2 (.agents/skills too broad):** Dropped `.agents/skills` from the bash writable set entirely — also dropped `packages/` (executable plugin code) by the same logic. The set is now deliberately narrower than the `nonWorkspaceWrite` allowlist: only genuinely free-write scratch zones (`workspace/`, `public/`, `mounts/`) get a blanket RW bind. Skill authoring and package writes go through the guarded `write`/`edit` tool only, so bash is never more powerful than write/edit for those surfaces.

Docs (`sandbox.mdx`) updated with the per-role writable table and the narrowing/intersection rationale.

PR #601, opened 2026-06-03
github: post inline PR reviews via a first-class tool instead of skill-driven gh api

devxoul:
## Background

When the bot reviews a PR it is supposed to post an **inline review** — `POST /repos/:o/:r/pulls/:n/reviews` with a `comments[]` array anchored to `path:line`. Today that whole flow lives in skill prose (`typeclaw-channel-github` step 4): the model writes `/tmp/review.json` by hand and shells out to `gh api --input`. That path is fragile in three ways the skill itself documents but cannot enforce:

- A `line` anchor that isn't part of the diff makes GitHub **422 the entire review**, and the model has been observed silently falling back to a flat top-level comment that *claims* a review was posted.
- The approval-disabled policy (`channels.github.review.approve: false`) is enforced only by the model remembering to downgrade `APPROVE` → `COMMENT`.
- The "verify the review actually landed" step is a separate `gh api` call the model can skip.

The net failure mode: a review with real `file:line` findings collapses into a flat `## Review` issue comment with the anchors surviving only as prose. (Observed on a recent PR.)

## Summary

Move the review-submission mechanics out of skill prose and into a typed, code-owned seam, mirroring the existing **review-thread resolver** architecture (tool → router method → github adapter resolver).

- **`post_github_review` tool** (github channel sessions only) takes `{ event, body, comments[] }`.
- **`router.submitReview`** dispatches to a per-adapter `ReviewSubmitter`, mirroring `resolveReviewThread` (same `unsupported`/transient-on-throw contract).
- **`createGithubReviewSubmitter`** owns the mechanics the model used to do by hand:
  - resolves the PR **head SHA** (`commit_id`),
  - fetches the PR files and builds the set of **valid diff anchors**,
  - **auto-demotes** out-of-diff findings into the top-level body (reported back as `reanchored`) instead of letting GitHub 422 the review,
  - **auto-downgrades** `APPROVE` → `COMMENT` when the operator approval policy is disabled,
  - POSTs the review and **verifies it landed** (re-reads the last review's `id`/`state`).

The skill's step 4 now just calls the tool; SHA resolution, anchor validation, the approval policy, and verify-after-write are no longer the model's responsibility.

Three atomic commits: (1) router seam + types, (2) github submitter impl + wiring, (3) the tool + agent wiring + skill rewrite.

## Key decisions

- **Auto-demote, not reject, on a bad anchor.** Rejecting reintroduces the exact silent-flatten risk if the model mishandles the error. Demotion guarantees the review still lands with every valid inline anchor; demoted findings go into the body with their `path:line` preserved and are surfaced to the model via `reanchored`.
- **Auto-downgrade APPROVE when approval is disabled.** Matches the prior skill intent (downgrade rather than error), keeps the agent from failing on a policy it can't control; the receipt reports `downgraded`.
- **Mirror the review-thread resolver, don't extend `OutboundMessage`.** A formal review isn't a chat message — the router's flood/dedup/turn-cap guards don't apply. The resolver registry is the established, trusted seam for adapter-specific GitHub API actions.

## What this does NOT do

- Does **not** change `OutboundMessage` or the outbound callback — `channel_reply`'s flat-comment path is untouched (still correct for non-review replies and the zero-actionable-findings branch).
- Does **not** move the **re-review / sticky-`CHANGES_REQUESTED` dismissal** logic into code — that stays skill-side (separate endpoint, lower frequency); a follow-up can fold it in.
- Does **not** paginate the PR **files** fetch beyond the first 100. On a PR with >100 changed files, anchors on later files fall through to demotion (safe degradation, never a 422). Worth a follow-up if it bites.

## Review notes

- Load-bearing: `partitionComments` + `addPatchAnchors` in `src/channels/adapters/github/review-submitter.ts` — the hunk parser tracks per-hunk left/right line counters (RIGHT for additions/context, LEFT for removals). The anchor key is `path\0side\0line`.
- `submitReview` clones `resolveReviewThread`'s register/unregister/dispatch verbatim; the github adapter registers/unregisters it in all three lifecycle spots (start, start-rollback, stop).
- `post_github_review` is gated on `origin.adapter === 'github'` in `buildChannelTools` and denies on any other adapter.
- **Not exercisable end-to-end in unit tests** (no live GitHub): the submitter is covered with a hand-rolled `fetchImpl` fake (in-diff passthrough, demotion, approve-downgrade, 403→permission-denied, verify-after-write, bad-target parse), the tool with origin-gate + mapping + receipt tests, and the router seam with register/unregister/dispatch/throw tests.

Checks: `bun run typecheck`, `bun run lint`, `bun run format:check`, `bun run test` (7017 pass, 0 fail) all green.


devxoul:
Closing in favor of #602.

This PR added a `post_github_review` tool + router submitter seam. On reflection the premise is wrong:

- The flat-comment failure was **not** a missing-capability bug — the bot already posts formal reviews fine (it posted a formal `APPROVED` review on the same PR via `POST /pulls/N/reviews`). The skill already instructs anchor-validation, demote-to-body, verify, and "never flatten." The model just didn't follow it on one turn.
- A new tool (or a review payload on `channel_send`) adds permanent surface and a second GitHub-write path but **does not enforce** anything — the model could still choose the flat path. The only thing code would buy is enforcement (a guard), and with n=1 that's premature.

So the right fix is the smaller one: front-load the existing "never flatten" rule in the skill (#602). If it recurs, we add a `tool.before` guard then — not a capability layer now.

PR #603, opened 2026-06-03
docs: document the todo continuation subsystem in internals

devxoul:
## What

Adds an Internals page for the **todo continuation** subsystem (`src/agent/todo/` + `src/agent/tools/todo/`), which had **zero** coverage anywhere under `docs/` despite being one of the more invariant-heavy, security-relevant subsystems in the runtime.

## Why

A grep of `docs/` for `todo` / `continuation` returns nothing, yet the source ships:

- **Three agent tools** (`todo_write` / `todo_read` / `todo_clear`) that fail closed on no-scope.
- **Durable scope resolution** (`resolveTodoScope`) keyed on origin identity so a list survives TUI reconnects, cron re-fires, and channel stale-rollover.
- **A fail-closed auto-continuation engine** with persisted episode budgets (max auto-turns / cumulative tokens / wall-clock / stagnation), restart-kick + user-abort suppressors, and a shrink-only progress test that closes the fake-progress loophole.

Per `AGENTS.md`, this is exactly what the Internals section exists for — load-bearing invariants, file paths, and "order matters" rules. The omission left a subsystem with crash-loop ceilings and fail-closed guards undocumented while smaller subsystems already have pages.

## Changes

- `docs/content/docs/internals/todo-continuation.mdx` — new page: the three tools, the scope table, storage/atomicity, episodes + budgets, the stagnation heuristic vs real-progress test, the six-step skip ladder, the two suppressors, fail-closed state parsing, and the wiring seam.
- Registered in `docs/content/docs/internals/meta.json` (after `engagement`) and `internals/index.mdx`.
- Added a row to the **Architecture** table in `AGENTS.md`.

Docs-only; no runtime code touched. Page content mirrors the load-bearing comments in `continuation-policy.ts`, `scope.ts`, `store.ts`, `continuation-state.ts`, and `continuation-wiring.ts`.

## Verification

Run in `docs/`:

- `bun run lint` — clean (pre-existing `page.tsx` warnings only, untouched).
- `bunx oxfmt --check .` — all files correctly formatted.
- `bun run build` — TypeScript clean, 47 static pages generated (the new page renders).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -149,17 +149,18 @@ Domain logic lives in `src/<domain>/`. Examples: `src/init/`, `src/config/`, `sr
 
 The architecture reference lives in the published docs under [Internals](https://typeclaw.dev/docs/internals). It's terse, file-path-heavy, and aimed at the same audience as this file — someone about to edit `src/`.
 
-| Subsystem                                                                                     | Where                                                                                |
-| --------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------ |
-| Skills loading sources, naming, lazy semantics                                                | [/docs/internals/skills](https://typeclaw.dev/docs/internals/skills)                 |
-| `.env` / `secrets.json`, the `Secret` shape, bridge idempotency, KakaoTalk encryption-at-rest | [/docs/internals/secrets](https://typeclaw.dev/docs/internals/secrets)               |
-| Roles, match-rule DSL, cron/subagent provenance, security guard tiers                         | [/docs/internals/permissions](https://typeclaw.dev/docs/internals/permissions)       |
-| `src/hostd/`, three trust channels, control protocol, portbroker                              | [/docs/internals/hostd](https://typeclaw.dev/docs/internals/hostd)                   |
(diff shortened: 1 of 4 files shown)


devxoul:
Both factual issues fixed in 33288de:

1. **Stagnation gate** — Corrected to describe the live behavior: `decideContinuation` keys stagnation off `lastIncompleteHash === hash`, not `hasRealProgress`. The page now states a reword/split changes the hash and *resets* the counter (no item completed), and explicitly notes `hasRealProgress` (shrink-only) exists with unit coverage but is **not wired into the decision** — a latent helper, not the enforced invariant. Updated the matching 'Rules of thumb' bullet too.

2. **scope.key wording** — Softened: `scope.key` is now described as a *traversal-safe relative key* that deliberately contains `/` (`channel/...`, `cron/...`) stored as nested paths under `todo/`, while only the encoded **components** are single-path-segment-safe.

Docs build + root `format:check`/lint/typecheck all green.

devxoul:
@typeey approve please

PR #605, opened 2026-06-04
github skill: triage inbound intent before acting on PRs

devxoul:
## Background

On a real PR (typeclaw/typeclaw#603), the bot posted a "looks good — thanks for tightening both spots" comment but the PR stayed **"Awaiting requested review / 1 approving review required"**. It never approved.

Root cause traced through two sessions on the host agent:

1. **The original "request changes" was never a formal review.** The reviewer subagent returned two `general`-location findings (doc line refs, not diff-anchorable), so the bot took the zero-actionable→comment branch and posted a flat conversation comment via `channel_reply`. GitHub recorded no `CHANGES_REQUESTED` review — the PR was *commented on*, not *blocked*.

2. **The re-review never entered the PR review flow.** When the author commented "Both factual issues fixed in <sha>", that inbound matched neither trigger (no synthetic `review_requested`, no plain-language "review this"). The bot's first reasoning jumped to *"address the review thread → resolve it"*, tried `resolve_review_thread: true`, got denied (no inline thread existed — see #1), and fell back to a plain `channel_reply`. No reviewer spawn, no `POST /pulls/603/reviews`, no `APPROVE`.

Net: an `issue_comment` carries no review state, so the verdict the bot *meant* to deliver never landed. This is the recurrence #601/#602 anticipated — the "always land a formal review" rule was prompt-mediated and the routing was action-shaped, not intent-shaped.

## Summary

Add a mandatory **Inbound triage** gate at the top of `## What to do, by inbound type` that forces classify-before-act. The four ordered questions resolve to a single path; the decisive one is **#1: "do I have unfinished review business on this PR?"** — answered against the formal review state (`{CHANGES_REQUESTED, APPROVED}` from the existing step-1 re-review query), *before* any prose pattern-match.

Key reframings:
- An author comment like "fixed both issues" / "pushed a fix" on a PR you previously blocked is a **re-review trigger**, not a thread-resolve trigger.
- `resolve_review_thread` is gated on an actual inline `thread` being present on the origin; if there's no thread, that branch is explicitly skipped rather than attempted-then-fallen-back.
- A verdict owed to a PR is only discharged by a formal review via `POST /pulls/<N>/reviews` — never `channel_reply` / `issue_comment`.

**Why a skill triage and not a runtime guard:** this fixes the *routing* hole that caused #603 directly and cheaply. The deeper enforcement (reject a flat reply that's closing out an owed verdict, at the GitHub review-posting boundary) is still the durable backstop #601 flagged; that's a separate, runtime-layer change and out of scope here.

**Why GitHub-only:** the other channel skills (kakaotalk, telegram-bot) are rendering skills with a single outbound action (post text) — there is no ambiguous intent to triage and no silent wrong-path failure mode. Triage belongs only where multiple action paths look alike in prose and the wrong one fails silently, which today is GitHub alone.

## What this does NOT do

- Does not add a runtime enforcement guard at the review-posting boundary (deferred; see #601).
- Does not touch any runtime code — docs/skill prose only.
- Does not change the PR review flow mechanics, the reviewer subagent, or the verdict→event mapping.
- Does not add triage to non-GitHub channel skills.

## Review notes

Load-bearing change is the new "Inbound triage" section and the one-line edit to the action-table intro pointing at it. Invariant to verify: **on any `pr:N` inbound, the prior-verdict check (#1) runs before the prose-based routing in the table** — that's the ordering that prevents a re-review from being misrouted to thread-resolve or a flat reply.

Docs-only; no runtime behavior changes. `lint` / `format:check` / `typecheck` clean. The one `test` failure (`typeclaw.schema.json` drift guard) is pre-existing on `main` and unrelated — this branch's only diff is a markdown file.


devxoul:
Addressed both in 751e1c0:

- **Concern (flat-comment blockers):** You're right — checking only `{CHANGES_REQUESTED, APPROVED}` misses the #603 case where the prior "request changes" was itself a flat comment, so no formal state ever existed. Triage #1 now checks **two** signals: formal review state **and** a scan of my own prior PR/issue comments for an unretracted blocker. Either one routes the inbound to the PR review flow, and the text now states a flat-comment blocker is as binding as a formal `CHANGES_REQUESTED` for routing.

- **Nit (wording too broad):** Narrowed "pending or prior verdict" → "unresolved blocking obligation" (a live `CHANGES_REQUESTED`, or an unretracted flat-comment blocker). A stale `APPROVED` or a past non-blocking comment explicitly imposes no closeout duty now.

The deeper runtime-boundary enforcement (reject a flat reply that's closing out an owed block) is still tracked separately per #601 — out of scope for this docs-only change.

devxoul:
@typeey approve please

devxoul:
Fixed the consistency gap in e7d72ca — you're right that recognizing flat-comment blockers in triage was pointless if the flow then dropped them.

The PR review flow now carries the triage result through end to end:

- **Step 1 (re-review detection)** checks **both** signals — formal review state **and** a scan of my own prior comments for an unretracted blocker — mirroring triage #1. Either one marks the request a re-review.
- **Step 2 (spawn payload)** triggers on either form and passes **which form** the prior block took into the reviewer. The flat-comment case is called out explicitly as must-pass-through, since the reviewer can't recover it from review state.
- **Step 4 re-review + the zero-actionable `comment` exception** now speak of an “unresolved blocking obligation” rather than formal `CHANGES_REQUESTED` only. Clearing mechanics are split by form: a flat-comment block has no sticky state to dismiss, so it's discharged as a **formal** review (`APPROVE`/`REQUEST_CHANGES`) — never another flat comment. The approval-disabled branch likewise distinguishes “nothing to dismiss” (flat) from “dismiss the prior review” (formal).

devxoul:
@typeey approve in github please

PR #606, opened 2026-06-04
github: register and deregister repo webhooks concurrently

devxoul:
## Background

On container start and stop the GitHub channel adapter iterated over registered repos one at a time — awaiting each repo's installation-token resolution followed by list/create/update/delete round-trips before touching the next repo. For an agent watching N repos this made start and stop latency O(N × network): each repo's token fetch plus up to three API calls had to finish before the next repo started.

Root cause: the original implementation used `for…of` loops with `await` at each iteration, giving purely sequential execution where no ordering dependency existed.

## Summary

Replace three serial loops in `src/channels/adapters/github/webhook-register.ts` with `Promise.all` over `.map`:

1. `registerGithubWebhooks` — the outer per-repo loop.
2. `deregisterGithubWebhooks` — the outer per-hook loop.
3. `registerOne` — the inner stale-orphan prune delete loop (success count collected with `.filter(Boolean).length`).

Each task is an `async` function that catches its own failures into `{ action: 'failed', error }` entries and never throws, so `Promise.all` never short-circuits. Order is preserved (spec-guaranteed by `Promise.all` over an array), keeping result shape identical.

Key decisions:

- **Why `Promise.all` and not `Promise.allSettled`?** Tasks are already written to resolve (never reject) on per-item failure — the existing `try/catch` pattern converts errors to `failed` entries. `Promise.all` is therefore sufficient and less noisy than `allSettled` wrapping.
- **Why not batch or rate-limit?** The GitHub Apps installation-token endpoint is per-installation, so N repos across K owners issue K distinct token fetches (often 1 for a single-org agent). Parallelism is bounded by the repo list configured by the user, not an open fan-out. A concrete throttle adds complexity with no benefit for the observed use case.

## Preview

No visual output. The change is behavioral: start/stop wall-clock latency for a multi-repo agent now scales with the slowest single repo rather than the sum of all repos.

## What this does NOT do

- Does **not** change the public API or return types — `WebhookRegistrationResult`, `WebhookDeregistrationResult`, and the per-item result shapes are unchanged.
- Does **not** change error-isolation semantics — a failing repo still produces a `failed` entry and does not block other repos.
- Does **not** add rate-limiting or concurrency caps.
- Does **not** touch the token-cache implementation, the `tryDeleteHook` helper, or any other file.

## Review notes

Load-bearing invariants to verify:

1. **Failure isolation preserved.** The `try/catch` in each `.map` callback returns (not throws) on error, so one repo's failure cannot cause `Promise.all` to reject.
2. **Result order preserved.** `Promise.all` resolves to an array in input order, matching the previous `push`-based accumulator. Both ordered `toEqual` and order-independent `.find` assertions in the test suite continue to hold.
3. **Stale-prune count.** `pruned.filter(Boolean).length` replaces the `stalePruned++` counter — semantically equivalent since `tryDeleteHook` returns a boolean.

Verification:

- `bun run typecheck` — clean.
- `bun run lint` — 0 errors (66 pre-existing warnings, none in the changed file).
- `bun run format:check` — clean.
- 273 tests pass / 0 fail, including all 22 cases in `webhook-register.test.ts` (failure isolation, multi-owner tokens, stale pruning, idempotent re-runs). Two unrelated failures in `src/update/index.test.ts` (`resolveSelfPackageJsonPath`) are a pre-existing environmental artifact — the assertion hardcodes a directory name that does not match the working directory used to run this branch — not a regression from this change.

### DIFF
--- a/src/channels/adapters/github/webhook-register.test.ts
+++ b/src/channels/adapters/github/webhook-register.test.ts
@@ -150,6 +150,23 @@ describe('registerGithubWebhooks', () => {
     expect(good?.action).toBe('created')
   })
 
+  test('deduplicates repeated repo slugs so concurrent fan-out cannot create duplicate hooks', async () => {
+    const { fetch: fetchImpl, calls } = makeFetch(({ url, init }) => {
+      if (url.includes('/repos/acme/widgets/hooks') && (init?.method ?? 'GET') === 'GET') {
+        return { status: 200, body: [] }
+      }
+      if (url.endsWith('/repos/acme/widgets/hooks') && init?.method === 'POST') {
+        return { status: 201, body: { id: 42 } }
+      }
+      return { status: 500 }
+    })
+
+    const result = await registerGithubWebhooks(baseOpts({ fetchImpl, repos: ['acme/widgets', 'acme/widgets'] }))
+
+    expect(result.repos).toEqual([{ repo: 'acme/widgets', action: 'created', hookId: 42 }])
+    expect(calls.filter((c) => c.init?.method === 'POST').length).toBe(1)
+  })
+
   test('resolves a separate token per repo so multi-owner repos use the right installation', async () => {
     const tokenByRepo: Record<string, string> = { 'acme/api': 'tok-acme', 'personal/blog': 'tok-personal' }
     const bearerByRepo: Record<string, string | null> = {}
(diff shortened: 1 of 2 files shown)


devxoul:
Good catch — fixed in 3ff12c7. `registerGithubWebhooks` now collapses `options.repos` to distinct slugs (`[...new Set(...)]`) before the `Promise.all` fan-out, so a repeated slug can no longer race two `registerOne()` calls into duplicate hook creation. Added a regression test (`deduplicates repeated repo slugs so concurrent fan-out cannot create duplicate hooks`) asserting a single POST/create per repo when the same slug appears twice. Deduping at registration also collapses `managedHooks`, so the deregister path inherits the fix.

devxoul:
@typeey approve please

PR #608, opened 2026-06-04
github: skip auto-review on draft PRs until explicitly requested

devxoul:
## Background

When an operator sets `channels.github.review.on: 'opened'`, TypeClaw auto-reviews every PR the moment it opens. That sweep includes **draft** PRs. A draft is an explicit "work in progress, not ready" signal from the author — auto-reviewing it is premature (the diff is still churning) and noisy (the bot posts a formal review nobody asked for, then has to re-review once the work lands).

There was no draft awareness anywhere in the GitHub adapter: `pull_request.draft` was never read. The `'opened'` path synthesized a review-trigger inbound (`isBotMention: true`) for any opened PR, draft or not.

## Summary

`classifyOpenedReviewTrigger` now returns `null` when the PR payload's `draft` flag is `true`. That drops the PR to the existing **awareness-only** context path — the same place a non-`'opened'` `reviewOn` already lands an opened PR — instead of waking a review session.

**Why guard only the `opened` path, not `review_requested`:** "auto-review" is precisely the automatic `opened` sweep. An explicit `review_requested` event is a human (or team) deliberately asking for a review, which the request says should still work — "until explicitly requested." So `classifyReviewRequest` is left untouched: requesting the bot on a draft still triggers a review. This keeps the two paths cleanly split — automatic vs. explicit — and means the only behavior change is "the automatic sweep now respects draft status."

A small `readBoolean` helper joins the existing `readString`/`readNumber` payload accessors.

## What this does NOT do

- Does not touch the explicit `review_requested` / `review_request_removed` path — explicit requests fire on drafts as before.
- Does not add a `ready_for_review` trigger. Marking a draft ready does not auto-review it today (that action isn't in the default event allowlist); this PR keeps that behavior. Re-opening for review still goes through an explicit `review_requested`.
- Does not add a new config knob — draft-skip is unconditional under `'opened'`, matching the "work in progress" semantics of a draft.

## Review notes

- Load-bearing change: the single `if (readBoolean(pr, 'draft') === true) return null` line in `classifyOpenedReviewTrigger` (`src/channels/adapters/github/inbound.ts`). Returning `null` is what hands the PR to the awareness-only fallthrough rather than the review trigger.
- Invariant to verify: explicit `review_requested` on a draft still synthesizes a review trigger. Covered by `still triggers on an explicit review_requested even when the PR is a draft`.
- New tests assert: draft `opened` → awareness-only (`isBotMention === false`, no "Please review" text); `draft: false` → still reviews; explicit request on a draft → still reviews. The draft-skip test fails if the guard is removed (verified).

### DIFF
--- a/src/channels/adapters/github/inbound.test.ts
+++ b/src/channels/adapters/github/inbound.test.ts
@@ -356,6 +356,34 @@ describe('classifyGithubInbound', () => {
         expect(msg?.text).toContain('requested your review on PR #7')
       })
 
+      it('skips the auto-review on a draft PR, landing it as awareness-only context', () => {
+        const msg = classifyGithubInbound('pull_request', openedPayload({ draft: true }), 'typeclaw-bot', {
+          reviewOn: 'opened',
+        })
+        expect(msg?.chat).toBe('pr:7')
+        expect(msg?.isBotMention).toBe(false)
+        expect(msg?.text).not.toContain('Please review the changes line-by-line')
+      })
+
+      it('still auto-reviews when draft is explicitly false', () => {
+        const msg = classifyGithubInbound('pull_request', openedPayload({ draft: false }), 'typeclaw-bot', {
+          reviewOn: 'opened',
+        })
+        expect(msg?.isBotMention).toBe(true)
+        expect(msg?.text).toContain('Please review the changes line-by-line')
+      })
+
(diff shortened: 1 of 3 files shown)


devxoul:
@typeey try approve again

PR #615, opened 2026-06-04
portbroker: stop the auth-failure reconnect loop on restart

devxoul:
## Background

After a restart, hostd logs flood with an endless loop:

```
portbroker-log {"kind":"portbroker-log","event":{"kind":"auth-failed","reason":"token mismatch"}}
```

repeating every ~1-2 seconds, indefinitely.

The broker token is minted fresh on every `typeclaw start` (`randomBytes(32)` in `src/container/start.ts`) and baked into the container at `docker run` time as `TYPECLAW_HOSTD_BROKER_TOKEN`. It is immutable for the lifetime of a broker instance — reconnecting can never make a wrong token match. When a host-side broker fails the broker-hello handshake, the client did `ws.close()` → `onClose` → `scheduleReconnect()` → reconnect → re-send the same wrong token → nack again. The TCP/WS connect itself succeeds (only the app-level hello is rejected), so `reconnectAttempt` resets to 0 each connect and the backoff never grows — ~1-2 s loop forever.

The tokens diverge after an ungraceful hostd death (crash / kill -9 / reboot): the persisted `registrations/<name>.json` survives with `T_old`, boot-time restore revives a `T_old` broker (restore errs toward reviving on a flaky `docker ps` probe), and `typeclaw restart` brings up a container expecting `T_new`. The existing serialized-swap fix in `portbroker-manager.ts` only collapses the one-shot race when a fresh register supersedes the stale broker; when nothing supersedes it, it loops.

## Summary

Treat an auth broker-hello-nack (`invalid token` / `expected broker-hello first`) as terminal rather than reconnectable.

- `src/portbroker/hostd-client.ts` — auth nack calls a `fatalStop` path (set stopped, clear reconnect timer, tear down forwarders, close WS) instead of reconnecting, and reports the reason via a new `onFatalAuthFailure` option. Non-auth closes still reconnect via the existing backoff. `stop()` and `fatalStop()` share an extracted `teardown()`.
- `src/hostd/portbroker-manager.ts` — bubbles `onFatalAuthFailure` upward and drops the dead broker from its map (identity-guarded so it can't evict a newer broker).
- `src/hostd/daemon.ts` — on fatal auth, GCs the matching registration (stop broker + clear maps + unlink persisted file) so it is NOT revived on the next boot. The removal is token-guarded: if a fresh register already overwrote the stored token with `T_new`, a late `T_old` callback is a no-op and the live registration is preserved.

The serialized swap-on-re-register contract and the restore-on-unknown probe behavior are both left intact.

## Preview

N/A — no UI changes.

## What this does NOT do

- Does **not** change behavior for non-auth disconnections — those still reconnect with the existing backoff.
- Does **not** change the serialized swap-on-re-register contract in `portbroker-manager.ts`.
- Does **not** change the restore-on-unknown probe behavior in `daemon.ts`.
- Does **not** affect any broker whose token matches the current container token.

## Review notes

- **Load-bearing invariant:** the token guard in `daemon.ts` (`storedToken === fatalToken`) must match the persisted token, not the manager's in-memory broker. If a fresh `T_new` register already overwrote the file, `storedToken` is `T_new` ≠ `T_old` and the callback is a no-op — the live registration is preserved. Verify this guard is checked against `readRegistration()` output, not against the manager map.
- **`fatalStop` vs `stop`:** the key difference is that `fatalStop` fires `onFatalAuthFailure` and bypasses the reconnect path entirely. A `stop()` call from outside (e.g., manager teardown) still goes through the normal `teardown()` path.
- **Tests:**
  - `hostd-client.test.ts` — auth nack stops permanently (no reconnect, fires `onFatalAuthFailure`); non-auth WS close still reconnects.
  - `daemon.test.ts` — fatal auth GCs the matching registration (broker stopped + file removed + evicted from list); a stale `T_old` fatal callback does NOT delete a freshly re-registered `T_new`.
- Checks: `bun run typecheck`, `bun run lint`, `bun run format`, `bun run test` all pass. One pre-existing unrelated failure (`resolveSelfPackageJsonPath > points at this package manifest`) triggers only in a worktree whose directory isn't named `typeclaw`; it passes in CI and on a normal checkout.

### DIFF
--- a/src/hostd/daemon.test.ts
+++ b/src/hostd/daemon.test.ts
@@ -589,6 +589,77 @@ describe('startDaemon', () => {
     expect(existsSync(registrationFilePath('coder'))).toBe(false)
   })
 
+  test('fatal auth failure GCs the matching registration (broker + file + list)', async () => {
+    const startCalls: PortbrokerStartInput[] = []
+    const stopCalls: Array<{ name: string; reason: string }> = []
+    const portbroker: PortbrokerCallbacks = {
+      start: async (input) => {
+        startCalls.push(input)
+      },
+      stop: async (name, reason) => {
+        stopCalls.push({ name, reason })
+      },
+      forwardedPorts: () => [],
+    }
+    daemon = await startDaemon({ exec: fakeExec(new Set(['coder'])), gcIntervalMs: 1_000_000, portbroker })
+    await send({
+      kind: 'register',
+      containerName: 'coder',
+      cwd: '/agent/coder',
+      wsHostPort: 12345,
+      portForward: { allow: '*' },
+      brokerToken: 'T1',
+    })
+    expect(existsSync(registrationFilePath('coder'))).toBe(true)
+
+    startCalls[0]?.onFatalAuthFailure?.({ brokerToken: 'T1', reason: 'invalid token' })
+
+    await waitFor(() => !existsSync(registrationFilePath('coder')))
+    expect(stopCalls).toEqual([{ name: 'coder', reason: 'fatal-auth' }])
+    const list = await send({ kind: 'list' })
+    expect(list.ok).toBe(true)
+    if (!list.ok) return
+    expect((list.result as ListResult).registrations).toHaveLength(0)
+  })
+
+  test('stale fatal auth (T_old) does not delete a freshly re-registered T_new', async () => {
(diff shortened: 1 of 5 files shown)


devxoul:
@typeey status?

PR #618, opened 2026-06-04
channels: stop the agent claiming a review it never posted

devxoul:
## Background

In a production incident, a channel agent was asked whether it had reviewed a PR. It answered that it had posted an **APPROVED** review — but it had not. The approval it pointed at was authored by a **shared CI-bot account** (a legacy GitHub Actions reviewer that authenticates as the same human-shared login the agent's setup also used), submitted ~2 minutes after the agent said "I'll review", so the timing made the false claim plausible. The agent's own reasoning trace admitted *"I can't definitively confirm whether it was submitted through my session,"* then asserted authorship anyway, repeating the false attribution across three sessions.

Ground truth: there were **zero** `POST .../pulls/<N>/reviews` calls in any of the agent's sessions; the approval correlated exactly with the legacy Actions run. The agent had no way to verify its own actions, so when asked it reconstructed authorship from GitHub state + timing and **guessed** — then presented the guess as fact.

Two mechanisms made this possible and are fixed here:
1. **No proof-of-authorship contract.** The agent treated a review's *author login* as evidence it posted the review. With a shared/`[bot]` identity that is not evidence.
2. **Stranded-text auto-recovery had no disclosure.** The router already auto-posts plain assistant text the model forgot to send via `channel_reply` (a real rescue path for Kimi/Fireworks). A narrated action claim ("I posted the review") would be broadcast verbatim, *as if it were a deliberate, confirmed reply*.

## Summary

- **`classifyReviewAuthorship`** (`src/channels/adapters/github/review-attribution.ts`): authorship proof is an **in-session POST id**, never a login match. A review whose `id` is not in this session's recorded POSTs is `exists-not-mine` — *even when the author login equals the agent's own `slug[bot]`* (a login match without an in-session POST means a different run or the legacy bot authored it). Why id-not-login: the entire bug is that a shared identity makes login worthless as proof.
- **Recovery disclosure** (`src/channels/router.ts`): auto-recovered stranded text is prefixed with a neutral marker (`AUTO_RECOVERY_MARKER`) so it can't pose as a confirmed reply. Why a marker and **not** a regex that suppresses action-claim words: a truth-detecting regex is language-fragile and would swallow *legitimate* recovered replies that truthfully report completed work. The marker is language-agnostic and preserves the rescue behavior while removing its false-confirmation power. It's applied **after** the existing suppression chain, as a disclosure layer rather than a new classifier.
- **GitHub skill** (`typeclaw-channel-github/SKILL.md`): an explicit authorship-proof rule mirroring the helper — never attribute a review you can't tie to your own in-session POST; never infer authorship from timing.

## What this does NOT do

- Does **not** fix the *delivery* side of the incident (a completed reviewer's result being stranded when the channel session rolls over, and the fragile parent-owned POST). Those are tracked separately and need durable completion-routing + reviewer-posts-itself + PR-level idempotency. This PR is scoped to the **honesty/attribution** failure only.
- Does **not** wire `classifyReviewAuthorship` into the skill's `gh` flow yet (the helper is the executable contract; the skill prose references the same rule). Wiring it into a tool path is a follow-up.
- Does **not** add NLP/regex action-claim detection (deliberately rejected, see Summary).

## Review notes

- Load-bearing invariant: **proof of authorship = an in-session POST id, full stop.** The `classifyReviewAuthorship` test `a self-login review WITHOUT an in-session POST record is still not claimable` guards the actual incident — if someone "simplifies" the helper to trust a `slug[bot]` login match, that test must fail.
- The recovery marker changes the text of auto-recovered sends, so several existing recovery tests were updated from exact-match to `toContain(text)` + `toContain(AUTO_RECOVERY_MARKER)`. The new test `recovery marks a false action claim so it cannot pose as a confirmed reply` pins the behavior.
- The marker is a `router.ts`-local constant; adapters receive ordinary text, no adapter contract change.


devxoul:
Closing in favor of a root-cause fix. This PR guards the *symptom* (don't let a false authorship claim through; tell the skill not to lie) but the actual root cause of both this confabulation bug and the sibling 'review ran but never posted' bug is one missing primitive: durable, session-independent per-PR review lifecycle state. A follow-up will introduce that lifecycle store (recording reviewer_started → review_generated → post_succeeded(reviewId)/post_failed, keyed by provider/repo/PR/engagement and surviving session rollover), make idempotent delivery post from it, and let authorship be a lookup over post_succeeded records rather than a session-local guess. The `classifyReviewAuthorship` helper and recovery-disclosure marker from this PR will be folded into that work. Superseding rather than stacking, since the abstraction changes the shape of the helper.

PR #620, opened 2026-06-04
start: untrack files that became gitignored

devxoul:
## Background

`.gitignore` only affects **untracked** files. When TypeClaw adds a new ignore rule to the agent `.gitignore` template, any file that was already committed before the rule existed keeps getting tracked forever — the rule can't retroactively untrack it.

This bit a real agent folder: `public/` was added to the template only recently (`init: gitignore public/ alongside workspace/`), but `public/review.json` had already been committed. The result was a perpetual `modified: public/review.json` in `git status` even though the agent's `.gitignore` clearly lists `public/`.

`refreshGitignore` already rewrites the template on every `typeclaw start`, so the rule is present — but nothing reconciles files that were tracked before the rule landed.

## Summary

On every `start()`, after `refreshGitignore`, reconcile the index: untrack any tracked file that now matches a **truly-ignored** rule, keep it on disk (`--cached`), and fold the removals into the `.gitignore` commit so the rule change and its cleanup land atomically.

Key decisions:

- **Truly-ignored vs system-managed.** The template has two categories. `sessions/`, `memory/`, `channels/`, `todo/` are gitignored **but force-committed** by TypeClaw (auto-backup, dreaming, channels). A blanket `git ls-files -ci` untrack would rip those out of the index and break that contract. So the reconciler matches against the *truly-ignored* set only (sourced from a new `TRULY_IGNORED_PATTERNS` export in `gitignore.ts`, now the single source of truth for both the template and the reconciler) and additionally **fails closed**: even a broad custom `git.ignore.append` pattern like `**` can never untrack a system-managed root.

- **Why `commit-tree` plumbing and not `git commit -- <paths>`.** A pathspec/`--only` commit re-snapshots the listed paths from the **working tree**, so a `git rm --cached` removal of a file that still exists on disk gets silently re-added — the removal never lands. A plain `git commit` gets the removal right but sweeps in unrelated user-staged work. The fix builds the exact tree in a throwaway index seeded from `HEAD`, `commit-tree`s it, and compare-and-swaps `HEAD` via `update-ref` (race-safe). This commits exactly `{ .gitignore + removals }` and leaves unrelated staged work untouched.

- **Best-effort, like `commitSystemFile`.** No-ops when the folder isn't a git repo, Bun is unavailable, there's no `HEAD`, or nothing matches. Never throws — start-time hygiene must not block boot.

## What this does NOT do

- Does not touch files on disk — removals are `--cached` only.
- Does not untrack system-managed dirs under any pattern.
- Does not rewrite existing history or touch already-correct repos (idempotent: a second run untracks nothing).
- Custom `git.ignore.append` entries *are* honored as untrack scope (user asked to ignore them), but the system-managed fail-closed guard still applies. Documented in code.

## Review notes

- Load-bearing: `commitGitignoreWithUntracks` in `src/git/reconcile-ignored.ts`. The temp-index/`commit-tree` dance is deliberate — see the empirically-derived "Why plumbing" comment. Reverting it to a pathspec commit silently reintroduces the re-add bug.
- Invariant to verify: `sessions/memory/channels/todo` stay tracked even with a `*` custom pattern (covered by the "fails closed" test) and unrelated staged work is preserved (covered by the "leaves unrelated staged work untouched" test).
- `gitignore.ts` refactor is byte-identical output — guarded by the existing `buildGitignore` tests.

### DIFF
--- a/src/container/start.ts
+++ b/src/container/start.ts
@@ -4,6 +4,7 @@ import { readFile, writeFile } from 'node:fs/promises'
 import { isAbsolute, join, resolve } from 'node:path'
 
 import { expandMountPath, loadConfigSync, type Config } from '@/config'
+import { commitGitignoreWithUntracks, untrackTrulyIgnoredFiles } from '@/git/reconcile-ignored'
 import { commitSystemFile as commitSystemFileShared } from '@/git/system-commit'
 import { send as sendToDaemon } from '@/hostd/client'
 import type { HttpInfoResult } from '@/hostd/protocol'
@@ -188,7 +189,12 @@ export async function start({
     // trigger the migration commit if the file was legacy.
     await refreshGitignore(cwd)
     const pkgRefresh = await refreshPackageJson(cwd)
-    await commitSystemFile(cwd, GITIGNORE_FILE, 'Update .gitignore')
+    const { untracked } = await untrackTrulyIgnoredFiles(cwd, (await loadTypeclawConfig(cwd)).git.ignore.append)
+    if (untracked.length > 0) {
+      await commitGitignoreWithUntracks(cwd, GITIGNORE_FILE, untracked, 'Untrack newly-ignored files')
+    } else {
+      await commitSystemFile(cwd, GITIGNORE_FILE, 'Update .gitignore')
+    }
     if (pkgRefresh.changed) {
       await commitSystemFile(cwd, pkgRefresh.files, 'Enable bun workspaces (packages/*)')
     }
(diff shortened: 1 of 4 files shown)


devxoul:
Good catch — fixed in 7ffdc611.

`commitGitignoreWithUntracks` now falls back to a plain `git commit` of the real index when the temp-index/commit-tree path can't run (no HEAD, `update-ref` CAS race, transient failure). The fallback is guarded to fire only when the staged set is *exactly* { .gitignore + the removals }, so any pre-existing user-staged work is never swept in — and the `.gitignore` hygiene rewrite (plus the removals) is never left stranded staged-but-uncommitted. Added two tests (fallback commits + repo clean; fallback refuses when extra work is staged) via a small injectable seam, since a real CAS race isn't deterministically reproducible.

devxoul:
@typeey review again

PR #631, opened 2026-06-04
channels: recover text from plain-text channel_reply/send leaks

devxoul:
## Background

A Discord bot using a Kimi-K2 model on Fireworks went completely silent — never posting replies. Root cause: the model was producing its reply but serializing the channel tool call as plain text instead of emitting a real structured tool call. The assistant message body was literally:

```
channel_reply({ text: "yeah i'm here, what's up?" })
```

The channel router (`validateChannelTurn`) detected this leak shape via `isLikelyPlainTextChannelToolCall` and suppressed it outright — logged `suppressed plain_text_channel_tool_call` and returned, posting nothing. From the channel's side the bot looked dead even though the model had produced the exact reply text it intended to send. Verified from container logs (every inbound message went `prompting -> suppressed plain_text_channel_tool_call -> prompted`) and the session transcript (assistant text was literally the channel reply call in plain prose).

## Summary

Recover instead of drop. Extract the `text` argument from leaked plain-text serializations of channel tool calls and deliver it through the existing `source:'system'` recovery path.

Key decisions:

- `skip_response(...)` is **still suppressed** — it's a genuine decline, never a user-facing reply. Posting it would leak "I'm staying silent" plumbing. The suppression log is now `suppressed plain_text_channel_skip_response` to distinguish the two paths.
- Extraction failure **falls back to the historical safe behavior** — if no text arg is found, the value is empty, or the serialization is fully unrecoverable, the message is dropped rather than leaking plumbing text.
- The extractor (`extractPlainTextChannelToolCallText`) is a tolerant single-pass scanner, not a regex. It handles double/single/unquoted keys, escaped quotes, `\n`/`\t` escape sequences, and mid-serialization truncation. The model sometimes truncates, so a half-leaked call is recovered too.
- `channel_send` ignores the model-supplied routing args (`adapter`, `chat`, `thread`). Recovery always routes back through the current channel context, never to a model-supplied destination. This prevents destination spoofing where a leaked call with a crafted target could route the reply somewhere unexpected.
- Extracted text is re-run through the existing downstream guards (`endsWithNoReplySignal`, `isUpstreamEmptyResponseSentinel`, `isLikelyKimiChannelToolLeak`, `isLikelyPlainTextChannelToolCall`) so a nested or no-reply payload cannot re-leak through the recovery path.

Two new exported helpers land near the existing detector:

- `getPlainTextChannelToolCallKind` — returns `'reply'`, `'send'`, `'skip'`, or `null`
- `extractPlainTextChannelToolCallText` — returns the extracted text string or `null`

## Preview

No visual output — library-internal changes only.

## What this does NOT do

- Does NOT change behavior for any correctly-structured tool call.
- Does NOT attempt to recover `skip_response(...)` — that's a deliberate agent choice to stay silent, not a serialization failure.
- Does NOT honor the routing arguments from a leaked `channel_send` call (deliberate; see destination-spoofing note above).
- Does NOT attempt recovery when extraction yields an empty or whitespace-only string.

## Review notes

Load-bearing files: `src/channels/router.ts` (+106/−5) and `src/channels/router.test.ts` (+159/−12).

The `validateChannelTurn` branch for `isLikelyPlainTextChannelToolCall` is replaced with a kind-dispatch: `skip` suppresses, `reply`/`send` extracts and recovers, `null` passes through.

The two existing suppression tests for plain-text `channel_reply` and `channel_send` are **flipped** to assert recovery of the text arg. `skip_response` suppression remains. "Prose that merely mentions channel_reply/skip_response passes through untouched" also remains.

New integration tests cover: unquoted-key Kimi shape, truncated serialization, `channel_send` ignoring model destination, extracted-text-that-is-a-no-reply-signal being suppressed, and unextractable-arg being dropped.

A focused `describe` block unit-tests all parsing edge cases of `getPlainTextChannelToolCallKind` and `extractPlainTextChannelToolCallText`.

**Note on test suite:** `bun test --parallel src/channels/router.test.ts` passes 363/363. Full suite: 7192 pass, 1 fail. The one failure (`resolveSelfPackageJsonPath > points at this package manifest` in `src/update/index.test.ts`) is a pre-existing environmental artifact — that test asserts the checkout directory is named `typeclaw`; it fails only when run from a worktree with a different directory name. It is unrelated to this change (no `src/update/` code is touched) and passes in CI.

Checks: `bun run typecheck` clean, `bun run lint` 0 errors (65 pre-existing warnings, none in changed files), `bun run format:check` clean.

### DIFF
--- a/src/channels/router.test.ts
+++ b/src/channels/router.test.ts
@@ -18,6 +18,8 @@ import {
   CHANNEL_MAX_OUTPUT_TOKENS,
   createChannelRouter,
   DUPLICATE_SEND_ERROR,
+  extractPlainTextChannelToolCallText,
+  getPlainTextChannelToolCallKind,
   MAX_CHANNEL_SENDS_PER_TURN,
   MAX_POLICY_DENIED_CHANNEL_SENDS_PER_TURN,
   MAX_TYPING_HEARTBEAT_MS,
@@ -2474,7 +2476,7 @@ describe('ChannelRouter channel-turn protocol', () => {
     expect(logs.some((m) => m.includes('suppressed kimi_tool_call_leak'))).toBe(false)
   })
 
-  test('suppresses leaked plain-text channel_reply(...) function-call serialization instead of posting it to the channel', async () => {
+  test('recovers the text arg from a leaked plain-text channel_reply(...) serialization instead of dropping it', async () => {
     const dir = await tempDir()
     const logs: string[] = []
     const sent: Array<{ text: string }> = []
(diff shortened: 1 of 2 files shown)


devxoul:
Good catch — fixed in 2fb12d98.

The classifier regex gated on a double quote (\`[^)]*"\`), so single-quoted leaks like \`channel_reply({text: 'it's me'})\` never reached the extractor and fell through to the raw-plumbing recovery. Widened it to accept either quote (\`[^)]*["']\`) so the classifier and extractor agree on accepted quote styles, and added an end-to-end \`validateChannelTurn\` test for the single-quoted shape (plus a comment documenting why the single-quote arm is load-bearing).

devxoul:
@typeey review again

devxoul:
@typeey are you up?

devxoul:
@typeey review

devxoul:
@typeey approve then

devxoul:
@typeey approve using gh

PR #635, opened 2026-06-04
channels: recognize a bare-slug @mention of an App-auth github bot

devxoul:
## Background

Observed in production (typeey on PR #631): a human commented "@typeey review again" on a PR the bot had been reviewing, and the bot stayed silent — it did not respond until a follow-up "@typeey are you up?" was posted. Both were direct @-mentions, yet the first was silently dropped.

Root cause is the GitHub App decoy-reviewer pattern. A GitHub App cannot be a `requested_reviewer`, so the supported workaround is a decoy user account whose login is the App slug (e.g. `typeey`). But the App's *actor* login — what `auth.getSelf().login` returns and what becomes `selfLogin` — is the `[bot]`-suffixed form `typeey[bot]`. GitHub renders a human's mention of the App as the bare slug `@typeey` (no `[bot]`, and there is no way to type the suffix).

The inbound classifier (`src/channels/adapters/github/inbound.ts`, `buildInbound`) set `isBotMention` by matching the comment body against `@${selfLogin}` only — i.e. `@typeey[bot]`. A body containing `@typeey` never matched, so `isBotMention` was `false`. In `decideEngagement` the guaranteed mention-engage (`config.trigger.includes('mention') && message.isBotMention`) was skipped; GitHub inbounds also carry `mentionsOthers=false`/`replyToBotMessageId=null`, so the comment fell through to the engagement fallback and was observed (dropped). The follow-up engaged only because an unrelated path (sticky credit from the now-recorded participant) happened to wake the session — flaky by nature.

## Summary

Resolve the set of @-handles that count as a self-mention, instead of matching a single login. New `resolveBotMentionLogins(selfLogin, authType)` returns the actor `selfLogin` always, plus the decoy slug under App auth — reusing the existing `resolveDecoyReviewerLogin` derivation (strip the `[bot]` suffix). `buildInbound` now takes this list and matches the body against any of them.

- Why reuse `resolveDecoyReviewerLogin`: it is the single existing seam that already encodes the slug-derivation convention used to drop the decoy reviewer; keeping mention detection on the same derivation means the classifier and the decoy logic can never drift on what "the bot's slug" is.
- PAT auth is unchanged: a PAT bot is a real user requested by its actual login, so there is no decoy and only `selfLogin` applies. A bare-prefix collision (e.g. `@typeclaw` for `typeclaw-bot`) does NOT register as a mention under PAT.

## What this does NOT do

- Does NOT change `classifyReviewRequest` / `classifyOpenedReviewTrigger` — those already synthesize `isBotMention: true` for review-request/opened triggers and still receive `selfLogin` for their own decoy/self-loop checks.
- Does NOT change PAT-auth behavior in any way (no decoy is derived).
- Does NOT touch the engagement gate, sticky credit, or any posting code — this is purely inbound mention classification.

## Review notes

Load-bearing change: `buildInbound`'s mention check moves from `text.includes('@' + selfLogin)` to `textMentionsBot(text, resolveBotMentionLogins(selfLogin, authType))`. The `mention` set is computed once at the top of `classifyGithubInbound` and threaded into all seven `buildInbound` call sites; the two `selfLogin` args passed into `classifyReviewRequest`/`classifyOpenedReviewTrigger` are intentionally left as `selfLogin`.

New tests in `inbound.test.ts` (describe: "isBotMention decoy-aware detection (App auth)"): bare-slug `@typeclaw` on a `typeclaw[bot]` actor → isBotMention=true; full `@typeclaw[bot]` still true; unrelated `@someone-else` false; and PAT auth where `@typeclaw` must NOT match `typeclaw-bot` but `@typeclaw-bot` must.

Checks: `bun run typecheck` clean, `bun run lint` 0 errors (65 pre-existing warnings, none in changed files), `bun run format:check` clean. `bun test --parallel src/channels/adapters/github/` 277/277 pass. Full suite: 1 failure, `resolveSelfPackageJsonPath > points at this package manifest` — a pre-existing environmental artifact that asserts the checkout dir is named `typeclaw`; it fails only from a worktree with a different directory name and passes in CI (unrelated to this change; no `src/update/` code touched).

### DIFF
--- a/src/channels/adapters/github/inbound.test.ts
+++ b/src/channels/adapters/github/inbound.test.ts
@@ -154,6 +154,19 @@ describe('classifyGithubInbound', () => {
       expect(msg?.isBotMention).toBe(false)
     })
 
+    it('does not treat a longer login sharing the decoy-slug prefix as a self-mention', () => {
+      // The decoy slug for `typeclaw[bot]` is `typeclaw`; `@typeclaw-bot` is a
+      // DIFFERENT GitHub user. A substring check would false-positive here, so
+      // the matcher must respect GitHub-login boundaries (- is a login char).
+      const msg = classifyGithubInbound(
+        'issue_comment',
+        issueCommentPayload({ pullRequest: true, body: '@typeclaw-bot can you review?' }),
+        'typeclaw[bot]',
+        { authType: 'app' },
+      )
+      expect(msg?.isBotMention).toBe(false)
+    })
+
     it('does not derive a decoy slug under PAT auth (only the real login mentions)', () => {
       // PAT bots are real users requested by their actual login; there is no
       // decoy slug, so a bare-prefix collision must not register as a mention.
--- a/src/channels/adapters/github/inbound.test.ts
+++ b/src/channels/adapters/github/inbound.test.ts
@@ -118,6 +118,63 @@ describe('classifyGithubInbound', () => {
     })
   })
 
+  // Regression: under GitHub App auth selfLogin is the actor login `slug[bot]`,
+  // but a human mentions the App by its bare slug `@slug` (the decoy account).
+  // The classifier must recognize the bare-slug mention or a direct
+  // "@typeey review again" lands with isBotMention=false and the engagement
+  // mention gate never fires (the comment is silently observed/dropped).
+  describe('isBotMention decoy-aware detection (App auth)', () => {
+    it('recognizes a bare-slug @mention of an App bot whose actor login is slug[bot]', () => {
+      const msg = classifyGithubInbound(
+        'issue_comment',
+        issueCommentPayload({ pullRequest: true, body: '@typeclaw review again' }),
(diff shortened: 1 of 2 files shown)


devxoul:
Good catch — fixed in 8b1e749a.

The bare-slug matcher used `text.includes(`@${login}`)`, so under App auth the decoy slug `typeclaw` (from actor `typeclaw[bot]`) false-matched `@typeclaw-bot` — a different GitHub user — as a self-mention.

Replaced the substring check with a GitHub-login boundary check: a `@login` token only counts when the following character is absent or not a login char (`[A-Za-z0-9-]`). `@typeclaw[bot]` still matches the decoy slug (the trailing `[` is not a login char), and `@typeclaw-bot` no longer does. Added an App-auth regression test where `selfLogin='typeclaw[bot]'` and `@typeclaw-bot can you review?` leaves `isBotMention=false`.

PR #642, opened 2026-06-04
github: mark App installation token as recommended auth

devxoul:
## Background

The GitHub channel auth-type selector (used by both `typeclaw channel add` and `typeclaw init`) presented two options with equal visual weight:

```
○ GitHub App installation token
○ Fine-grained personal access token
```

Operators who haven't read the docs have no signal that App tokens are the better choice. App installation tokens are short-lived and automatically rotate, giving them a meaningfully smaller blast radius than long-lived PATs.

## Summary

Appends `(recommended)` to the App option label at both call sites — `promptGithubCredentials` in `src/cli/channel.ts` and `runGithubFlow` in `src/cli/init.ts` — so the selector reads:

```
○ GitHub App installation token (recommended)
○ Fine-grained personal access token
```

Both sites present the identical selector, so both are updated for consistency. The change mirrors the existing `(recommended)` suffix convention already present in the same selectors (e.g. the Cloudflare Quick Tunnel option).

No logic, schema, or runtime behavior changes. The `value` field (`'app'`) is untouched.

## Preview

Before:
```
○ GitHub App installation token
○ Fine-grained personal access token
```

After:
```
○ GitHub App installation token (recommended)
○ Fine-grained personal access token
```

## What this does NOT do

- Does not change which auth method is accepted or how tokens are validated.
- Does not reorder the selector options.
- Does not affect any non-GitHub channel.

## Review notes

Two one-line edits, one per call site. The only thing to verify is that `value: 'app'` is untouched and both labels match.

`bun run typecheck`, `bun run lint`, and `bun run format` all pass. The single test failure (`resolveSelfPackageJsonPath`) is a pre-existing environmental artifact unrelated to this change.

### DIFF
--- a/src/cli/channel.ts
+++ b/src/cli/channel.ts
@@ -631,6 +631,7 @@ async function promptGithubCredentials(cwd: string): Promise<{
       { value: 'pat', label: 'Fine-grained personal access token' },
       { value: 'app', label: 'GitHub App installation token (recommended)' },
     ],
+    initialValue: 'app',
   })
   if (isCancel(authType)) {
     cancel('Aborted.')
--- a/src/cli/channel.ts
+++ b/src/cli/channel.ts
@@ -629,7 +629,7 @@ async function promptGithubCredentials(cwd: string): Promise<{
     message: 'GitHub authentication type',
     options: [
       { value: 'pat', label: 'Fine-grained personal access token' },
-      { value: 'app', label: 'GitHub App installation token' },
+      { value: 'app', label: 'GitHub App installation token (recommended)' },
     ],
   })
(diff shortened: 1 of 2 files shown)


devxoul:
Good catch — fixed in 28679160. Added `initialValue: 'app'` to both GitHub auth-type selectors (`src/cli/channel.ts` and `src/cli/init.ts`), so the recommended GitHub App option is now the default selection. This matches the existing convention used by the nearby recommended selectors (Cloudflare Quick Tunnel `initialValue: 'cloudflare-quick'`, Slack bot token `initialValue: 'bot'`). The UI no longer says "recommended" while defaulting to the other option.

PR #657, opened 2026-06-05
tui: brand banner, spinners, status bar, and init welcome

devxoul:
## Background

The TUI had no visual identity — a plain text stream with no version card, no
"thinking" signal between submit and first token, and no cost visibility after
a turn. The `typeclaw init` wizard likewise started with a bare
`Initializing TypeClaw...` line and ended with a plain success message. This
PR introduces a cohesive brand layer across both surfaces, built on the typeey
mascot palette (cornflower `#5B7FD4` / amber `#E78F37`).

A note on the accent choice: the mascot's true navy (`#182A5B`) is too dark to
read against dark terminals, so the accent lifts to a cornflower of the same
hue that passes minimum contrast on common default backgrounds.

### TUI

- **Startup banner**: ANSI Shadow "typeclaw" wordmark + a session/version/URL
  card rendered immediately on `connected`.
- **Thinking loader**: animated "thinking…" indicator between prompt submit and
  first token, using `pi-tui`'s `Loader` — no hand-rolled `setInterval`.
  Reload and restart states get the same treatment.
- **Boxed assistant turn headers**: `╭─ typeclaw · HH:MM:SS ───────` before
  each assistant reply so turn boundaries are visually distinct.
- **Bottom status bar**: persistent `connection · session · version · ⚡ tok · $cost` line that updates live as turns complete.

### server (`src/server/index.ts`, `src/shared/protocol.ts`)

The `done` `ServerMessage` now carries an optional `usage` field
`{ input, output, totalTokens, cost }`. The server buffers usage from the
existing `message_end` event (via `readMessageUsage`) into
`SessionState.lastUsage` and attaches it to the outgoing `done` frame.
The change is backward compatible (the field is optional; old TUI versions
receive a plain `done`).

### init (`src/cli/ui.ts`, `src/cli/init.ts`)

- **Welcome banner**: cornflower wordmark + tagline printed before the clack
  wizard, fully NO_COLOR-aware (zero ANSI when set), width-aware (compact
  `typeclaw` fallback below ~70 cols), and suppressed on non-TTY/piped output.
- **Hatched outro**: `✦ hatched!` in amber/cornflower replaces the bare success
  line — same NO_COLOR and TTY guards.

### Shared wordmark (`src/shared/wordmark.ts`)

`WORDMARK_LINES` / `WORDMARK_WIDTH` / `COMPACT_WORDMARK` live once here and
are imported by both the TUI banner renderer and the init CLI. Neither
consumer duplicates the art; each applies its own color layer.

## What this does NOT do

- Does not change wire protocol semantics: `done.usage` is optional and
  ignored by clients that don't read it.
- Does not add a truecolor dependency: the `CORNFLOWER_FG` / `AMBER_FG` SGR
  strings are inlined as constants, gated on `colorsEnabled()`.
- Does not render the banner in piped/CI contexts (`isTTY` guard on every
  print path).

## Review notes

- Load-bearing: `state.lastUsage = null` is reset at each turn start (before
  `session.prompt()`), so a turn that produces no `message_end` sends a bare
  `done` rather than stale usage from the previous turn.
- `readDoneUsage` delegates to the existing `readMessageUsage` rather than
  re-implementing the usage-extraction logic — keeps the two in sync.
- The `forwardSessionEvents` signature changed from `(ws, session, …)` to
  `(ws, state, …)` to give the subscriber write access to `state.lastUsage`.
  All callers updated.
- `formatAssistantHeader` renders an open box top-rule only — no closing rule
  — because the assistant reply length is unknown at the start of a streaming
  turn.

## Testing

Full suite: 7363 pass / 0 fail / 1 pre-existing skip. `typecheck` clean.
`lint` 0 errors. `format:check` clean.

New test coverage:

- `src/shared/wordmark.test.ts`: shape (6 rows), no escapes in source art,
  `WORDMARK_WIDTH` derivation, trailing-space preservation, compact fallback.
- `src/cli/ui.test.ts`: `cornflower` colors under forced color / NO_COLOR;
  `renderInitWelcome` — full art vs compact (narrow), color vs no-color,
  non-TTY suppression; `renderHatchedFlourish` — branded vs plain vs non-TTY.
- `src/tui/format.test.ts`: `formatAssistantHeader` box label + timestamp
  fallback; `formatTokenCount` / `formatCost` / `formatUsageSummary` boundary
  values.
- `src/tui/index.test.ts`: banner renders on `connected` with session/version;
  `done.usage` updates the bottom status bar with token count and cost.

### DIFF
--- a/src/cli/init.ts
+++ b/src/cli/init.ts
@@ -36,7 +36,16 @@ import { API_KEY_DASHBOARD_URL, validateApiKey, type KeyValidationResult } from 
 
 import { buildOAuthCallbacks } from './oauth-callbacks'
 import { CANCEL_SYMBOL, promptPrivateKeyPem } from './prompt-pem'
-import { c, done, errorLine, printDiscordInviteHint, printSlackAppManifestSetup } from './ui'
+import {
+  c,
+  cornflower,
+  done,
+  errorLine,
+  printDiscordInviteHint,
+  printHatchedFlourish,
+  printInitWelcome,
+  printSlackAppManifestSetup,
+} from './ui'
 
 // ESC and Ctrl+C both produce clack's cancel symbol (the keypress layer
 // aliases both to the same "cancel" action — there's no way to tell them
@@ -119,6 +128,7 @@ export const init = defineCommand({
       }
     }
 
+    printInitWelcome()
     intro('Initializing TypeClaw...')
     log.info('Press ESC at any prompt to go back. Press ESC twice in a row to abort.')
 
@@ -272,7 +282,8 @@ export const init = defineCommand({
           'Claim ownership before chatting',
         )
       }
-      done({ title: c.green('Hatched. Your agent is ready.'), hints })
+      printHatchedFlourish()
+      done({ title: `${cornflower('✓')} ${c.bold('Hatched.')} Your agent is ready.`, hints })
     }
   },
 })
--- a/src/cli/ui.test.ts
+++ b/src/cli/ui.test.ts
@@ -1,21 +1,28 @@
 import { afterEach, beforeEach, describe, expect, test } from 'bun:test'
 
 import { SLACK_SLASH_COMMAND_NAMES } from '@/channels/adapters/slack-bot'
+import { WORDMARK_LINES } from '@/shared/wordmark'
 
 import {
   c,
+  cornflower,
   done,
   errorLine,
   link,
   prepareStdinForClack,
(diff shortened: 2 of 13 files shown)


devxoul:
Fixed in 9af3499c: the restart_result handler now stops, removes, and clears restartLoader before appending the final status line, mirroring the reload_result cleanup. Added a regression test asserting the loader appears in exactly one repaint (verified it fails without the fix).

PR #678, opened 2026-06-06
fix(dockerfile): pin Bun to 1.3.13-slim to dodge 1.3.14 ARM64 crash

devxoul:
## Background

A bug report from an Apple-silicon (ARM64) host: the container reports a Bun
version, but every attempt to run an external package crashes.

```
$ bun run dist/src/platforms/webex/cli.js --help
[1]    3 Aborted                 bun run dist/src/platforms/webex/cli.js --help
EXIT: 134
```

`bunx`, `bun add`, and `bun run <pkg>` all abort with SIGABRT (exit 134) or a
`NotDir` internal error; even `bun upgrade` self-fails with `StackOverflow`. The
agent-messenger CLIs (`agent-webex`, etc.) were effectively unrunnable inside the
container.

Root cause is upstream Bun, surfaced by our own tag policy. The base image
`FROM` used the **floating** `oven/bun:1-slim` tag, which silently rolled the
base onto **Bun 1.3.14**. That release has two ARM64 Linux regressions that hit
exactly the external-package launch path:

- A self-referencing symlink created in `createFakeTemporaryNodeExecutable`
  (`/tmp/bun-node-*/node -> bun -> bun -> ...`) -> ELOOP / SIGABRT
  ([oven-sh/bun#30711](https://github.com/oven-sh/bun/issues/30711)).
- `require()`-of-ESM aborting with a PAC IB trap (silent SIGTRAP/SIGABRT)
  ([oven-sh/bun#30281](https://github.com/oven-sh/bun/issues/30281)).

The fix for #30711 ([oven-sh/bun#30713](https://github.com/oven-sh/bun/pull/30713))
is not yet in any stable release, so there is no forward fix to roll to -- the only
remedy today is to stop landing on 1.3.14.

## Summary

Pin the base image to an **exact patch**, `oven/bun:1.3.13-slim` -- the last
known-good release -- instead of the floating `1-slim` tag. Floating tags make
base-image rebuilds non-deterministic; a single upstream patch release can break
every installed agent on the next rebuild, which is exactly what happened here.

The pin lives in one `BUN_IMAGE` constant that feeds **both** the prebuilt base
image (`buildBaseDockerfile`) and the inline dev/test path (`renderInlineHead`),
so production and dev mode can't drift onto different Bun versions.

The three CI Bun pins (`ci.yml`, `release.yml`, `base-image.yml`) move
`1.3.14 -> 1.3.13` so the runner validating/releasing the image uses the same Bun
as the container. They were already pinned (not `latest`); this just holds them
at the known-good patch and documents why.

## What this does NOT do

- Does **not** rebuild or publish a new `typeclaw-base` image. That happens via
  the Release workflow; this PR only changes what the next release will build.
- Does **not** touch the affected agent folders on hosts -- they pick up the fix
  on their next `typeclaw start` once a release ships the new base.
- Does **not** work around the upstream bug in our own code; it avoids the broken
  Bun version entirely.

## Review notes

- Load-bearing change: `src/init/dockerfile.ts` `BUN_IMAGE = oven/bun:1.3.13-slim`.
  Invariant to verify -- `bun run scripts/emit-base-dockerfile.ts | grep ^FROM`
  now prints `FROM oven/bun:1.3.13-slim`.
- The drift guard in `dockerfile.test.ts` asserts base and per-agent Dockerfiles
  share the same `FROM`; tests updated to the new tag. The two
  `not.toContain('FROM oven/bun:...')` assertions were widened to `FROM oven/bun:`
  so they stay meaningful regardless of the pinned patch.
- The two `start.test.ts` "stale Dockerfile" input fixtures still use `1-slim` on
  purpose -- they represent an *old* on-disk Dockerfile that `refreshDockerfile`
  overwrites.
- Bump to the next stable once oven-sh/bun#30713 lands; the `BUN_IMAGE` comment
  flags that follow-up.

### DIFF
--- a/.github/workflows/base-image.yml
+++ b/.github/workflows/base-image.yml
@@ -39,7 +39,9 @@ jobs:
           # Dockerfile this workflow emits doesn't itself depend on the
           # Bun version (only its TypeScript imports do); matches the pin
           # in ci.yml and release.yml so all three jobs use the same Bun.
-          bun-version: 1.3.14
+          # Held at 1.3.13 — see the dockerfile.ts BUN_IMAGE comment for the
+          # 1.3.14 ARM64 SIGABRT regressions (oven-sh/bun#30711, #30281).
+          bun-version: 1.3.13
 
       - name: Generate base Dockerfile
         # No `bun install` needed — scripts/emit-base-dockerfile.ts only
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,10 @@ jobs:
           # to resolve the version, which intermittently 401s under GHA
           # token contention; a pinned version skips that lookup and goes
           # straight to the release download URL.
-          bun-version: 1.3.14
+          # Held at 1.3.13 (not 1.3.14) to match the container Bun pin in
+          # src/init/dockerfile.ts — 1.3.14 SIGABRTs on ARM64 Linux when
+          # launching external packages (oven-sh/bun#30711, #30281).
+          bun-version: 1.3.13
 
       - name: Install dependencies
         run: bun install --frozen-lockfile
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -44,7 +44,10 @@ jobs:
           # to resolve the version, which intermittently 401s under GHA
           # token contention; a pinned version skips that lookup and goes
           # straight to the release download URL.
-          bun-version: 1.3.14
+          # Held at 1.3.13 (not 1.3.14) to match the container Bun pin in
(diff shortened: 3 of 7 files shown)


devxoul:
Closing: this fix is incorrect. Reproduced the crash on ARM64/OrbStack and proved via strace that it is NOT a Bun version regression — Bun 1.3.13 crashes identically. The real root cause is the bwrap sandbox's `--tmpfs /proc`: the spawned child of `bunx` reads `/proc/self/fd` and `/proc/self/maps` which don't exist under the empty tmpfs proc, surfaced as Bun's `NotDir`. Fix moves to src/sandbox (opt-in real-proc strategy). Superseded by a follow-up PR.

PR #679, opened 2026-06-06
fix(agent): route explicit research requests to the researcher subagent

devxoul:
## Problem
When a user explicitly asks the agent to "do a research" / "investigate this", the agent sometimes skips the `researcher` subagent and answers directly — from training memory or a single inline `web_search`. The user asked for the work to be done; getting back a quick guess instead is a silent failure to follow the instruction.

## Root cause
The only research-routing guidance in the system prompt is advisory, not binding:
- `src/agent/system-prompt.ts` (Mode A): "a fuzzy question that needs decomposition … goes to `researcher`"
- `src/agent/tools/websearch.ts`: "PREFER delegating to the `scout` subagent … Only call this tool directly for a single query …"

Nothing tells the model that an explicit research directive is a non-negotiable instruction to delegate. With multiple legitimate options (answer inline, `web_search`, `scout`, `researcher`), the model exercises discretion and can downgrade an explicit research request to an inline answer.

## Fix
Add one binding sentence to Mode A:

> When the user *explicitly* says "research"/"investigate" (or equivalent), that's a binding instruction to spawn `researcher` — not to answer from training memory or a single `web_search`, even if you think you know the answer.

Deliberately kept to a single imperative — no rationale padding, and no duplicate copy in the `spawn_subagent` tool description — so it doesn't bloat an already-dense prompt (net ~+50 tokens). The existing scout-vs-researcher depth heuristic for implicit information needs is unchanged; this only affects behavior when the user explicitly asks for research.

## Tests
A drift-guard test in `src/agent/system-prompt.test.ts` asserts the rule's load-bearing phrases survive in the rendered prompt, so it can't be dropped without CI failing.

- `bun run typecheck` — clean
- `bun run lint` — 0 errors
- `bun run test` — 7651 pass, 0 fail


devxoul:
Addressed in `c0b6b229`.

**1. Downgrade path in the rule (`system-prompt.ts`).** Mode A now reads:

> When the user *explicitly* says "research"/"investigate" (or equivalent), you MUST spawn `researcher` — answering from training memory or a single inline `web_search` does not satisfy the request, even if you think you know the answer. (Fanning out `scout`/`explorer` underneath is fine, but it does not replace `researcher`.)

`researcher` is now mandatory; scout/explorer are explicitly demoted to optional helpers *underneath* it, never a substitute.

**2. Loophole repeated in `spawn-subagent.ts`.** That sentence was already removed in an earlier revision on this branch — the tool description no longer frames `scout` as a research alternative. To keep it from creeping back, I added a guard test (see below).

**3. Drift guard too weak.** Replaced the loose-substring check with an invariant assertion on both prompt surfaces:
- `system-prompt.test.ts` now asserts the mandatory rule (`MUST spawn \`researcher\``, "does not satisfy the request", and the "does not replace `researcher`" clause).
- `spawn-subagent.roster.test.ts` adds a guard that `spawnSubagentDescription` never frames `scout` as a research handler — pinning shut the second surface you flagged.

Full gate green: typecheck / lint / format:check clean, 7653 tests pass.

PR #688, opened 2026-06-07
docs: fix plugin API reference to match src/plugin

devxoul:
## Background

The plugin docs had drifted far enough from `src/plugin/types.ts` and `src/plugin/define.ts` that they described an API that no longer exists. On `reference/plugin-api.mdx` every one of the eight code blocks was wrong, and the `Tool`, `Subagent`, `Cron`, `Hooks`, and `ToolBeforeResult` blocks would fail to typecheck if a plugin author copied them. The same drift had leaked into the concept page (`plugins-and-stages.mdx`) and the guide (`write-a-plugin.mdx`).

Concretely, the docs claimed:

- `definePlugin({ name, factory })` — but there is no `name` field (the name is derived from `package.json#name` via `derivePluginNameFromPackage`), and the factory key is `plugin` and is **async**.
- `PluginContext` exposes `stream` / `pluginDir` / `containerName` — none of which exist. The real context is `name`, `version`, `agentDir`, `config`, `logger`, `permissions`, `github`, `spawnSubagent`.
- A `PluginContributions` type with `channels` and array-shaped `skills`/`crons` — the real type is `PluginExports`, `skills`/`cronJobs` are `Record`s, there is no `channels` field, and `skillsDirs` / `hooks` / `doctorChecks` were missing entirely.
- `Tool` uses `inputSchema` and returns arbitrary `TOut` — it's `parameters` and returns a `ToolResult`.
- `ToolBeforeResult` has an `{ ok: 'ack'; severity; guard; message }` variant "used by the security plugin" — this was fabricated. The real shape is `void | { block: true; reason }`, and the `security` plugin returns the `block` variant (`src/bundled-plugins/security/index.ts:82`).
- Hooks: only 3 documented; there are 8, each taking a second `HookContext` arg.

This surfaced from a reader's question that conflated "plugins can't drive the agent" with "plugins can't intercept the agent's shell execution." The first is true; the second is false — and the docs didn't make that explicit, so the wrong inference was reasonable.

## Summary

Rewrote the three plugin docs against the actual source, which is now cited at the top of the reference page.

The reference page (`plugin-api.mdx`) is corrected block-by-block and gains the previously-undocumented surfaces: `commands` (the host/container/either `PluginCommand` split + `defineCommand`), `doctorChecks` (`--fix` flow), `skillsDirs`, the `handler` cron kind with `CronHandlerContext`, and the `defineTool`/`defineSubagent` + builtin-tool-ref helpers.

The concept page and guide get the matching `plugin`/`parameters`/`ToolResult`/`PluginExports` corrections, and the `ctx.stream` subscription guidance is replaced with the real reactive mechanism: hooks.

A new "Hooks intercept the agent's own tools" section makes the load-bearing fact explicit — `tool.before`/`tool.after` fire for the agent's **own** built-in tools (`read`/`bash`/`edit`/`write`/`grep`/`find`/`ls`), not just plugin tools, and a `tool.before` returning `{ block: true, reason }` aborts the call, including a `bash` command. This is verifiable in `src/agent/plugin-tools.ts`: `buildBuiltinPiToolOverrides` wraps every pi builtin via `wrapAgentToolAsCustomToolDefinition`, which runs `hooks.runToolBefore` before execute and throws on a block — and it's how the bundled `security` plugin guards shell execution.

## What this does NOT do

- No source changes. Docs only (`.mdx`).
- Does not touch the other reference pages whose `scheduledByRole` usage is correct (that field is real on `cron.json` user jobs / cron origins; it's only the plugin-contributed `PluginCronJob` type that omits it and defaults to `owner`).

## Review notes

The authority for every claim is `src/plugin/types.ts`, `src/plugin/define.ts`, and `src/agent/plugin-tools.ts`. The one previously-fabricated claim worth confirming is the `ToolBeforeResult` shape: it is `void | { block: true; reason }` (`types.ts:213`), and `security/index.ts:82` returns `{ block: true, ... }` — there is no `ack`/`severity` hook-return variant.

Pre-commit (`format:check`, `lint`, `typecheck`) passes. The only failing test on this branch is the pre-existing `typeclaw.schema.json` drift guard, which also fails on clean `main` and is unrelated to these doc-only changes.

### DIFF
--- a/docs/content/docs/concepts/plugins-and-stages.mdx
+++ b/docs/content/docs/concepts/plugins-and-stages.mdx
@@ -18,35 +18,43 @@ Pulling the boundary down to the plugin level would buy almost no extra security
 
 ## What plugins receive
 
-The `factory(ctx)` callback gets a `PluginContext` carrying:
+The `plugin(ctx)` factory is an async callback that gets a `PluginContext` carrying:
 
 - **`ctx.config`** — the parsed, frozen config block for this plugin (validated by your `configSchema`). Boot-only fields are snapshots; reloadable fields update on `typeclaw reload`.
 - **`ctx.permissions`** — the live permission service. Plugins query it (`ctx.permissions.has(origin, perm)`) rather than implementing their own checks.
-- **`ctx.stream`** — the in-process message stream. Plugins subscribe to broadcasts, publish to typed targets.
-- **`ctx.spawnSubagent(name, payload, options)`** — for plugins that contribute subagents and want to invoke them.
+- **`ctx.spawnSubagent(name, payload, options)`** — for plugins that contribute subagents and want to invoke them; provenance is resolved from the spawning origin.
 - **`ctx.agentDir`** — absolute path to the agent folder (mounted at `/agent` in the container).
+- **`ctx.name` / `ctx.version` / `ctx.logger` / `ctx.github`** — the plugin's derived name, its package version, a namespaced logger, and GitHub token resolution.
 
 Everything plugins talk to flows through this object. There are no module-global imports of agent state, no singletons reaching into the runtime — the surface is the `ctx`.
 
 ## What plugins can contribute
 
-A plugin's `factory` returns an object with optional fields:
+A plugin's factory returns a `PluginExports` object with optional fields:
 
 - **`tools`** — Zod-typed functions the agent can call
-- **`skills`** — markdown procedures the agent loads on demand
+- **`skills`** / **`skillsDirs`** — markdown procedures the agent loads on demand, inline or from directories
 - **`subagents`** — focused worker sessions with their own system prompt and payload schema
-- **`channels`** — new inbound/outbound message platforms
-- **`crons`** — scheduled prompts or exec commands contributed at boot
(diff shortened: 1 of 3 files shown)


devxoul:
Both confirmed against source and fixed in 1665c253.

1. **Plugin naming** — you're right, it's path-dependent. `loadLocal` names a local/path plugin from the resolved basename (`src/plugin/loader.ts:41`); only `loadNpm` uses `package.json#name` via `derivePluginNameFromPackage` (`:77`). Reworded the naming notes in all three docs to cover both, and fixed the guide's config-key explanation (the example is a local `./packages/notify`, so the key is the basename `notify`, not `package.json#name`).

2. **`ctx.config` immutability** — also right. `createPluginContext` `Object.freeze`s the ctx wrapper (`context.ts:26`) but `manager.ts` passes `parsed.data` through unfrozen (`:91`,`:103`), so the config object's nested fields are mutable. Dropped 'frozen' from the config descriptions; now they say the ctx is frozen but the config object isn't, so treat it as read-only.

Pre-commit (format:check / lint / typecheck) passes.

PR #707, opened 2026-06-08
fix(sandbox): default to a cap-free 'proc-bind' /proc strategy so OrbStack bunx works

devxoul:
## Problem

Sandboxed external-package CLIs — `bunx agent-*`, `bun add <pkg>`, `bun run <pkg-bin>` — still abort with Bun's opaque `error: An internal error occurred (NotDir)` on **OrbStack / Apple Silicon**, the most common host, despite four prior fixes (#672, #683, #696, #697). This is the 4th attempt at the same crash.

Reproduced live on the failing OrbStack container (Bun 1.3.14, aarch64): the exact `--tmpfs /proc` sandbox argv → `error: An internal error occurred (NotDir)`.

## Why every prior fix failed on OrbStack

All four pursued **one** mechanism: a fresh procfs mounted in a new PID namespace via `unshare --pid --fork --mount --mount-proc` (the `real-proc` strategy). OrbStack's hardened VM kernel **rejects `mount("proc", ...)` from user namespaces even with `CAP_SYS_ADMIN`**. So on OrbStack:

- the container probe `canMountRealProc()` (`unshare … -- /bin/true`) returns **exit 1** ("Operation not permitted"),
- the runtime falls back to `--tmpfs /proc`,
- which is the **exact empty-`/proc` profile that crashes Bun**.

The real-proc path is dead on arrival there, so the fallback (the bug) always fires. Each "fix" was effectively a container restart that changed nothing on this host. Confirmed: the live container had only `CAP_NET_ADMIN`, and the `unshare --mount-proc` probe exits 1.

## Root cause

Bun's external-package runner spawns a child that reads `/proc/self/{fd,maps,exe}`. The empty `--tmpfs /proc` (even with the `/proc/self/exe` symlink) lacks `/proc/self/fd` and `/proc/self/maps` → the child aborts with `ENOTDIR`. The child needs a **real** `/proc` — but *not necessarily a freshly-mounted one*. The container already has a real procfs.

## Fix: a cap-free `proc-bind` strategy, made the default

`bwrap --unshare-all … --ro-bind /proc /proc` binds the container's **already-real** procfs straight into the sandbox — no `unshare --mount-proc`, no `CAP_SYS_ADMIN`. It works on every host, OrbStack included.

**The agent runtime's `/proc/<agent>/environ` (FIREWORKS_API_KEY, GH_TOKEN) is NOT leaked.** `--unshare-all` puts the sandboxed bash in a **child user namespace** (mapped-root, not real-root). The kernel's `PTRACE_MODE_READ_FSCREDS` check on `/proc/<pid>/environ` requires the reader's userns to be the target's userns or an ancestor — a child userns is neither, so the read fails `EACCES`. The same boundary blocks `kill()`/`ptrace` (`EPERM`, no `CAP_KILL` in the parent userns). The only residual is **non-secret** PID metadata (other pids' `cmdline`/`status` visible), accepted on the single-tenant boundary where the API-key surface is what matters and is provably blocked.

Because the no-leak property is a *kernel fact* rather than a typeclaw invariant, `canBindProcSafely()` **probes it before selecting proc-bind**: it plants a secret in a sentinel sibling spawned in the parent userns (modeling the real agent runtime), enters the exact proc-bind bwrap shape, and asserts the sentinel's `environ`/`maps` are unreadable and it is unsignalable while the sandbox's own `/proc/self/{fd,maps}` ARE readable. Any failure falls through to `--tmpfs /proc` — **fail-closed, never a silent leak**.

## Strategy order & the `CAP_SYS_ADMIN` default

```
real-proc   (only if sandbox.realProc set AND canMountRealProc())
  -> proc-bind   (the default; canBindProcSafely())
  -> tmpfs       (degraded; external packages unavailable)
```

Since `proc-bind` unblocks the core workflow **without any outer capability**, this also:

- flips `sandbox.realProc` to default **`false`**, and
- stops `planStart()` granting `--cap-add=SYS_ADMIN` by default — a **smaller outer blast radius** than the previous default.

`realProc: true` remains an opt-in for the stricter `real-proc` PID-isolation posture (pids absent, not just unreadable) at the cost of the cap; the resolver still falls back to `proc-bind` where the mount is a no-op. Oracle reviewed the strategy ordering and the userns leak-block reasoning and concurred `proc-bind` is the safer default.

## Verification (live, on the failing OrbStack container, no CAP_SYS_ADMIN)

Running the exact argv `build.ts` now emits:

| Property | Result |
| --- | --- |
| `bunx cowsay` (the crashing workload) | ✅ exit 0 |
| Planted sentinel secret in `/proc/N/environ` | ✅ `NO_LEAK_CONFIRMED` (all reads `Permission denied`) |
| Sandbox's own `/proc/self/{fd,maps}` | ✅ `SELF_PROC_OK` |
| Signal sentinel (`kill -0`) | ✅ `sentinel-unsignalable` |
| Container caps | only `CAP_NET_ADMIN` — no `CAP_SYS_ADMIN` |

`scripts/verify-procbind-sandbox.sh` codifies this as an operator-runnable acceptance check (runs WITHOUT `--cap-add`, unlike `verify-realproc-sandbox.sh`).

## Changes

- **New `proc-bind` branch + type** — `src/sandbox/{build,policy}.ts`
- **New `canBindProcSafely()` probe** — `src/sandbox/availability.ts`
- **Strategy resolution** in `applyBashSandbox` — `src/agent/plugin-tools.ts`, read from the boot-time `config` snapshot (not live `getConfig()`) to stay coherent with the boot-time cap decision; `sandbox` is `restart-required` for exactly this reason
- **`realProc` default `false`** — `src/config/config.ts`; SYS_ADMIN grant gated in `src/container/start.ts`
- **`scripts/verify-procbind-sandbox.sh`** — operator acceptance check (no `--cap-add`)
- **Tests** — proc-bind argv shape, probe contract (boolean/cache/dedup), `start.ts` default-no-cap
- **Docs** — `sandbox.mdx`, `typeclaw-json.mdx`, `AGENTS.md`

## Checks

`bun run typecheck` ✅ · `bun run lint` (0 errors) ✅ · `bun run format:check` ✅ · `bun run test` (7882 pass / 0 fail) ✅

## Lineage

Supersedes the OrbStack-broken default from #696 (`fix/sandbox-realproc-default`) and #697; builds on the diagnosis in #672 (`fix/sandbox-jsruntime-runners`) and #683 (`fix/sandbox-real-proc`). `real-proc` is retained as an opt-in; this PR removes it from the default path because it never functioned on OrbStack.

## Review notes

- Load-bearing security invariant: `proc-bind` is safe **only** because `--unshare-all` includes `--unshare-user`. Do not remove/relax it, and keep `canBindProcSafely()` gating selection — together they are the secret-leak guard. The inline comments in `build.ts` / `availability.ts` / `policy.ts` capture the exact kernel mechanism; please keep them.
- The `verify-procbind-sandbox.sh` argv mirrors `buildArgv()` — keep in sync if that helper changes.


devxoul:
## Self-review hardening (pushed `c0d76dd`)

An adversarial self-review (Oracle + live re-verification on OrbStack) found and fixed a **catastrophic false-pass** in the `canBindProcSafely()` probe before it could ship. Posting for reviewer visibility since it's security-critical.

### The bug: ESRCH/EACCES indistinguishability

The original probe asserted the sentinel's environ was unreadable with `cat /proc/<pid>/environ && exit 1`. That exits non-zero for **both**:
- `EACCES` ("Permission denied") — the real cross-userns block (**safe**), and
- `ESRCH` ("No such process") — the sentinel **died** (proves nothing).

The sentinel was `sleep 30`. On a slow/loaded host it could exit before the assertion ran, so **on a host that actually leaks**, the probe could return `true`, select `proc-bind`, and bind a `/proc` exposing the agent runtime's `FIREWORKS_API_KEY`/`GH_TOKEN`. Verified live that a dead pid and a properly-blocked pid produce the **identical** `cat` exit code — the false-pass was reachable.

### The fix: two-sided, errno-specific proof

- **Sentinel via `/usr/bin/env -i SECRET=leaked /bin/sleep 300`** — carries only the marker, never the real secrets, independent of `Bun.spawn` env semantics; 300s so it can't exit mid-probe.
- **Parent-side bracket**: assert the *parent* (container root, parent userns, same uid) CAN read the sentinel's environ → proves it's dumpable + same-uid, so the only thing that can deny the in-sandbox read is the userns boundary (rules out a false "blocked" from dumpable=0 / uid mismatch). Fail closed if the parent can't read it.
- **In-sandbox**: positively verify `/proc/self/{fd,maps}` readable + `/proc/<pid>/status` readable (alive → not ESRCH) + environ AND maps fail with **EACCES specifically** (`grep 'Permission denied'`).
- **Post-probe liveness re-check** from the parent.
- **Dropped `kill -0` and `readlink ns/user`** — the busybox `kill` builtin exits 1 with no stderr for both EPERM and ESRCH, and `readlink` prints nothing on EACCES; neither distinguishes safe from dead, so asserting them was theater.

### Verified live on OrbStack
- Safe host → probe passes → `proc-bind` selected → `bunx cowsay` exit 0
- Simulated leaking host (bwrap **without** userns → environ readable) → probe **fails** → falls back to `tmpfs` (no leak)
- Dead sentinel → fails at the status check

### Also (Oracle #3)
The degraded `tmpfs` fallback (reached only when **both** real-proc and proc-bind are unavailable — an exotic no-userns host) now warns once that external package runners will fail, instead of leaving only Bun's opaque `NotDir`. Deliberately **not** command-parsing (PR #696 rejected that as fragile) — a strategy-level, command-agnostic notice.

Full gate green: typecheck · lint (0 err) · format:check · test (7882 pass / 0 fail).

devxoul:
## Second self-review pass (pushed `9f6777e`)

Another adversarial review (Oracle + live empirical testing) of the *hardening itself*. Found two more issues that could silently reintroduce the crash — both fixed, plus the missing regression test.

### 1. Locale + PATH fragility in the probe (could silently re-break `bunx`)
The probe asserted the secret block by grepping `cat` stderr for the literal English `"Permission denied"`. That's a **localized diagnostic string, not an API** — on a host with a non-C locale installed, glibc translates it, `grep` misses, the probe returns false, and the resolver falls back to `tmpfs` → the original Bun `NotDir` crash returns **with no signal**. It also depended on `PATH` resolving `cat`/`grep` under `--clearenv`.

**Fix:** replace with a shell read-redirect open — `(: < /proc/<pid>/environ) 2>/dev/null && exit 1`. The shell itself does the `open(2)` (the same path Bun/an attacker uses); a cross-userns EACCES fails the redirect, a successful open (leak) runs `exit 1`. No external command, no error string → **locale- and PATH-independent**.

### 2. `access()` ≠ `open()` across userns (avoided a tempting wrong fix)
Empirically verified: `test -r /proc/<pid>/environ` reports **READABLE** for a same-uid sibling even when the ptrace-gated `open(2)` is actually **BLOCKED** — `access(2)` only checks permission bits + uid, not the `PTRACE_MODE_READ_FSCREDS` check that enforces the userns boundary. So `test -r` must **not** be used for the protected files (it would have made the probe mis-classify a safe host). Only an open attempt exercises the real leak path. Documented inline.

### 3. Stronger final liveness
Replaced `!sentinel.killed && sentinel.exitCode === null` (reflects only whether Bun *signalled* the child, not kernel liveness) with a parent-side re-read of the sentinel's environ — proves the pid still resolves to our live sentinel, so the in-sandbox open-failures were EACCES not a post-exit ESRCH.

### 4. Regression guard (the missing test)
Extracted the in-sandbox assertion as a pure `buildProcBindProbeScript(pid)` and added unit tests pinning its shape: must open (not `test -r`) the protected files, must prove the sentinel alive, and must **not** contain `grep` / `Permission denied` / the false-passing `cat … && exit 1`. The real probe needs a Linux container + bwrap (unrunnable in CI), so this shape test is the high-value guard against regressing to the leaking form.

### Verified live on OrbStack (exact generated script)
- Safe host → probe passes → proc-bind → `bunx cowsay` exit 0
- Leak sim (bwrap **without** userns → open succeeds) → probe fails → tmpfs
- Dead sentinel → fails at the status-alive check

Full gate: typecheck · lint (0 err) · format:check · test (**7886 pass / 0 fail**, +4 regression tests).

devxoul:
## Third self-review pass (pushed `8468d3a`)

Adversarial review (Oracle) + a whole-PR coherence audit (explore) of the iterated state. Found **one real bug** and several fail-closed hardening gaps — all fixed.

### 1. Cache not keyed by `bwrapPath` (real bug)
`canBindProcSafely({ bwrapPath })` accepted a path but cached a **singleton** result — a caller pinning a non-default bwrap (tests, or a future deployment) would inherit the default binary's "safe" answer **without probing its own**. Switched to a per-path `Map` cache + in-flight `Map`, matching `ensureBwrapAvailable`. Production always uses the default path, but the API shape made this a latent correctness/security bug. Added a regression test (a non-existent `bwrapPath` must probe its own binary → `false`, never inherit the default's cache).

### 2. Sentinel-death / PID-reuse race (fail-closed hardening)
The in-sandbox blocked-open assertions are only valid while the sentinel pid resolves to *our* live sentinel. Now `Promise.race` the bwrap probe against `sentinel.exited`: if the sentinel dies first, the open-failures could be ESRCH (pid gone) rather than the EACCES userns block → verdict void → **fail closed**. Makes the invariant explicit instead of relying on Bun's child-reaping timing.

### 3. Probe timeout
Added a 5s ceiling so a wedged runtime where `bwrap` hangs can never stall the first low-trust bash call indefinitely — time out → tmpfs.

### 4. Shorter sentinel
`sleep 300` → `sleep 30`: still vastly outlives the sub-second probe, but self-reaps in seconds if cleanup ever fails to `.kill()`.

### 5. `parentCanRead` via `Bun.file().bytes()`
Replaced the `cat` subprocess with a direct `open(2)+read` — the real leak path (matching the in-sandbox `(: < path)` check), no subprocess, no PATH dependence in the parent context.

### 6. Coherence fixes (stale comments from the 3 iterations)
- `resolveProcStrategy` comment said "prefer proc-bind … only fall to real-proc", but the code checks **real-proc first** (when opted in). Reworded to the actual branch order, noting proc-bind is the de-facto default because `realProc` defaults false.
- `verify-procbind-sandbox.sh` header claimed a third `kill -0` property the script never tests; dropped it (the signal boundary is a corollary of the userns isolation the leak scan already proves).

### Verified live on OrbStack
The real `canBindProcSafely()` (with all the above) runs against the container in **~4ms** and returns `true` → proc-bind selected.

Full gate: typecheck · lint (0 err) · format:check · test (**7887 pass / 0 fail**).

---

**Audit confirmed coherent** (explore): policy.ts, build.ts, config.ts, start.ts, sandbox.mdx, typeclaw-json.mdx, AGENTS.md all consistently describe proc-bind as default / realProc=false / no SYS_ADMIN-by-default. The two stale comments above were the only contradictions, now fixed.

devxoul:
## Self-review passes 4–6 (pushed `b957da9`, `ea43aa4b`)

Continued the review loop until a pass produced **zero actionable findings** (pass 6).

### Pass 4 — `b957da9` (caught a crash-class regression)
Testing the **actual** `canBindProcSafely()` imported from the module (not a hand-typed argv) revealed the probe's bwrap args were **missing `--dev /dev`**, so the in-sandbox `(: < path) 2>/dev/null` redirect failed → probe returned false → would have fallen back to tmpfs and **reintroduced the bunx crash**. Earlier "live verifications" used a hand-typed argv that happened to include `--dev`, masking it. Also fixed (Oracle): cache-poisoning via timeout/sentinel-death (now `{safe, cacheable}` — transient outcomes not cached), a `Bun.sleep(5000)` timer leak that hung the process ~5s on drain (→ `setTimeout`+`clearTimeout`), zombie cleanup (`SIGKILL` + `await exited`), and a marker-bytes check (PID-reuse + empty-buffer guard).

### Pass 5 — `ea43aa4b` (cache-poisoning, final variant)
The probe cached `exitCode !== 0` as a definitive `safe:false`. But — confirmed live — a real leak AND a transient setup/bwrap-startup failure **both exited 1**, so a transient failure got cached as permanent unsafe (silent bunx breakage until restart). Fixed with **distinct exit codes**: `0`=safe, `10`=definitive leak, `20`=setup-failed. Only `0` and `10` are cacheable; everything else (20, bwrap-startup, signals, 127) is inconclusive and retried.

### Pass 6 — CLEAN (zero findings)
Adversarial review (Oracle) + empirical testing of the dead/zombie-sentinel edge cases found **no bugs**. Confirmed:
- A dead/zombie sentinel → `test -r status` fails → exit `20` (inconclusive), never a false verdict.
- A dead sentinel's `environ` open **fails ENOENT** → cannot produce a false `exit 10` (false leak).
- The exit-0 path's `parentReadsSentinelMarker` re-check converts any mid-script death into inconclusive (not false-safe).
- `cmd && exit 10; exit 0` → dash's `exit 0` is unconditional.

Oracle's verdict: *"No actionable finding from this pass. The distinct exit codes plus 'only cache safe after marker re-check' close the prior ambiguous-exit/cache-poisoning class well."*

The probe has converged: every failure mode (timeout, sentinel death, zombie, PID reuse, setup failure, bwrap startup, locale, PATH, access-vs-open) is handled fail-closed or fail-inconclusive, and only definitive verdicts are cached.

Full gate on every pass: typecheck · lint (0 err) · format:check · test (**7887 pass / 0 fail**).

PR #710, opened 2026-06-08
feat(github-cli-auth): allow gh reader pipelines under token scoping

devxoul:
## Background

A repo-targeting `gh` command runs with a minted GitHub App **installation token** (repo-write-capable) in its `bash -c` process env. The guard therefore required a **single bare `gh`** and blocked every pipeline — including the high-frequency `gh api … | jq …` idiom. Each block forces a full agent round-trip to rewrite the command, which is slow and frequent.

## Summary

Allow a **trailing reader pipeline** when every downstream stage is a stdin-only allowlisted reader (`jq`, `cat`, `wc`, `sort`, `uniq`). Each downstream stage is rewritten to run under `/usr/bin/env -u GH_TOKEN`, so the token is absent from its environment; the leading `gh` stage keeps it.

    gh api repos/o/r | jq '.[] | {id}'
      → gh api repos/o/r | /usr/bin/env -u GH_TOKEN jq '.[] | {id}'

## Why this is safe (the token is repo-write-capable, so the bar is high)

The naive `env -u` is **not** sufficient on its own; the boundary is *"a downstream stage cannot open its own file/network/process sink."* Concretely:

- **No file operands on readers.** A reader handed a file argument could read the sibling `gh` process's `/proc/<pid>/environ` and recover the token. Readers are restricted to pure-stdin form.
- **Strict per-command flag allowlist (allow-known-good).** coreutils exposes file reads *and* code execution as **flags**, not just operands — `wc --files0-from=F`, `sort --files0-from=F`, `sort --compress-program=PROG`. Only known stdin-shaping flags are permitted; an unrecognized flag rejects the stage.
- **jq file-reading flags forbidden:** `-f`, `--from-file`, `--rawfile`, `--slurpfile`, `--argfile`, `-L`, `--run-tests`. jq's filter (a single positional) is the only allowed operand.
- **Backslashes rejected in reader stages.** The tokenizer does not model shell backslash escaping. A forbidden flag spelled with a leading backslash would look like a harmless positional to the parser while reaching bash as the real flag. Rejecting bare backslash characters in reader args closes that bypass.
- **Quote-aware pipe splitting.** A `|` inside a single-quoted jq filter is not a stage boundary. All other shell metachars, redirections, command/process substitution, `||`, `|&`, and unbalanced quotes still block.
- **Excluded readers:** `awk` (exposes `system()` and `/inet/` sockets), `sed` (GNU `e` flag), `tee`/`xargs` (write/spawn), `less` (shell escape via `!cmd`), and for now `grep`/`head`/`tail` (operand parsing too risky to allowlist yet).

## What this does NOT do

- Does not change the single-stage `gh` guard — that path is unaffected.
- Does not allow any non-`gh` leading stage or any write-capable reader.
- Does not enable `awk`, `sed`, `tee`, `xargs`, `less`, `grep`, `head`, or `tail` as pipeline readers.

## Review notes

The load-bearing change is `parsePipeline` + `validateReaderStage` in `gh-command.ts`. The invariant: every reader stage must pass `validateReaderStage` before `env -u GH_TOKEN` rewriting proceeds; any failure short-circuits to a block. The token-absent rewrite happens at the shell level, not via masking, so even a replaced binary at the reader path cannot observe the env.

## Tests

`gh-command.test.ts` gains a `reader pipelines` suite covering: allowed `gh | jq` (filter / multi-stage / quoted-filter), allowed stdin-shaping flags, and blocked cases — file operands, jq file-flags, coreutils file/exec flags (`--files0-from`, `--compress-program`), backslash-smuggled flags, non-allowlisted/exfil readers (`awk`/`sed`/`tee`/`xargs`/`less`/`grep`/`head`/`tail`), non-`gh` leading stage, output-file flags, and re-introduced metachars/redirections.

```
bun run test       (github-cli-auth: 165 pass)
bun run typecheck  ✓
bun run lint       ✓ (0 errors)
bun run format     ✓
```

### DIFF
--- a/src/bundled-plugins/github-cli-auth/gh-command.test.ts
+++ b/src/bundled-plugins/github-cli-auth/gh-command.test.ts
@@ -341,6 +341,31 @@ describe('analyzeGhCommand', () => {
       expect(analyzeGhCommand('gh api /repos/acme/widgets/issues | jq -L /tmp/mods .').kind).toBe('block')
     })
 
+    // Regression for PR #710 review: exact-token deny-listing missed jq's
+    // attached/clustered short-option forms, reopening the file-read path.
+    it('blocks attached and clustered jq short-option file/module flags', () => {
+      expect(analyzeGhCommand('gh api /repos/acme/widgets/issues | jq -f/proc/self/environ').kind).toBe('block')
+      expect(analyzeGhCommand('gh api /repos/acme/widgets/issues | jq -L/proc').kind).toBe('block')
+      expect(analyzeGhCommand('gh api /repos/acme/widgets/issues | jq -rf/proc/self/environ').kind).toBe('block')
+      expect(analyzeGhCommand('gh api /repos/acme/widgets/issues | jq -Lpath .').kind).toBe('block')
+    })
+
+    it('blocks unknown jq flags and jq filters that load modules', () => {
+      expect(analyzeGhCommand('gh api /repos/acme/widgets/issues | jq --run-tests').kind).toBe('block')
+      expect(analyzeGhCommand('gh api /repos/acme/widgets/issues | jq --args .').kind).toBe('block')
+      expect(analyzeGhCommand('gh api /repos/acme/widgets/issues | jq -Z .').kind).toBe('block')
+      expect(analyzeGhCommand('gh api /repos/acme/widgets/issues | jq \'import "m"; .\'').kind).toBe('block')
+      expect(analyzeGhCommand('gh api /repos/acme/widgets/issues | jq \'include "m"; .\'').kind).toBe('block')
+    })
+
+    it('allows jq safe boolean and value flags (clustered short, --arg, --indent)', () => {
+      expect(analyzeGhCommand('gh api /repos/acme/widgets/issues | jq -r .').kind).toBe('inject')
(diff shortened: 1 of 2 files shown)


devxoul:
Fixed in 6f7ad9fc. You were right — exact-token deny-listing missed jq's attached/clustered short-option forms (`-f/proc/self/environ`, `-L/proc`, `-rf/proc/...`), reopening the file-read path.

Switched jq to the same **allow-known-good** model the coreutils readers already use:
- **Short flags** are now decomposed char-by-char against a safe-boolean set (`r c s n e a S R j C M b`); any `f` (filter-from-file), `L` (module path), or unknown char rejects the whole stage — so `-f…`, `-L…`, and clustered `-rf…` all block regardless of attachment.
- **Long flags** matched as whole tokens against safe-boolean + safe-value sets (`--arg`/`--argjson` consume 2 tokens, `--indent` 1); jq's no-`=value` syntax means no `--flag=value` smuggling.
- Still exactly one positional (the filter); a second positional is an input file jq would open.
- Also closed a related vector: filters using `import`/`include` (module load from the default search path, no `-L` needed) are now rejected.

Added regression tests for attached/clustered/unknown short forms and the module-loading filters. All 168 plugin tests pass; typecheck/lint/format clean.

PR #728, opened 2026-06-08
feat(agent): surface configured channels in the system prompt

devxoul:
## Symptom

In a Slack DM, an operator asked the agent to set up its GitHub webhook/review config. The agent had **no idea it already runs a GitHub channel** watching specific repos with auto-review. Instead it invented a flow TypeClaw has no concept of — "forward GitHub webhooks to a **Slack channel**" — and kept asking for a webhook URL + destination channel. Its actual `channels.github` config (4 repos, `review.on: opened`, auto-approve, tunnel-registered webhook) was completely invisible to it.

## Root cause

The system prompt never describes the agent's **configured** channels. The origin block (`renderSessionOrigin`) only describes the **current** conversation ("you are in a Slack DM"); nothing tells the agent "you also run a GitHub channel watching repos X, Y, Z." `getConfig().channels` is available at prompt-compose time but was never read. Confirmed green-field — no existing prompt section, tool, or skill surfaces this.

## Fix

A new `renderChannelsBlock(channels)` renders a compact `## Channels` summary from `getConfig().channels`:

```
## Channels

These channels are configured for you (from `typeclaw.json`). This is your standing setup — the session origin above tells you which one THIS conversation is in.

- **github** — repos: owner/a, owner/b; review: on `opened`, may approve
- **slack-bot** — enabled

To change them … edit `typeclaw.json`; do not invent channels or destinations that are not listed here. This list (including repo names) is internal configuration: only disclose it to a current speaker with owner/trusted authority …
```

Wired into `composeSystemPrompt` between the origin/role block and gitNudge — the correct cache-suffix position (channels config is less volatile than git/memory, changes only on reload).

### Properties

- **Full mode only** (tui, channel). Slim sessions (cron, subagent) stay lean; the dump tests guard the <1000/<500 token budgets.
- **Gated empty** — returns `''` when no channel is enabled, so TUI-only agents pay zero bytes.
- **No secrets** — channel tokens/signing keys live in `secrets.json`/`.env`, never in the `channels` block; `webhookUrl` is deliberately omitted in case it embeds a token. (Self-review verified the schema has no secret-bearing fields.)
- **Disclosure guard** — channel sessions can carry low-privilege speakers, and repo slugs are internal metadata, so the block instructs the agent to decline reciting the list to a member/guest. (Self-review finding.)
- **Cache-stable** — zod `.parse` preserves schema-declared key order, so the rendered block is byte-stable for a given config.

## Commits

1. `feat(agent): add renderChannelsBlock …` — the pure renderer + its unit tests (lands self-contained, unused).
2. `feat(agent): render the configured-channels block …` — wires it into `composeSystemPrompt`, updates the debug dumper, pins ordering/breakdown.

## Verification

Self-reviewed (Oracle, adversarial) for cache position, credential leak, token budget, field-required choice, key-order stability, and subagent/cron paths — no blockers; the repo-slug disclosure concern was addressed with the guard line. Full gate green: `typecheck`, `lint` (0 errors), `format:check`, `test` (8038 pass / 0 fail).

`bun run debug:prompt --origin channel` shows the section in place; `--origin cron` confirms slim sessions omit it.

### DIFF
--- a/scripts/dump-system-prompt.test.ts
+++ b/scripts/dump-system-prompt.test.ts
@@ -61,6 +61,20 @@ describe('dumpSystemPrompt', () => {
     },
   )
 
+  test.each(fullKinds)('%s origin includes the configured-channels block', (kind) => {
+    const out = dumpSystemPrompt(kind)
+    expect(out).toContain('## Channels')
+    expect(out).toContain('**github**')
+  })
+
+  test('cron origin omits the configured-channels block (slim mode)', () => {
+    expect(dumpSystemPrompt('cron')).not.toContain('## Channels')
+  })
+
+  test('subagent origin omits the configured-channels block (override path)', () => {
+    expect(dumpSystemPrompt('subagent')).not.toContain('## Channels')
+  })
+
   test('cron origin omits the tmux shell-work guidance (slim base prompt)', () => {
     expect(dumpSystemPrompt('cron')).not.toContain('## Long-running and interactive shell work')
   })
@@ -178,6 +192,7 @@ describe('dumpSystemPrompt', () => {
       'Runtime block',
       'Session origin',
       'Role context',
+      'Channels',
       'Git nudge',
       'Memory (MEMORY.md + streams)',
     ])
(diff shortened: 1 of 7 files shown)


devxoul:
Closing in favor of a better approach. A static `## Channels` system-prompt block only helps when the referenced channel is already configured; it does nothing for the real failure case (an operator asking to ADD a github channel / repos that aren't configured yet). The actual gap is procedural: the agent doesn't know how to configure the github channel. Fixing that in the existing `typeclaw-config` skill instead, which already owns config-editing + reload knowledge but omits the github adapter from its Channels section.

PR #733, opened 2026-06-09
Mint GitHub App tokens for plain git commands

devxoul:
## Background

The `github-cli-auth` plugin already mints per-repo GitHub App installation tokens for `gh` commands (via `-R`/`--repo`). Plain `git` commands — `git push`, `git fetch`, `git clone`, etc. — have no `-R` flag and read credentials through `GIT_ASKPASS`, not `GH_TOKEN`. Without this PR, a `git push` to a private repo in a GitHub App-authenticated session either fails or falls back to whatever the container's default credentials are. There was no minting path for `git` at all.

## Summary

Four commits add end-to-end GitHub App token minting for plain `git` commands, parallel to the existing `gh` path.

**1. Repo resolver for plain git (`git-command.ts`, 391 lines).** `gh` names its target repo in argv; `git` implies it via a remote. The resolver walks: explicit `github.com` URL in argv → remote name via `git remote get-url` → the `branch.<cur>.pushRemote` / `remote.pushDefault` / `origin` fallback chain. Returns an `owner/name` slug or `pass-through` when unresolvable, so `git` fails honestly rather than us guessing a repo. Multi-owner commands (e.g. a clone list targeting two orgs) are blocked — a single minted token cannot authenticate all of them. Compound commands (`&&`, `;`, pipes, redirections, subshells) are blocked too, because the token arrives as an env overlay and any sibling process would inherit it. The one accepted prefix is `cd <path> && git …`, which is rewritten to `git -C <path> …` before execution.

**2. Token-free `GIT_ASKPASS` helper (`git-askpass.ts`, 49 lines).** `git` reads credentials from a `GIT_ASKPASS` script, not from `GH_TOKEN`. The helper emits `x-access-token` as username and the value of `TYPECLAW_GIT_TOKEN` (from env) as password. The script itself is constant and secret-free — it contains only the env var name. It is written once per process to `/usr/local/bin/typeclaw-git-askpass` (which is `--ro-bind` mounted into the bwrap sandbox) via a write-then-rename sequence so concurrent readers never see a partial file. `TYPECLAW_GIT_ASKPASS_PATH` overrides the default path for tests/CI.

**3. Plugin wiring (`index.ts`).** The `tool.before` hook is split into `handleGhCommand` and `handleGitCommand` helpers. `handleGitCommand` is gated on the App token class — classic PATs reach every owner, fine-grained PATs cannot be re-minted per repo, so both pass through unchanged. For App auth, the minted token is injected via `TYPECLAW_INTERNAL_BASH_ENV` (the existing overlay mechanism the bash wrapper delivers to the bwrap `--setenv`): `GIT_ASKPASS` points at the helper, `TYPECLAW_GIT_TOKEN` carries the minted token, `GIT_TERMINAL_PROMPT=0` prevents hang-on-prompt, and two `GIT_CONFIG_*` `insteadOf` entries rewrite SSH/scp remotes to `https://github.com/` so the credential applies regardless of remote URL scheme.

**4. Security ordering (`bundled-plugins.test.ts`).** Both the `gh` and `git` mint paths must fire after the security plugin's `tool.before` guards — a blocked exfil or tainted command must never reach the mint path. A focused assertion documents and enforces that ordering.

## What this does NOT do

- Does not affect classic PATs or fine-grained PATs — those pass through untouched for `git`, identical to how the `gh` path treats them.
- Does not change any other plugin, the bwrap sandbox config, or the security plugin itself.
- Does not mint tokens for non-remote git subcommands (`git status`, `git log`, etc.) — those return `pass-through` immediately.

## Review notes

**Load-bearing design decisions reviewers should verify:**

- The token never enters the command string or `git config`. It rides only in `TYPECLAW_GIT_TOKEN` env, delivered by the bash wrapper's env overlay. The askpass script is a constant that reads the env var; the token is absent from the script itself.
- The compound-command block (`COMPOSITION_REASON`) is the security fence. A token injected into env is visible to any sibling spawned in the same shell. The `cd … && git …` rewrite is the only permitted relaxation.
- The multi-owner block prevents one minted token (scoped to a single installation) from being attempted against a different owner's repo.
- The security-plugin ordering assertion in `src/run/bundled-plugins.test.ts` is the guard that keeps both mint paths downstream of the taint/exfil checks.

**Coverage:**

- `git-command.test.ts` — 194 lines covering URL parsing, remote resolution fallback chain, pass-through cases, inject cases, block cases, and the `cd`-rewrite path.
- `git-askpass.test.ts` — 46 lines covering write idempotency, concurrent-call coalescing, content correctness (no token in the script), and executable bit.
- `index.test.ts` additions — 121 lines of integration tests covering the full `tool.before` path for the git branch: env overlay shape, resolver receives correct slug, bridge-unavailable block, compound-command block, classic/fine-grained PAT pass-through, non-GitHub URL pass-through.

### DIFF
--- a/src/bundled-plugins/github-cli-auth/git-askpass.test.ts
+++ b/src/bundled-plugins/github-cli-auth/git-askpass.test.ts
@@ -44,3 +44,47 @@ describe('ensureGitAskPassHelper', () => {
     expect(mode).not.toBe(0)
   })
 })
+
+describe('ensureGitAskPassHelper — host-scoped behavior (executed)', () => {
+  async function run(promptArg: string): Promise<{ code: number; out: string }> {
+    const path = await tmpHelperPath()
+    await ensureGitAskPassHelper(path)
+    const proc = Bun.spawn({
+      cmd: [path, promptArg],
+      env: { TYPECLAW_GIT_TOKEN: 'ghs_secret_token' },
+      stdout: 'pipe',
+      stderr: 'ignore',
+    })
+    const code = await proc.exited
+    const out = (await new Response(proc.stdout).text()).trim()
+    return { code, out }
+  }
+
+  test('emits the token for a github.com password prompt', async () => {
+    const { code, out } = await run("Password for 'https://github.com': ")
+    expect(code).toBe(0)
+    expect(out).toBe('ghs_secret_token')
+  })
+
+  test('emits x-access-token for a github.com username prompt', async () => {
+    const { code, out } = await run("Username for 'https://github.com': ")
+    expect(code).toBe(0)
+    expect(out).toBe('x-access-token')
+  })
+
+  test('refuses (exit 1, no token) for a non-github host prompt', async () => {
+    const { code, out } = await run("Password for 'https://evil.example': ")
+    expect(code).toBe(1)
+    expect(out).toBe('')
+  })
+
+  test('is not fooled by a lookalike host (evil-github.com)', async () => {
+    const { out } = await run("Password for 'https://evil-github.com': ")
+    expect(out).toBe('')
+  })
+
+  test('is not fooled by a suffix lookalike host (github.com.evil)', async () => {
+    const { out } = await run("Password for 'https://github.com.evil.test': ")
+    expect(out).toBe('')
+  })
+})
--- a/src/bundled-plugins/github-cli-auth/git-askpass.test.ts
+++ b/src/bundled-plugins/github-cli-auth/git-askpass.test.ts
@@ -1,0 +1,46 @@
+import { afterEach, describe, expect, test } from 'bun:test'
+import { mkdtemp, readFile, stat } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+
+import { ensureGitAskPassHelper, resetGitAskPassHelperForTests } from './git-askpass'
+
+afterEach(() => {
+  resetGitAskPassHelperForTests()
+})
+
+async function tmpHelperPath() {
+  const dir = await mkdtemp(join(tmpdir(), 'typeclaw-askpass-'))
+  return join(dir, 'typeclaw-git-askpass')
+}
+
+describe('ensureGitAskPassHelper', () => {
+  test('writes the helper and returns its path', async () => {
+    const path = await tmpHelperPath()
+    expect(await ensureGitAskPassHelper(path)).toBe(path)
+    expect((await stat(path)).isFile()).toBe(true)
+  })
+
+  test('helper content contains no token, only the env var name', async () => {
+    const path = await tmpHelperPath()
+    await ensureGitAskPassHelper(path)
+    const content = await readFile(path, 'utf8')
+    expect(content).toContain('TYPECLAW_GIT_TOKEN')
+    expect(content).toContain('x-access-token')
+    expect(content).not.toMatch(/gh[ps]_/)
+  })
+
+  test('concurrent calls share one write and resolve to the same path', async () => {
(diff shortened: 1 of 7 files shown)


devxoul:
Both gaps fixed in b074f721:

1. **Push url vs fetch url.** `GitRemoteResolver` now takes a `forPush` flag, and `push` flows resolve `git remote get-url --push <remote>`. A remote with a distinct `pushurl` now mints against the actual push destination; fetch/pull/ls-remote still use the fetch url.

2. **Multi-remote.** Resolution now collects **every** named remote, not just the first, and feeds them all to the multi-owner guard. `git fetch --multiple origin upstream` across two owners now blocks; across one owner it injects. Later positionals on a non-`--multiple` command stay refspecs (`git push origin main` → remote `origin`, refspec `main`).

Added tests for distinct fetch/push urls (incl. asserting the `forPush` flag is passed per-subcommand) and for multi-remote fetch (block on cross-owner, inject on same-owner, refspec-not-remote). Full suite green.

devxoul:
Ran a self-review with two independent adversarial passes (security + correctness) and they surfaced a class of issues beyond the original two — pushed fixes in ced1950d and f2d9e15f.

**Token-redirection hardening (`git-command.ts`)** — a minted token rides in env, so any command that could re-point git's auth or destination is now blocked rather than minted for:
- Leading env assignment (`GIT_ASKPASS=/evil git …`) — blocked.
- User `-c`/`--config-env` (`-c url.https://evil/.insteadOf=…`, `-c core.askPass=…`, `credential.*`, `http.*`) — blocked.
- `--git-dir`/`--work-tree`/`--namespace`/`--exec-path` (git operates on a different repo than resolved) — blocked.
- `git fetch --all` / `git pull --all` (contacts every remote, can't enumerate them here) — blocked.
- Push-default fallback (`branch.pushRemote`/`pushDefault`/`origin`) is now **push-only**; bare `fetch`/`pull`/`ls-remote` with no remote pass through instead of minting for the push remote.
- `cd && git -C` (double `-C` would change cwd) and `cd ~` / `cd -` (shell expansion) — no longer mis-rewritten.
- Parser: `ssh://…:22/…` now parses; scp urls with `#`/`?` suffixes rejected; a throwing resolver fails safe to pass-through.

**Askpass hardening (`git-askpass.ts`)** — defense-in-depth at the credential edge:
- The helper now answers **only** prompts whose host is `github.com` (matched on `//github.com/` / `//github.com'`, immune to `evil-github.com` / `github.com.evil` lookalikes) and exits non-zero without printing the token otherwise. So even a redirect that slipped the analyzer can't hand the token to a non-github host.
- Helper written with an unpredictable name + exclusive (`wx`) create so a planted symlink can't redirect the write.

Added adversarial + correctness regression tests for every item above (incl. executed-script tests proving the askpass host scoping). Full suite green (8153 pass).

devxoul:
Good catch — fixed in 9b5a0e23.

`hasDangerousGitConfig` now also matches the inline `--config-env=<name>=<envvar>` form (`startsWith('--config-env=')`), not just the separate-arg form. I also added `--config-env`/`--namespace`/`--exec-path` to the value-flag set so their separate-arg form consumes its value and the real subcommand is still located (otherwise `git --config-env core.askPass=EVIL clone …` parsed `core.askPass=EVIL` as the "subcommand" and slipped through as a pass-through before reaching the guard).

Regression tests added for both forms:
- `git --config-env=core.askPass=EVIL clone https://github.com/acme/widgets.git` → block
- `git --config-env core.askPass=EVIL clone https://github.com/acme/widgets.git` → block

Verified against real git too; full suite green.

devxoul:
@typeey approve please

PR #737, opened 2026-06-09
feat(github-cli-auth): block and guide gh api /user under App auth

devxoul:
## Background

A PR-review agent running under GitHub App auth ran `gh api /user --jq '.login'` as a self-identity check. It got an error (`Resource not accessible by integration`, or in the bwrap sandbox a "set GH_TOKEN" message), misread it as "I have no auth at all", and aborted the review — even though a `gh pr view -R owner/repo` call had succeeded moments earlier. The agent posted a comment claiming it couldn't access the diff due to missing GitHub CLI auth.

## Summary

`gh api /user` resolves the **calling user's** identity. A GitHub App installation token is not a user identity — GitHub rejects `/user` on a **token-class** basis regardless of whether the token is valid. This is not an auth failure. It is not a regression of the `git` token-mint path added in #733; it predates that PR and is inherent to the per-repo App-token minting model.

**Root cause:** the `gh api` pass-through path in `index.ts` had no guard for the identity-model mismatch, so the App token was forwarded to an endpoint the token class can never satisfy.

**Fix:** two touch points in `src/bundled-plugins/github-cli-auth/`.

`gh-command.ts` — new exported pure helper `usesGhApiAuthenticatedUserEndpoint(command)`: detects `gh api /user`, `gh api user`, and any `gh api /user/...` descendant across every `gh` invocation in the command string. Narrow by design: `/users/{username}`, `/meta`, `/rate_limit`, `graphql`, and repo endpoints do not match. No token-class policy here — it is a pure shape detector.

`index.ts` — token-class-scoped hard block in `tool.before`, gated on `tokenClass === 'app' || tokenClass === 'none'` only. A classic or fine-grained PAT is a real user identity and works, so those pass through untouched. `none` covers the multi-owner App case where `GH_TOKEN` is intentionally unseeded. The block returns a guiding reason string that tells the agent this is **not** a missing-auth signal (repo-scoped calls still work) and points to `-R owner/repo` / `/repos/owner/repo/...` and the PR/issue/comment context for the actor identity.

The block lives in `index.ts`, not in the classifier. Moving it into the classifier would surface it before the fine-grained-PAT early-return and wrongly block `/user` for PAT users, who have a valid user identity.

## What this does NOT do

- Does not affect classic or fine-grained PAT auth — `/user` passes through for both.
- Does not match `/users/{username}`, `/meta`, `/rate_limit`, `graphql`, or any repo endpoint — only the authenticated-user family.
- Does not change the `git` token-mint path or any other plugin.
- Does not change behavior when the agent uses repo-scoped `gh` commands.

## Review notes

**Load-bearing invariant:** the `tokenClass` check must precede the `usesGhApiAuthenticatedUserEndpoint` call for correctness, and both must remain downstream of `analyzeGhCommand` (which handles compound-command blocks and the `pass-through` early-return that exits before the mint path). The existing `tool.before` ordering is preserved; this guard was inserted at the App-token decision point, after `analyzeGhCommand` returns and before `resolveToken` is called.

**Test coverage (`gh-command.test.ts`):** true cases — `gh api /user`, `gh api user`, `gh api /user/emails`, `gh api user/orgs`, with a `-H` header flag; false cases — `/users/octocat`, `/users/octocat/repos`, `/meta`, `/rate_limit`, `graphql`, repo endpoints, `echo gh api /user`. Plus a test asserting `analyzeGhCommand` still returns `pass-through` for `/user` (the block lives in the caller, not the classifier).

**Test coverage (`index.test.ts`):** App seeded token blocks `/user` with a guiding reason and does not call the resolver; multi-owner App (no seeded token) blocks; classic PAT passes through; fine-grained PAT passes through; App + `/users/octocat` is not blocked.

Checks: typecheck clean, lint 0 errors (66 pre-existing warnings, none in changed files), format:check clean, test 8163 pass / 0 fail.

### DIFF
--- a/src/bundled-plugins/github-cli-auth/gh-command.test.ts
+++ b/src/bundled-plugins/github-cli-auth/gh-command.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from 'bun:test'
 
-import { analyzeGhCommand } from './gh-command'
+import { analyzeGhCommand, usesGhApiAuthenticatedUserEndpoint } from './gh-command'
 
 describe('analyzeGhCommand', () => {
   it('passes through commands that do not invoke gh', () => {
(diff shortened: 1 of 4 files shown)


devxoul:
@typeey review again

devxoul:
Addressed in `a426ebca`.

The guard no longer keys off `process.env.GH_TOKEN`. It now resolves the **effective token per `gh api /user` invocation** and blocks only when *that* token is App / none-with-minter:

- New `effectiveGhTokensForAuthenticatedUserEndpoint(command, env)` in `gh-command.ts` walks the leading `VAR=val` assignments before each `gh` (reusing the existing quote-aware tokenizer + env-assignment detection) and applies gh's precedence: **command-local `GH_TOKEN` > process `GH_TOKEN` > command-local `GITHUB_TOKEN` > process `GITHUB_TOKEN`**.
- `index.ts` blocks iff some `/user` invocation's effective token would mint App auth, so a command-local PAT override (`GH_TOKEN=ghp_… gh api /user`) now passes through.
- Added `GITHUB_TOKEN` to the guard since gh reads it; `index.test.ts` now restores it in `afterEach`.

Regression coverage added: command-local `GH_TOKEN`/`GITHUB_TOKEN` PAT overrides (not blocked), command-local App token (blocked), quoted values, values containing `=`, the `GH_TOKEN`-beats-`GITHUB_TOKEN` precedence cases, and per-invocation resolution across a compound command. Full suite green.

PR #741, opened 2026-06-09
Add sandbox.symlinks: auto-create symlinks and allow their targets

devxoul:
> **Stacked on #739** (`sandbox.writablePaths`). Review/merge that first; this PR's base is `feat/sandbox-allow-symlink`, so the diff shown here is only the symlink layer.

## Why

#739 lets an operator mark a dir writable inside the bwrap sandbox. But the original report — `metabase-cli` failing with a read-only-filesystem error on `~/.metabase-cli` — needs **two** things: the symlink *created* AND its target *writable*. Hand-writing the symlink in the Dockerfile doesn't help, because inside the low-trust sandbox `$HOME=/tmp` (a per-session tmpfs), so `~/.metabase-cli` never resolves to the Dockerfile's `/root` symlink.

## What

`sandbox.symlinks: [{ from, to }]` does both in one entry:

```json
{ "sandbox": { "symlinks": [{ "from": "~/.metabase-cli", "to": "workspace/.metabase-cli" }] } }
```

- **`from`** = the symlink location, **fully configurable**: an absolute container path (`/root/.foo`) or a `~/`-prefixed path. (Per your request: source and destination are both configurable, not home-dir-only.)
- **`to`** = agent-root-relative target, automatically folded into the writable set (`getSandboxWritablePathSpecs`), so you never list it twice.

### Two stages, because `$HOME` differs

| Trust tier | bash runs | `$HOME` | who creates the symlink |
|---|---|---|---|
| trusted/owner | **unsandboxed** | `/root` | entrypoint shim `link_configured_symlinks` (boot) |
| guest/member | **inside bwrap** | `/tmp` | bwrap `--symlink` op from `applyBashSandbox` |

- **Entrypoint** (`src/init/dockerfile.ts`): `planStart` emits `TYPECLAW_SANDBOX_SYMLINKS` (base64 JSON, only when non-empty); the shim decodes + creates `from -> /agent/<to>` via `bun -e` (real JSON parser + fs, no shell-metachar hazard on arbitrary operator paths). Refuses to clobber an existing non-symlink; skips bad entries without failing boot.
- **bwrap** (`src/sandbox/build.ts`): `resolveSandboxSymlinks` resolves `from` against the **sandbox** HOME (`/tmp`), emits `--dir <parent>` + `--symlink /agent/<to> /tmp/<name>` after the `/tmp` bind (last-op-wins). Only symlinks whose `to` survived as a writable dir are emitted — a dropped/masked target never yields a dangling symlink onto an EROFS/hidden path.

### `from` security

Rejects null bytes, the filesystem root `/`, any path under `/agent` (self-loop), and the kernel/virtual roots `/proc`, `/sys`, `/dev`, `/run`. `/etc/...` is allowed (legit use case) — the entrypoint's no-clobber guard prevents overwriting an existing system file. `sandbox` stays `restart-required`; both consumers read the boot-time `config` snapshot.

## Commits (atomic, in dependency order)

1. Add `SandboxSymlinkOp` policy type
2. Add `resolveSandboxSymlinks` resolver
3. Add `sandbox.symlinks` config schema + `getSandboxWritablePathSpecs`
4. Emit `--dir`/`--symlink` in the bwrap builder
5. Create symlinks in the entrypoint for unsandboxed bash
6. Emit in-jail symlinks for low-trust sandboxed bash + docs

## Tests

`symlinks.test.ts` (resolver: `~/` → sandbox HOME, absolute verbatim, `to` resolution); `build.test.ts` (`--dir`+`--symlink`, ordering after `/tmp` bind, one op per entry); `config.test.ts` (schema accept/reject matrix, `getSandboxWritablePathSpecs` folding); `start.test.ts` (env emitted only when configured, base64-JSON round-trips); `dockerfile.test.ts` (shim defines + calls on both network paths; behavioral: creates `from -> /agent/<to>` with `~/` expanded against `$HOME`; refuses to clobber a real file).

`bun run typecheck`, `bun run lint` (0 errors), `bun run format`, and the full `bun run test` (8234 pass) all clean. Schema regenerated (`generate:schema`).


devxoul:
@typeey review please

devxoul:
Fixed in b4afcd80. @typeey good catch — the bans inspected the raw string while both consumers expand `~/` against `$HOME` at runtime, so `~/../agent/...` and `~/../proc/...` slipped through.

`symlinkFromSchema` now:
1. **rejects any `..` segment outright** — a traversal segment is the only way the post-`$HOME` effective path can re-enter a banned root, so this makes the raw-string bans equivalent to checking the effective path, with no need to expand `$HOME` at parse time;
2. runs the `/agent` and kernel-root bans on the **POSIX-normalized** form, so an absolute `from` like `/var/../proc` is caught too.

Regressions added for `~/../agent/workspace/.foo`, `~/../proc/x`, `~/../sys/x`, `/var/../proc/x`, and `/foo/../agent/...`, plus a positive case confirming legit `~/.foo` and `/etc/foo` still pass. (๑˃̵ᴗ˂̵)و

PR #751, opened 2026-06-09
fix(channels): make channel_disengage fire on explicit quiet commands

devxoul:
## Background

A channel bot (Slack) was told to stop replying in a thread. Instead of disengaging it answered with a normal acknowledgement ("ok, I'll be quiet") and kept its engagement. Inspecting the session transcript confirmed the root cause: the turn was a single `channel_reply` call and then ended; **`channel_disengage` was never invoked**, so the mid-conversation sticky credits were never cleared.

This is the mechanism that makes it a bug, not just a one-off: while engaged, every reply *re-grants* a fresh stickiness credit. So acknowledging a "be quiet" request with a normal reply actively renews the exact engagement the user asked to drop — the bot stays woken on the next participant message and gets pulled back into the thread. Earlier "stay quiet" requests in the same thread were met the same way (a reply, no disengage), which is why it took repeated, increasingly explicit stop commands that *also* failed.

The tool (`channel_disengage`) and a system-prompt section already existed. The gap was discoverability/framing: the model read the stop command as conversational and acked in text instead of treating it as a command to call the tool.

## Summary

Prompt-only fix (no router/keyword-matching change — kept the blast radius minimal):

- **System prompt (`session-origin.ts`):** add an explicit directive that an explicit quiet command — *in any language* — is a direct order to call `channel_disengage` that same turn, and that an ack alone does **not** disengage (it re-grants stickiness). Ack-first ordering is preserved: reply first if you want, then disengage.
- **Tool description (`channel-disengage.ts`):** lead with the trigger keywords across several languages so the model recognizes the command the moment it arrives, and state plainly that a "ok I'll be quiet" reply does not drop stickiness — only this tool does.

Why prompt-only and not deterministic router keyword-matching: the failure is the model not *reaching for* an existing, correct tool, so strengthening the signal it sees is the targeted fix. Hard-coded multilingual keyword detection in the router would be a larger, maintenance-heavy surface for a behavior the prompt can already drive.

Multi-language coverage is intentional (en/ko/ja/zh/fr/de/es examples) with an explicit "match intent, not exact words, whatever the language" instruction, since bots operate in mixed-language channels.

## What this does NOT do

- No change to router engagement/stickiness mechanics or `clearSticky`.
- No deterministic keyword auto-disengage — the model still decides; this only makes the decision obvious.
- No change to `skip_response` / `channel_reply` semantics.

## Review notes

- Load-bearing change: the new `**An explicit quiet command is a direct order to call this tool.**` block in `renderChannelOrigin` (`src/agent/session-origin.ts`). New regression test in `session-origin.test.ts` pins the "MUST call `channel_disengage`", the trigger keywords, and the "ack is not enough" wording.
- Existing ack-first-ordering test and `channel-disengage.test.ts` behavior are unchanged and still green.
- `bun run typecheck` / `lint` (0 errors; pre-existing `docs/` warnings only) / `format:check` / full `test` (8410 pass, 0 fail) all pass.


devxoul:
Good catch — fixed in 53ba8031. Dropped `static` (it collides with `static analysis` / `static typing` / `static method` in technical chats and would over-trigger `channel_disengage`) and replaced it with `тихо`, an unambiguous quiet command, to keep the Russian coverage without the English collision. The "match intent, not exact words" instruction still backstops phrasings not in the example list.

devxoul:
@typeey approve if there is no more blocking

Issue #761, opened 2026-06-09
doc-render: extend to docx/xlsx/pptx via the same dynamic-dep pattern


devxoul:
## Context

The bundled `doc-render` plugin (added alongside the `typeclaw-render-pdf` skill) renders **PDF only** in v1, via `@myriaddreamin/typst-ts-node-compiler` installed on first use (`bun add` into the agent's `node_modules`), with a bundled render script and first-party guided-error on a missing dep.

The architecture was deliberately scoped so additional document formats can slot in by the **same pattern** without baking anything into the base image.

## Proposed extension

Add per-format renderers + `typeclaw-render-<ext>` skills:

- **docx** authoring → `docx` (npm) — fits the dynamic `bun add` + dynamic-import pattern cleanly.
- **xlsx** authoring → `exceljs` (npm) — same.
- **pptx** authoring → `pptxgenjs` (npm) — same.

Each would mirror `render.ts`: probe the dep, guided-error on absence, render in-process via dynamic `import()`.

## Known boundary (do NOT fold in naively)

Office **conversion** (md→docx, docx→pdf) and faithful round-tripping realistically need a **system binary** (LibreOffice headless / Pandoc), **not** an npm package. That breaks the no-bake promise (it'd force baking the binary into the base image, the thing this design avoids). Conversion is out of scope for this pattern and should be evaluated separately.

## Non-goal (unchanged)

Reading/parsing/extracting from existing documents stays out of `doc-render` — it produces documents, it doesn't ingest them. (pypdf/qpdf for existing PDFs.)

See `src/bundled-plugins/doc-render/render.ts` (`TODO(doc-render)`) for the in-code marker.

PR #762, opened 2026-06-09
feat(doc-render): bundle a PDF render plugin with a lazily-installed Typst toolchain

devxoul:
## What & why

Replaces the standalone `typeclaw-markdown-pdf` skill with a bundled **`doc-render`** plugin. The motivation came from auditing real Markdown→PDF failures:

- A skill-authored `workspace/.tools/` toolchain (npm install + render script written into user-facing scratch) — arbitrary, unstructured, and cluttering the directory users look at for deliverables.
- Agents flailing on a Bun `NotDir`/exit-134 sandbox `/proc` failure (since fixed in 0.35.1) and nearly reaching for forbidden fallbacks (ReportLab).
- The styled `.typ` wrapper hand-retyped every render, so output drifted in fonts/margins.

The fix is a plugin that owns the render logic as first-party code, installs the heavy compiler **dynamically** (no image bloat), and gives **precise guidance on failure** because it's first-party.

## Design

- **Bundled plugin `doc-render`** contributes the `typeclaw-render-pdf` skill (via `skillsDirs`) and ships a `render.ts` on disk. It publishes that script's path to `/tmp/typeclaw-doc-render-script` at boot, so the skill can `bun run` it without hardcoding a `node_modules` path — the same hint-file contract `agent-browser` uses for its proxy port.
- **No registered tool.** Rendering stays a lazily-loaded skill + `bun run` script, so it costs **zero always-on system-prompt context**.
- **Dynamic dep, no bake.** The Typst compiler (`@myriaddreamin/typst-ts-node-compiler@0.7.0`) is installed on first use via agent-driven `bun add` into the agent's own `/agent/node_modules`. `render.ts` dynamic-imports it, resolving against **cwd** (verified necessary: bare `import()` resolves relative to the bundled script under `node_modules/typeclaw/...`, which need not see the agent's deps).
- **First-party guided error.** On a missing dep, `render.ts` exits with the exact pinned `bun add` line and forbids jsPDF/pdfkit/ReportLab. It also recognises the `NotDir`/proc degraded mode and tells the agent not to switch libraries over it.
- **Creativity preserved.** The `.typ` wrapper is framed as a per-document starting point, not a fixed template.
- **CJK.** Detect tofu after render → offer `cjkFonts` + restart, rather than a pre-render gate.

## Removed / repointed

- Deleted `src/skills/typeclaw-markdown-pdf/`.
- Repointed report-delivery guidance (`system-prompt.ts`, `session-origin.ts`) and their tests to `typeclaw-render-pdf`.

## Tests

New `render.test.ts` (helper routing + CLI usage/exit codes) and `index.test.ts` (plugin contributes skillsDirs, writes the hint, skill content invariants). Full suite green: 8444 pass, 0 fail. Manually verified end-to-end: missing-compiler → exit 3 with guidance; installed → `wrote report.pdf`, valid `%PDF-1.7`.

## Scope

PDF only in v1. Future formats (docx/xlsx/pptx) and the npm-vs-system-binary boundary for Office *conversion* are tracked in #761. Reading/parsing existing documents stays a non-goal (pypdf/qpdf).

Relates to #761.


devxoul:
@typeey can you approve

PR #764, opened 2026-06-09
fix(inspect): show the timestamp once per turn, not per event

devxoul:
## Background

`typeclaw inspect` stamped a `HH:MM:SS` on **every** rendered event — `user`,
each `thinking` block, both `tool` start/end lines, `assistant`, and `done`.
On a real turn that means a wall of timestamps: a run of three consecutive
thinking blocks printed three near-identical times, and every tool call carried
a timestamp no one reads. The noise is worst in the live TUI transcript view,
where each component (including back-to-back thinking) got its own time prefix.

The redundancy is structural: within a turn, most consecutive events share the
same category, so their timestamps add no information over the first one.

## Summary

Introduce a small `TimeGate` (in `src/inspect/render.ts`) that emits the
timestamp only when an event's category differs from the previous line. A run
of same-category events (consecutive thinking blocks, a tool start/end pair)
now carries a single leading stamp; a new role marks a new stamp.

Both renderers thread the same gate:

- **Line view** (`index.ts`) passes `showTime` into `renderEvent`. When false,
  the time column is **blanked to the same width** rather than dropped, so the
  tag column stays aligned.
- **TUI transcript view** (`transcript-view.ts`) drops the time `Text`
  component entirely for suppressed entries (vertical stack, no alignment
  concern).

Why category-change as the boundary (and not "only on user messages" or "every
event"): it matches how a reader scans the transcript — one timestamp per
visible role switch (user → assistant/think → tool → assistant), which is the
turn structure the report asked for, while collapsing the repetition that
prompted this.

## Preview

```
=== BEFORE (timestamp on every event) ===
11:07:39  user       curate Model X accessories
11:07:40  think      Researcher is running in background. I should end my turn.
11:07:41  think      Re-reading: the system will not surface the subagent result on its own.
11:07:42  think      I already sent my ack reply. Wait for completion.
11:07:43  tool ▸     spawn_subagent({"subagent_type":"researcher"})
11:07:43  tool ◂     spawn_subagent → ok "Spawned researcher. task_id=bg_c51b" (15ms)
11:07:45  assist     Researcher is running in the background; I will reply when it completes.
11:07:45  done       tokens: 100 in / 50 out · $0.0012

=== AFTER (timestamp once per same-category run) ===
11:07:39  user       curate Model X accessories
11:07:40  think      Researcher is running in background. I should end my turn.
          think      Re-reading: the system will not surface the subagent result on its own.
          think      I already sent my ack reply. Wait for completion.
11:07:43  tool ▸     spawn_subagent({"subagent_type":"researcher"})
          tool ◂     spawn_subagent → ok "Spawned researcher. task_id=bg_c51b" (15ms)
11:07:45  assist     Researcher is running in the background; I will reply when it completes.
11:07:45  done       tokens: 100 in / 50 out · $0.0012
```

## What this does NOT do

- Does not change the JSON output path (`--json` still emits the full `ts` per
  event).
- Does not change event ordering, filtering, `--since`, or the header/footer.
- Does not merge categories: `assistant`, `thinking`, and `done` remain
  distinct, so an assistant turn can still show a few stamps at its role
  switches — only *consecutive same-category* runs collapse.

## Review notes

- Load-bearing change: `TimeGate.shouldShow(cat)` in `render.ts`. Both render
  paths share one instance per stream so suppression is consistent.
- The line view blanks the column (`' '.repeat(TIME_WIDTH)`) instead of
  dropping it — verify alignment holds with the `showTime=false` test in
  `render.test.ts`.
- `index.test.ts` asserts the end-to-end pattern `[true,true,false,true,false]`
  for a meta/user/divider/tool/tool sequence; the transcript-view test asserts
  the stamp appears exactly once across two consecutive thinking blocks (and
  fails on the pre-change renderer).

### DIFF
--- a/src/inspect/index.test.ts
+++ b/src/inspect/index.test.ts
@@ -82,6 +82,7 @@ const ID_PICK1 = '019dda40-3333-7000-9000-000000003333'
 const ID_PICK2 = '019dda40-4444-7000-9000-000000004444'
 const ID_AMB_1 = '019dda40-5555-7000-9000-000000000001'
 const ID_AMB_2 = '019dda40-5555-7000-9000-000000000002'
+const ID_RESETBND = '019dda40-6666-7000-9000-000000006666'
 const AMB_PREFIX = '019dda40-5555'
 
 describe('runInspect — direct session id (replay-then-exit)', () => {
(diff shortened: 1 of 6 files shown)


devxoul:
Thanks @typeey — both edge cases addressed.

**1. Gate not reset at the live divider** (`84a546f4`)
Added `TimeGate.reset()` and call it when the live divider is printed, in both the line renderer (`index.ts`) and the TUI view (`transcript-view.ts`). The first live row is now always stamped, even when it shares the last replayed event's category. Covered by a line-view test (replay ends on a `tool`, live yields a `tool`, assert the live row carries `HH:MM:SS`) and a `TimeGate.reset()` unit test.

**2. Suppressed entries outliving the evicted stamp in the bounded TUI history** (`0a7ea1ce`)
Went with the in-place approach (your second suggestion): each history entry now carries `cat`/`ts`/`stamped` + its (possibly blank) timestamp `Text`. On eviction, if the evicted entry was the stamped head of a run that continues into the visible window, the new first-visible row is promoted by filling its already-positioned `Text` via `setText` — no child add/remove/reorder. Made the history cap injectable (`maxHistoryEntries`) so the eviction path is testable without 250 events; the new test overflows a cap-2 window with a same-category run and asserts the surviving visible head is stamped.

Both fixes are split into their own commits. `typecheck`/`lint`/`format`/`test` all green.

devxoul:
@typeey so approve?

PR #798, opened 2026-06-11
Route undecryptable LINE messages instead of dropping them as empty_text

devxoul:
## Background

Inbound LINE messages from a real DM chat were silently dropped: [line] inbound ... content_type=NONE text_len=0 attachments=0, then [line] dropped ... reason=empty_text. Investigation against the live session proved the chat uses Letter Sealing E2EE and this auth-token session has no E2EE self-key. Empty key storage makes the SDK return missing_e2ee_key for every message. The SDK version shipped in current main surfaces these events as text: null, content_type: NONE, with decryption_error set, which is wire-identical to a genuinely empty message. The LINE adapter had no decryption_error handling, so every undecryptable message was dropped as empty_text with no distinct log, making the agent look dead.

## Summary

When event.decryption_error is present, splitInboundLine now synthesizes a stable, agent-readable placeholder before the NONE short-circuit, for example [LINE message could not be decrypted: missing E2EE key]. This lets the message route instead of dropping. The adapter also logs a distinct [line] undecryptable ... warning for operator diagnosis. Codes are mapped to safe phrases rather than forwarding the SDK's verbose or localized message verbatim.

## What this does NOT do

- Does not make undecryptable messages readable. That requires provisioning an E2EE self-key via QR re-auth, which an auth-token session cannot do. This change only ensures such messages are surfaced with a placeholder instead of silently lost.
- Genuinely-empty NONE messages without decryption_error or attachments still drop as empty_text, unchanged.
- No dependency bump. Main already ships agent-messenger 2.20.1.

## Review notes

- Load-bearing: the decryption_error check in splitInboundLine must come before the content_type === NONE short-circuit. Otherwise the empty-string synthesis re-triggers empty_text.
- Codes are mapped via a small table. Unknown codes fall back to a generic phrase.
- Verified: 34 LINE adapter tests, full channels suite with 1855 tests, and full repo suite with 8679 tests pass. Typecheck, lint, and format are clean. New tests cover splitter placeholder, classifier routing, and an end-to-end adapter test asserting the warning log and placeholder route.

### DIFF
--- a/src/channels/adapters/line-attachment.test.ts
+++ b/src/channels/adapters/line-attachment.test.ts
@@ -51,6 +51,18 @@ describe('splitInboundLine', () => {
     expect(result).toEqual({ text: '', attachments: [] })
   })
 
+  test('routes an undecryptable message with a placeholder instead of dropping it', () => {
+    const missingKey = splitInboundLine(
+      event({ content_type: 'NONE', text: null, decryption_error: { code: 'missing_e2ee_key', message: 'no key' } }),
+    )
+    expect(missingKey).toEqual({ text: '[LINE message could not be decrypted: missing E2EE key]', attachments: [] })
+
+    const failed = splitInboundLine(
+      event({ content_type: 'NONE', text: null, decryption_error: { code: 'decrypt_failed', message: 'boom' } }),
+    )
+    expect(failed).toEqual({ text: '[LINE message could not be decrypted]', attachments: [] })
+  })
+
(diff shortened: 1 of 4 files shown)


devxoul:
Closing as not useful. A placeholder for undecryptable messages is surface-level noise — the real fix is provisioning the E2EE self-key via QR re-auth (typeclaw channel reauth line), which makes messages actually readable. Routing an unreadable placeholder adds no value over the empty_text drop.

PR #799, opened 2026-06-11
Write LINE E2EE keys to the agent workspace on host login

devxoul:
## Background

Inbound LINE messages from a Letter-Sealing DM chat were dropped as `empty_text` even after a successful QR re-auth. Live inspection confirmed the root cause: the SDK persists the E2EE self-key under AGENT_MESSENGER_CONFIG_DIR/line-storage, while the host-stage LINE login did not set that env.

The container already sets the env to `/agent/workspace/.agent-messenger` via src/init/dockerfile.ts. But typeclaw init and typeclaw channel reauth line run QR login on the host, so the SDK wrote the key to its default user config directory instead. The container then read its own empty line-storage, so inbound Letter-Sealing messages arrived undecryptable with missing_e2ee_key, text null, and content type NONE, then were dropped as empty_text.

The auth credential persisted fine; only the E2EE key landed in the wrong directory. That is why reauth appeared to succeed while decryption stayed broken.

## Summary

Default `AGENT_MESSENGER_CONFIG_DIR` at the start of runLineBootstrap to the per-agent workspace config dir. This mirrors the existing kakaotalkConfigDir pattern and makes host-stage QR login write LINE storage to the same directory the container reads.

The default uses nullish assignment so an already-set value is left untouched. The container stage is unaffected because the Dockerfile already provides the env; only the host login path changes. A new exported lineConfigDir(agentDir) helper names the path.

## What this does NOT do

- Does not change the container runtime; it already reads the correct dir via the Dockerfile env.
- Does not recover past undecryptable messages; only messages received after a key is present in the correct dir decrypt.
- Does not change dependencies.

## Review notes

- Load-bearing: use nullish assignment, not plain assignment, so the container's Dockerfile-provided value is never overridden.
- Mirrors the existing kakaotalkConfigDir pattern of using the per-agent workspace config dir.
- Verified live: copying the existing key from the default dir into the agent workspace and restarting made all messages in the affected Letter-Sealing DM chat decrypt with zero errors. This fix makes reauth write the key there automatically, without a manual copy.
- TDD: a new test fails without the env default and passes with it; a second test asserts an already-set value is not overridden. Full suite, typecheck, lint, and format pass.

### DIFF
--- a/src/init/line-auth.test.ts
+++ b/src/init/line-auth.test.ts
@@ -7,7 +7,7 @@ import type { LineLoginResult } from 'agent-messenger/line'
 
 import { SecretsLineCredentialStore } from '@/secrets/line-store'
 
-import { runLineBootstrap, type LineLoginClient } from './line-auth'
+import { lineConfigDir, runLineBootstrap, type LineLoginClient } from './line-auth'
 
 async function withDir<T>(fn: (dir: string) => Promise<T>): Promise<T> {
   const root = await mkdtemp(join(tmpdir(), 'typeclaw-line-auth-'))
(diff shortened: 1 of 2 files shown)


devxoul:
Addressed in 2376f819. You're right — the `??=` left `AGENT_MESSENGER_CONFIG_DIR` set for the process lifetime, so a second bootstrap with a different `agentDir` would inherit the first's path.

Fixed by scoping the env to the bootstrap operation: a new `withLineConfigDir` helper sets it only around client construction + login (when the SDK reads it to locate `line-storage`) and restores it in `finally`, mirroring the existing `suppressLineTokenInfoDump` pattern. An already-set value (the container's Dockerfile env) is still left untouched.

Added two tests for the edge case you flagged:
- a second bootstrap with a different `agentDir` uses its own path (not the first's) — verified to fail against the old sticky behavior;
- the env is restored to undefined after a host login.

Full suite (8680) + typecheck + lint + format all green.

PR #816, opened 2026-06-12
feat(memory/vector): drop stream-fragment embeddings; index topics only

devxoul:
> **Stacked on #811** (`feat/memory-parent-child-retrieval`). Review/merge #811 first; this PR is based on that branch so the diff is scoped to just this change.
>
> **Note:** #811 has since merged, so the base is now `main` and all 5 commits are visible here.

## Background

With parent-child retrieval (#811), a query matches fine-grained fragments but the returned unit is always the parent topic. That made the per-fragment **vector** embedding redundant: a stream fragment's embedding only added semantic-paraphrase recall during the ~30-minute window before dreaming consolidates it into a topic — and it added that on the **write hot path** (`makeAppendHook` embedded every fragment as memory-logger appended it).

The keyword lane already covers undreamed fragments: `keywordLane` → `searchAll` reads the daily-stream JSONL directly (no vectors) and matches fragment topic/body by substring/token. After #811, a keyword hit on an undreamed fragment self-resolves to a `source:"stream"` result, preserving the freshness window. Once dreaming consolidates the fragment, its parent topic's vector carries durable semantic recall.

So the only thing stream vectors bought — semantic-paraphrase recall on <30-min-old fragments — was a thin benefit for a hot-path cost, on content the keyword lane already reaches.

## What this does

Embed **topics only**. Stream fragments are never written to the vector index.

- **`passages.ts`** — `buildPassages` drops the stream branch; `collectPassages` loads shards only. `Passage.source` narrows to `'topic'`.
- **`index.ts`** — removes the `appendVectorStore` open and the `onFragmentsAppended: makeAppendHook(...)` wiring, so no per-fragment embedding fires on append.
- **`index-on-write.ts`** — deleted (`makeAppendHook` is now unused).
- **`dreaming.ts`** — removes `deleteStreamVectorsForDroppedFragments` (and its call); with no stream vectors, there is nothing to delete on compaction GC.

The keyword lane still yields `stream` results, so the `source: 'topic' | 'stream'` discriminator on `HybridSearchResult` / search results is kept — it just never originates from a vector row anymore.

## Migration

None needed. Vector memory is unreleased (`memory.vector.enabled` defaults to `false`). Any leftover `stream:*` rows in an existing index are pruned by the startup index build's stale-row prune (added in #811) on the next boot — they fall out of `collectPassages` and `pruneStaleRows` deletes them. The doctor also flags/prunes them.

## What is lost

Only **semantic-paraphrase recall on undreamed (<~30-min-old) fragments** — a query that paraphrases a just-logged fragment with zero keyword overlap, fired before the next dreaming run. After consolidation the topic vector covers it. Exact/substring/token matches on fresh fragments are unaffected (keyword lane). The non-vector default path (`memory_search` keyword only) is entirely unaffected.

## Review notes

- **Boot side effect removed:** opening `appendVectorStore` at plugin boot used to create the index DB as a side effect. It no longer does, so the empty-agent doctor test now builds the startup index explicitly (matching the real boot order in `run/index.ts`).
- **Tests:** `index-on-write.test.ts` deleted (covered removed code); `startup.test.ts` asserts topic-only embedding + legacy-stream-row prune; dreaming QA 2.4 (stream-vector deletion) removed, QA 2.5 (`droppedFragmentIds` reporting) kept.

## Verification

- `bun run typecheck` clean
- `bun run lint` 0 errors
- `bun run format:check` clean
- `bun run test` — 8767 pass / 0 fail / 1 skip


devxoul:
Closing — this PR over-cut the design.

#811 established parent-child retrieval: "match fragments (vector OR keyword), collapse to the parent topic." This PR removed fragment **embeddings**, leaving fragments matchable by **keyword only**. That defeats the pattern's purpose:

- The vector index became topic-only, so the only semantic surface is the **dreamed belief prose** (`heading + body`), not the **specific wording of the cited evidence**.
- A paraphrase query that semantically matches a fragment's detail (e.g. "fast package manager" → fragment "switched to pnpm for faster installs") matches the **child embedding** but not the summarized parent ("uses pnpm"), and BM25/keyword can't bridge the paraphrase.

This is the canonical MultiVectorRetriever / RAPTOR result: "match children, return parent" **requires child embeddings** — RAPTOR's ablation shows removing any layer's embeddings degrades detail recall.

The defensible cut was narrower (only *undreamed/uncited* transient fragments, not *cited* evidence hooks). Reworking: restore fragment embeddings so the parent-child match surface is whole, and revisit the freshness-window trim later as a measured optimization.

PR #821, opened 2026-06-12
feat(agent-browser): drop dashboard compatibility proxy for native proxied origin

devxoul:
## Background

TypeClaw shipped a homegrown reverse proxy (`dashboard-proxy`), JS monkey-patch shim port rewriting, and CORS rewriter so the agent-browser observability dashboard — which bound a hardcoded loopback port — would be reachable over Tailscale, LAN, and forwarded origins. That mechanism was necessary because the old dashboard used absolute `http://127.0.0.1:<session-port>/…` URLs for every API call, making it unreachable from any origin other than the host itself.

agent-browser 0.27.0 (vercel-labs/agent-browser#1111) changes the architecture: the frontend now uses relative URLs, and the dashboard server proxies per-session HTTP and WebSocket through its own single origin at `/api/session/<port>/{tabs,status,stream}`. It binds `127.0.0.1:4848` (loopback) and validates connections by Host-header match, with forwarded origins included. One port, one origin, no CORS workarounds needed.

That makes the entire compatibility layer dead code.

## Summary

This PR removes the compatibility proxy and replaces its host-reachability role with a minimal new primitive: a **reserved host-forward** in the port broker.

**Commit 1 — `feat(portbroker): add reserved host-forward primitive`** (additive, back-compatible):

- Protocol: new `port-forward-request` message (ContainerToHostd) carrying `targetPort`, `hostCandidates`, and optional `reason`; `hostPort` added to `port-forward-opened` so a forward's host port can differ from its container target port.
- `forward-request-bus`: new in-process bus mirroring `forward-result-bus`, so plugin code can request a forward without holding a `ContainerBroker` reference.
- `container-server`: stores reserved requests and re-emits them after `broker-hello-ack` so they survive a broker WS reconnect.
- `hostd-client`: binds the first free host candidate (4848–4857 in order) targeting `127.0.0.1:targetPort`; suppresses the auto-watcher 1:1 forward for a reserved target; keeps the listener across target close and reconnect; re-keys forwarders by host port.
- `portbroker-manager`: serves `hostPort ?? port` on Tailscale so the broker-managed URL reflects the actual bound port.
- `run/index.ts`: wires the request bus into `createContainerBroker`.

Ordinary 1:1 auto-forwards (where `hostPort === port`) are unchanged.

**Commit 2 — `feat(agent-browser): drop dashboard compatibility proxy for native proxied origin`**:

- Plugin requests a reserved forward for port 4848 at boot and writes the won host port to the hint file. On broker-disabled or all-candidates-busy, writes a clear diagnostic instead of a confident but dead URL.
- Shim trimmed to User-Agent injection only — anti-fingerprint UA behavior is unchanged; dashboard `--port` rewriting and port-hint logic removed.
- `dashboard-proxy.ts`, `dashboard-discovery.ts`, and the now-unused `bind-with-forward.ts` (plus their re-exports and tests) deleted.
- `agent-browser@^0.27.0` pinned in the Dockerfile (Layer 4). Layer 4.5 headed-mode wrapper for upstream #1083 is intentionally untouched — unrelated.
- `SKILL.md` and docs updated for the single-origin model; the hint file now carries the host port or a diagnostic, never a per-session port.

Net diff: +553 / −1790.

## Preview

No visible change for the common single-container case: `agent-browser dashboard start` → `http://localhost:4848` opens as before.

For multiple containers on one host, the reserved forward falls back through 4849–4857 and the hint file reflects the won port. When the broker is disabled or all candidates are busy, the hint file carries a diagnostic with a clear explanation rather than silently surfacing a dead URL.

## What this does NOT do

- Does not touch the headed-mode wrapper (Layer 4.5 Xvfb shim, upstream #1083) — unrelated.
- Does not change ordinary 1:1 auto-forward behavior; `hostPort === port` paths are unchanged.
- Does not bump the version. The new `port-forward-request` message is plugin-facing protocol surface and is a **minor bump candidate at release time** — that decision belongs in the release workflow, not here.
- Does not remove the hint file contract — the path is unchanged; only the content changes (host port or diagnostic, never a per-session port).

## Review notes

- **Load-bearing in commit 1:** the reconnect-survival path in `container-server.ts` (re-emits reserved requests after `broker-hello-ack`) and the `hostPort`-keyed forwarder map in `hostd-client.ts`. Verify: a broker WS reconnect must not drop the reserved listener, and the Tailscale-served URL must use `hostPort` (not `port`) when they differ.
- **Auto-watcher suppression:** confirm the guard in `hostd-client.ts` prevents a second listener from racing the reserved one on the same target port.
- **New ContainerToHostd message:** reviewers unfamiliar with the union should verify the exhaustive `switch` in `hostd-client.ts` handles `port-forward-request`.
- **Hint-file behavior change:** old shim wrote `http://localhost:<session-port>` (a per-session URL that expired); new plugin writes a static host port or a diagnostic. External tooling reading the hint file should treat an `http://` prefix as the legacy format and a bare number as the new format.
- Automated suite: 8746 tests, 0 fail. typecheck / lint / format:check all pass.


devxoul:
Thanks @typeey — both correct. Addressed in 18c3944a.

**1. Tailscale serve leak on close.** Added `hostPort` to the `port-forward-closed` event and emit `fwd.hostPort` at both close sites (`removeForwarder`, `teardownAllForwarders`); `portbroker-manager.ts` now calls `tailscale.stopPort(event.hostPort ?? event.port)`. For a reserved forward bound on a fallback host port (e.g. 4849 for target 4848), close now stops the port that was actually served. 1:1 auto-forwards are unchanged (`hostPort === port`).

**2. TOCTOU duplicate-forward race.** `installReservedForwarder` now claims the target in a synchronous `pendingReservedTargets` set **before** awaiting `listenHost()`, and evicts any auto-forward that already won the race via `removeAutoForwarderForTarget`. The auto-install paths (`installForwarder`, `port-listen-snapshot/opened/closed`) are gated through a shared `isReservedTarget()` that checks both the resolved and pending sets, so a snapshot/opened arriving mid-bind can no longer install a competing forward for 4848.

Both are covered by regression tests in `hostd-client.test.ts` (`…racing the in-flight reserved bind…` and `port-forward-closed carries the reserved host port…`). I verified the race test fails if the synchronous claim is removed, so it's a real guard rather than a tautology.

Note: the prior CI red was an unrelated flake — `cloudflare-named.test.ts` timed out at the 30s limit on the runner (passes locally in ~400ms); my diff only touches portbroker + agent-browser + their docs. Re-running on this push.

PR #829, opened 2026-06-12
feat(memory): add memory_search topic-slug lookup to skip redundant re-search

devxoul:
## Background

#827 reduced channel per-turn vector memory injection to **headings only** — the agent fetches a body on demand via `memory_search`. That left a redundancy: the only way to recover a body was to call `memory_search({ query: "<heading text>" })`, which runs a **second** fuzzy search (vector + keyword, RRF-fused) over all shards to re-find a topic the per-turn `hybridSearch` had **already located**. Worse, `hybridSearch` calls `loadAllShards`, so the full body was in memory at injection time and discarded — then the agent pays a whole search to read it back.

So a channel turn that opens one topic does: hybridSearch (find top-K) → strip bodies → agent re-search by heading (find the same shard again) → read body. Two searches for one shard we already had.

## Summary

Add a direct slug-lookup mode to the existing tool: `memory_search({ topic: "<slug>" })`.

- `topic` reads `memory/topics/<slug>.md` directly via `loadShard` and returns that one shard with its full body — **no search**. `query` and `topic` are mutually exclusive (exactly one required).
- The channel injection now prints each topic's `slug:` line, and the directive names the exact call (`memory_search({ topic: "<slug>" })`). The agent opens a topic **by identity**, not by re-querying text the retrieval layer already matched.
- **Why a param, not a new tool:** the user's call — "not a separated tool, `memory_search(topic: <id>)`". Keeps one memory-search surface; the model already knows the tool.
- **Why slug and not path:** `topicShardPath` already guards slugs against traversal (`..`, `/`, `\`, leading `.`) — a bad slug throws and is surfaced as a structured error, never a crash. The agent never handles raw filesystem paths.
- **Streams have no shard yet:** undreamed `source: "stream"` items get no slug hint and remain reachable only via `query`.

## What this does NOT do

- Does not change search behavior for `query` (phrase-first, token fallback, regex, ordering all unchanged).
- Does not change retrieval (`hybridSearch`, top-K, ranking) or the non-channel injection path.
- Does not auto-fetch bodies — the agent still decides which topics to open; this only makes opening one cheap and exact.
- Does not remove the `query` path as a fallback for fuzzy needs.

## Review notes

- Load-bearing change: `lookupTopic` / `topicMatchWithFullBody` in `search-tool.ts` (slug → single full-body `TopicMatch`), and the `slug:` line + directive in `renderRetrievedMemorySection` (`load-memory.ts`). `RetrievedMemoryItem` gained `source` + `key` (the injection call site already passes `HybridSearchResult`, which carries both).
- Invariants to verify: exactly-one-of `query`/`topic` (XOR guard); traversal slug → `{ error }` not throw; missing slug → `{ matches: [] }`; topic match always includes `fullBody`; stream items never emit a slug hint.
- Tests: 5 new in `search-tool.test.ts` (lookup full body / missing / traversal / neither / both) and 4 in `load-memory.test.ts` (slug exposed / directive names `topic:` / stream omits slug). Full memory suite green (661/661).


devxoul:
Good catch — addressed in 70f33224. Stream items now render a `memory_search({ query: ... })` recovery hint (they have no shard slug to look up), and the channel directive names both paths so every stripped entry — topic or stream — has an explicit way back to its full text.

PR #839, opened 2026-06-12
feat(cli): offer to resume on start/restart when init is incomplete

devxoul:
> **Stacked on #838** (`feat/init-resume-wizard-answers`). Review/merge that one first. This PR's diff is only its own 4 files.

## What

`typeclaw start` / `typeclaw restart` previously charged ahead on a half-initialized agent — one where `typeclaw init` was aborted (double-ESC) or failed after the wizard (install / dockerfile / git / hatching) — and then died deep inside the container preflight with a confusing low-level error (missing `node_modules`, missing `Dockerfile`, etc.).

Now both commands detect an incomplete init up front and offer to resume instead of failing cryptically.

## How

- Reuses **`detectInitProgress()`** from #838 — the shared "init incomplete" predicate (`a wizard checkpoint exists AND the agent never hatched`). No new detection logic; the two PRs key off the exact same signal so they can't drift.
- New `src/cli/incomplete-init.ts`:
  - `resolveIncompleteInitDecision(status, interactive)` — a **pure** policy function (continue / prompt / block), unit-tested across the full branch matrix.
  - `guardIncompleteInit(...)` — a thin orchestrator that detects progress, opportunistically clears a **stale** checkpoint (one that outlived a hatched agent, so a working agent is never falsely blocked), and returns continue / block / abort.
- **Interactive** (TTY): prompts `Try starting anyway?` (default **no**).
- **Non-interactive** (CI, hostd): fails fast with actionable guidance to run `typeclaw init`. TTY is detected via `process.stdout.isTTY`, matching the existing convention in `status.ts` / `usage.ts` / `doctor.ts`.
- `start.ts` / `restart.ts` stay thin shells; `restart` runs the check **before** `stop` (a half-init agent usually has no container to stop).

## What this does NOT do

- Does not auto-run `typeclaw init` from `start` (no CLI→CLI coupling); it guides the user to run it.
- Does not broaden "incomplete" to missing `node_modules` / `Dockerfile` on a *hatched* agent — those are start's job to regenerate. Only `checkpoint-exists AND not-hatched` counts.

## Testing

- `incomplete-init.test.ts`: 10 tests — the pure decision matrix (none / stale / incomplete × interactive / non-interactive) and the orchestrator (continue, block-non-interactive, prompt-continue, prompt-abort, stale-clear, no-clear-on-incomplete).
- `start.ts` / `restart.ts` are intentionally untested directly (thin CLI shells over the tested `guardIncompleteInit` + the already-tested domain `start()`), per the repo testing philosophy.
- Full suite: **8873 pass, 0 fail**. typecheck / lint (0 errors) / format:check clean.

### DIFF
--- a/src/cli/incomplete-init.test.ts
+++ b/src/cli/incomplete-init.test.ts
@@ -1,0 +1,138 @@
+import { describe, expect, test } from 'bun:test'
+
+import { checkpointFromSelections, type WizardCheckpointStore } from '@/init/checkpoint'
+import type { InitProgressStatus } from '@/init/progress'
+
+import { guardIncompleteInit, resolveIncompleteInitDecision } from './incomplete-init'
+
+const incomplete: InitProgressStatus = {
+  kind: 'incomplete',
+  checkpoint: checkpointFromSelections({ cwd: '/agent', vendorId: 'fireworks' }),
+}
+const stale: InitProgressStatus = {
+  kind: 'complete-stale-checkpoint',
+  checkpoint: checkpointFromSelections({ cwd: '/agent', vendorId: 'fireworks' }),
+}
+const none: InitProgressStatus = { kind: 'none' }
+
+function fakeStore(initial?: ReturnType<typeof checkpointFromSelections>): {
+  store: WizardCheckpointStore
+  cleared: () => boolean
+} {
+  let current = initial
+  let cleared = false
+  return {
+    store: {
+      load: async () => current,
+      save: async (_cwd, checkpoint) => {
+        current = checkpoint
+      },
+      clear: async () => {
+        current = undefined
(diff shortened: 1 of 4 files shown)


devxoul:
Addressed the ordering issue in 430ca9dd. `guardIncompleteInit` now runs **before** the `isInitialized` check in both `start.ts` and `restart.ts`, so a wizard-aborted init that left a checkpoint but no `typeclaw.json` now reaches the resume/block guidance instead of the generic "config file not found" error. A `continue` decision (no incomplete checkpoint, or the user chose "try anyway") falls through to `isInitialized`, which still catches a truly uninitialized directory. Rebased on the latest #838.

PR #848, opened 2026-06-13
fix(memory): quiet onnxruntime cpuid warning and downgrade embed-bound log

devxoul:
## Background

Two separate log lines spammed container boot on every vector index build,
both tied to `configureTransformers` / `embed` in the memory plugin embedder.
Neither represents a real problem — quieting them makes the remaining signal
(real errors, `[dreaming] done over_budget=N`) legible again.

---

### Fix A — suppress onnxruntime `cpuid_info warning: Unknown CPU vendor`

```
onnxruntime cpuid_info warning: Unknown CPU vendor. cpuinfo_vendor value: 0
```

**Root cause.** onnxruntime-node runs a cpuid probe when it creates an inference
session. The probe uses pytorch/cpuinfo, which initializes successfully but cannot
match the CPUID vendor string against its known-vendor table when the container
runs on an emulated/virtualized CPU — Rosetta 2 / QEMU on Apple Silicon hosts
present a non-standard CPUID string. On success with unknown vendor it logs this
warning and sets `vendor=0`.

**Why it is safe to suppress.** The vendor field is purely informational: device-info
reporting only. SIMD kernel dispatch (AVX, AVX2, AVX512, NEON, F16C, …) is detected
on a completely independent codepath (`X86Init` via raw CPUID instructions,
`ArmLinuxInit` via auxval/hwcap) — vendor=unknown has zero perf or correctness
impact. onnxruntime already suppresses this same warning on wasm
(`cpuid_info_vendor.cc`), and upstream issue microsoft/onnxruntime#27336 was closed
with no code fix needed.

**Fix.** In `configureTransformers`, set the onnx backend log level to `'error'`
(`env.backends.onnx.logLevel = 'error'`, mirrored via `setLogLevel(3)`) before
calling `pipeline()`. onnxruntime-node reads the level when it creates the
inference session — inside `pipeline()` — so the level is in place before the
cpuid probe fires. Guarded with optional chaining because the backend object is
absent under the test mock.

**Tradeoff.** Setting `'error'` also hides a separate, genuinely serious
onnxruntime warning: "Failed to initialize PyTorch cpuinfo library. May cause CPU
EP performance degradation." That warning does NOT fire in our container — `vendor=0`
proves cpuinfo init succeeded. Real session/inference errors still surface at
`'error'` level, so the tradeoff is acceptable. Reviewers should keep this in mind
if the cpuinfo-init-failure warning is ever needed for diagnosis.

---

### Fix B — downgrade per-embed bound log `warn` → `info`

```
[memory] vector embedding: bounded 40/3283 passage input(s) to the 512-token model
limit (worst ~1538 est. tokens); their tail is not embedded
```

**Why it was a false alarm.** Bounding is the designed, deterministic path
(`boundEmbeddableText`). The real remediation for over-budget shards is the dreaming
subagent, which compacts verbose topic shards on its own schedule
(`*/30 * * * *` default). A non-zero bounded count on a routine boot index build is
expected and non-actionable — legacy verbose shards, long fragments, and long queries
all trigger it by design until dreaming catches up. The actionable signal is the
separate `[dreaming] done over_budget=N` line, not this per-embed count.

**Fix.** `console.warn` → `console.info` in `warnIfBounded`. Message is unchanged
(still content-free: counts + worst-case token estimate only, never the text itself).

---

## What this does NOT do

- Does not change any embedding logic, SIMD dispatch, vector store schema, or
  relevance scoring.
- Does not remove the embed-bound log line — it stays at `info` for debugging.
- Does not suppress real onnxruntime session or inference errors (those surface at
  `'error'` level).
- Does not change dreaming's compaction schedule or over-budget routing (see the
  `fix(memory): bound embedding inputs` PR for that).

---

## Review notes

- **Timing invariant (Fix A):** `configureTransformers` runs before `pipeline()`
  creates the inference session. That ordering is what makes the log-level write
  effective — if it were moved inside `pipeline()`'s callback, it would be too late.
  `embedder-onnx-loglevel.test.ts` guards this: the mock captures the onnx env at
  session-creation time and asserts `logLevel === 'error'` and `setLogLevelArg === 3`.
- **Tradeoff (Fix A):** `'error'` also suppresses the cpuinfo-init-FAILURE warning.
  Acceptable because that warning does not fire here (vendor=0 means init succeeded),
  but worth knowing if symptoms ever suggest cpuinfo did NOT initialize.
- **Test update (Fix B):** `embedder-truncation.test.ts` now spies on `console.info`
  instead of `console.warn`. Assertions are otherwise unchanged.
- **Sibling PR:** the agent-browser plugin log-noise fix ships in PR #847.

---

## Testing

- `bun run test --parallel` on the three embedder files — 10 pass, 0 fail
  (includes new `embedder-onnx-loglevel.test.ts` and updated
  `embedder-truncation.test.ts`)
- `bun run typecheck` — clean
- `bun run lint` — 0 errors
- `bun run format` — clean

### DIFF
--- a/src/bundled-plugins/memory/vector/embedder-onnx-loglevel.test.ts
+++ b/src/bundled-plugins/memory/vector/embedder-onnx-loglevel.test.ts
@@ -1,0 +1,32 @@
+import { describe, expect, mock, test } from 'bun:test'
+
+// Guards the onnxruntime cpuid-warning suppression: configureTransformers must
+// drop the onnx backend log level to 'error' BEFORE pipeline() creates the
+// inference session (where onnxruntime-node reads the level and runs the cpuid
+// probe). The mock captures the env onnxruntime sees at session-creation time.
+let onnxAtPipeline: { logLevel?: string; setLogLevelArg?: number } | undefined
+
+const onnx: { logLevel?: string; setLogLevel: (level: number) => void; setLogLevelArg?: number } = {
+  setLogLevel(level: number) {
+    this.setLogLevelArg = level
+  },
+}
+
+mock.module('@huggingface/transformers', () => ({
+  env: { backends: { onnx } },
+  pipeline: async () => {
+    onnxAtPipeline = { logLevel: onnx.logLevel, setLogLevelArg: onnx.setLogLevelArg }
+    return () => ({ data: new Float32Array(768) })
+  },
+}))
+
+describe('onnxruntime log-level suppression', () => {
+  test("lowers the onnx backend to 'error' and mirrors via setLogLevel before the session is created", async () => {
+    const { getEmbedder } = await import('./embedder')
+
+    await getEmbedder()
+
+    expect(onnxAtPipeline?.logLevel).toBe('error')
+    expect(onnxAtPipeline?.setLogLevelArg).toBe(3)
+  })
+})
--- a/src/bundled-plugins/memory/vector/embedder-truncation.test.ts
+++ b/src/bundled-plugins/memory/vector/embedder-truncation.test.ts
@@ -60,37 +60,37 @@ describe('embedder bounding', () => {
 })
 
 describe('embedder bounding observability', () => {
-  let warnings: string[]
-  const originalWarn = console.warn
+  let notices: string[]
+  const originalInfo = console.info
 
   beforeEach(() => {
-    warnings = []
-    console.warn = (message?: unknown) => {
-      warnings.push(String(message))
+    notices = []
+    console.info = (message?: unknown) => {
+      notices.push(String(message))
     }
   })
 
   afterEach(() => {
-    console.warn = originalWarn
+    console.info = originalInfo
   })
 
-  test('emits a content-free warning naming the type and count when an input is bounded', async () => {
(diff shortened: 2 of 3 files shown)


devxoul:
Closing in favor of #NEW (branch `fix/embed-batch-oom-diagnostics`).

This PR's premise was wrong. It treated two boot-log lines as noise to suppress/quiet, but investigation revealed the agent (`typeeybot`) was actually **OOM-killed** (`OOMKilled=true`, exit 137) during the startup vector-index build — embedding 3283 passages in a single onnxruntime pass. The `bounded 40/3283` log was a load-bearing breadcrumb, not noise; downgrading it warn→info would have hidden the symptom. And the onnxruntime `Unknown CPU vendor` line is a native warning we shouldn't mute (doing so also hides the separate, genuine 'cpuinfo init failed' perf warning).

Replacement PR keeps the embed-bound log at warn and **adds** an explicit single-pass batch-size signal that warns loudly above 256 with an OOM hint, so a future OOM boot leaves a clear breadcrumb. No log suppression. A separate change will batch the embed to bound peak memory (the actual fix).

PR #857, opened 2026-06-13
fix(memory): resolve reference slugs in memory_search topic lookup

devxoul:
## Background

The per-turn memory injection renders reference entries with the same `slug:` line and the same recovery hint (`memory_search({ topic: "<slug>" })`) as topic entries. But `memory_search`'s `topic` lookup path only read `memory/topics/<slug>.md` via `loadShard` — it never checked `memory/references/`. A reference slug surfaced in channel memory therefore returned `{ matches: [] }` with a note that explicitly says "do not fall back to grep/find/bash — accept no memory exists." The agent hit a dead end and could only recover the verbatim content by disobeying that note and running a filesystem search the tool contract forbids.

The root cause was that the injected memory block and the lookup tool made the same claim about the `topic` parameter — "use this to read an entry whose slug you already have" — but the tool only honored the claim for topic shards. References were effectively unreachable via the one path a channel agent is told to use.

## Summary

Make `lookupTopic` try the topic shard first (common case + precedence), then fall back to a reference of the same slug when `loadShard` returns `null`. A new exported `loadReference(agentDir, slug)` is added to `references/load-references.ts`; it reuses the existing `readAndParseReference` parse path and the traversal guard from `referenceFilePath`, so a path-traversal slug still returns a structured error rather than a throw. A matching `referenceMatchWithFullBody` helper in `search-tool.ts` projects the reference into the `ReferenceMatch` shape the tool already knows how to return.

The tool description is updated to reflect that `topic` resolves either a topic shard or a reference, with topic taking precedence.

## What this does NOT do

- Does not change the behavior when a topic shard exists — that path is untouched and takes precedence.
- Does not change the search (`query`) path at all.
- Does not modify reference storage, frontmatter, or the decay/saturation logic.
- Does not relax the traversal guard — bad slugs still return a structured error, not a crash.

## Review notes

The load-bearing change is the three-way branch in `lookupTopic` in `search-tool.ts`:

1. `loadShard` throws — structured error (traversal guard, unchanged).
2. `loadShard` returns non-null — return topic match (unchanged).
3. `loadShard` returns null — call `loadReference`; return reference match or empty matches.

New tests in `search-tool.test.ts` cover both branches of the fallback: (1) topic lookup falls back to a reference of the same slug when no topic shard exists, and (2) topic lookup returns the topic shard and ignores the same-slug reference when both exist.

All checks green: `bun run typecheck`, `bun run lint` (0 errors), `bun run format:check`, `bun run test` (8962 pass, 0 fail).


devxoul:
Addressed in a follow-up commit (`address review: bump access on topic-to-reference lookup; mention references in empty-result note`):

- **Access not recorded:** the topic→reference fallback now routes through `bumpReturnedReferences` before returning, so a reference reached via the injected slug advances `accessCount`/`lastAccessed` exactly like a query hit. Added a test asserting the bump.
- **Empty-result guidance:** updated to name `memory/references` alongside topics and streams.

PR #858, opened 2026-06-13
feat(memory): index references on write, not only at startup

devxoul:
## Background

The memory reference tier had a blind spot: `store_reference` wrote the markdown file but never touched the vector index. The startup index build caught up on the next container restart, but any reference stored mid-session was invisible to vector retrieval until then. Stream fragments don't have this problem — the fragment append hook embeds them immediately on every append. References had no equivalent.

This was confirmed in live testing: a freshly stored reference showed 0 rows in the vector index until a restart.

## Summary

Mirrors the existing stream-fragment on-write hook pattern for references.

**Why extract `referencePassagesForOne` (commit 1)?** The on-write hook needs to derive chunk ids, content hashes, and passage shapes for a single reference — the same way the startup build does it for the whole corpus. Pulling that derivation into an exported function means the on-write rows are byte-identical to what a later startup rebuild produces (same ids, chunks, content hashes). No churn, no stale duplicates, no divergence between the two paths.

**Why the stale-chunk pruning (commit 2)?** Re-storing a slug with a shorter body produces fewer chunks. Without pruning, the old higher-index rows stay in the index and surface as orphaned retrieval hits for content that no longer exists. The hook computes the wanted id set, upserts changed chunks, and deletes any existing row for the slug that is no longer wanted.

**Why gate on the vector store (commit 3)?** The new hook is attached under the same condition that gates fragment on-write indexing. Both hooks share the same vector store lifecycle. Demoted references yield no rows, matching the startup exclusion.

## What this does NOT do

- Does not change the startup index build, the dreaming pass, or any other memory tier.
- Does not alter the passage output shape — the commit 1 refactor is behavior-identical. An existing dreaming test caught the first attempt (the demoted-exclusion flag was silently dropped); the fix threads the flag through explicitly.
- Does not add new config fields or schema changes.

## Review notes

**Load-bearing change:** `referencePassagesForOne` in vector/passages.ts must produce identical output to the old inline code inside the startup build's reference loop. Key invariant: the demoted flag is forwarded from reference frontmatter so a demoted reference yields an empty passage list both at startup and on write.

**Stale-chunk pruning:** `makeReferenceStoredHook` scans all stored rows for the slug-scoped id prefix, drops anything absent from the wanted set, then upserts changed chunks. The prefix is slug-scoped so no other slug's rows are affected.

**Regression note:** An existing dreaming test failed on the first refactor attempt because the demoted flag was silently dropped. That test is now green and confirms the refactor preserves the startup-exclusion behavior.

**Test coverage:** 4 new tests in `reference-index-on-write.test.ts` — embeds on store, skips unchanged body, prunes stale chunks on shrink, no rows when demoted. All checks pass: typecheck, lint (0 errors), format, test suite (8964 pass, 0 fail).


devxoul:
Addressed in a follow-up commit (`address review: prune vector rows when a reference is demoted`):

`runReferenceSaturationPass` now tracks demoted slugs and calls `deleteReferenceVectors(agentDir, demotedSlugs)` in the same pass, so demotion prunes the `reference:<slug>#*` rows immediately instead of leaving them live until the next restart. This closes the gap where the on-write hook indexed a reference while `demoted:false` and demotion alone didn't remove those rows. Added a test that seeds a reference vector, runs the demotion path, and asserts the rows are gone.

PR #859, opened 2026-06-13
feat(memory): expose store_reference to the main agent for inline capture

devxoul:
## Background

`store_reference` lived only on the `memory-logger` subagent, which fires on the post-turn idle cycle (`idleMs` debounce + a fast-model spawn). When a user asked the agent to remember something verbatim — a SQL query, a runbook, a pasted spec — no durable artifact existed until that cycle ran. A same-turn follow-up ("give me that back") arrived before the subagent had fired and found nothing.

The fix is to register `store_reference` directly in the main agent's tools map so the agent can persist the artifact inline, without waiting for the idle cycle.

## Summary

Register `store_reference` as a main-agent tool in the memory plugin's `tools` map (`index.ts`). The tool is wired to `makeReferenceStoredHook` (from PR #858) when vector is enabled, so the new reference is embedded immediately and searchable in the same turn. When vector is disabled it writes the reference file and returns the slug with no further side-effects — same behaviour the `memory-logger` path already has.

References are always-on as of the upstream commit that removed the `memory.references.enabled` gate, so `store_reference` is registered unconditionally — no config check.

**Why not just fire the memory-logger sooner?** The idle cycle exists to batch transcript distillation affordably; pulling it forward defeats the batching. Inline `store_reference` is the right scope: a single artifact, user-initiated, with no transcript scan involved.

## What this does NOT do

- Does not duplicate the `memory-logger`'s transcript-distillation path. The inline tool only persists what the agent explicitly passes to it.
- Does not change the `store_reference` tool's implementation, storage format, or the on-write hook contract — those land in PR #858.
- Does not add a config gate. References are always-on; the registration is unconditional.

## Review notes

Load-bearing addition is in `index.ts` at the `tools` map:

```ts
store_reference: createStoreReferenceTool(
  appendVectorStore !== undefined ? makeReferenceStoredHook(appendVectorStore) : undefined,
),
```

The conditional mirrors exactly how the `memory-logger` subagent wires the same hook. Verify that `makeReferenceStoredHook` is imported from the same source PR #858 adds it to — no new hook definition lives here.

**Stacking note:** this branch is stacked on `feat/memory-reference-on-write-index` (PR #858) for `makeReferenceStoredHook`. Merge after #858, or rebase onto `main` if #858 merges first.

**Test coverage:** new assertion in `index.test.ts` that `exports.tools?.store_reference` is defined after plugin boot. All checks green: `bun run lint` (0 errors), `bun run format:check`, `bun run typecheck`, `bun run test` (8966 pass, 0 fail).


devxoul:
Closing this — on reflection it's not worth the cost.

The memory-logger subagent already carries `store_reference` and fires after every turn (idle/buffer-trip/session-end), so references are persisted without this. The only thing the inline tool adds is same-turn availability, and that's mostly theoretical: within a session the agent can already reproduce the verbatim content from its own context, and the real value of a stored reference is cross-session — which the memory-logger covers.

Against that thin benefit, the inline tool adds main-agent system-prompt surface on every turn, leaks the observation responsibility back into the main agent (the design deliberately keeps "what to remember" in the memory-logger), and risks duplicate references (no content-hash dedup exists, so inline + memory-logger could store the same content twice).

Net: not worth it. The other memory PRs stand on their own.

PR #866, opened 2026-06-13
fix(todo): auto-clear resolved list inside todo_write

devxoul:
## Background

In a live channel session, an agent marked all 5 todos as cancelled (0 incomplete items remaining), posted a reply, then had the **turn aborted** (`stopReason: "aborted"`, tokens 0) on the very next step — exactly where the follow-up `todo_clear` call would have run. The resolved list lingered on disk indefinitely.

The continuation system already stops nudging once there are no incomplete items (`incompleteTodos()` excludes both completed and cancelled), so the list wasn't causing any functional harm. But it was a cosmetic lie: a fully-resolved list sitting on disk looked like unfinished business to anything reading the file directly, and the agent had been told (via tool description and continuation nudge) to explicitly call `todo_clear` at the end — a step that was structurally racy against turn aborts.

## Summary

Move the cleanup into `todo_write` itself: when the written list has no incomplete items left (`remaining === 0 && list.length > 0`), immediately call `writeTodos([])` and return `"All N todo(s) done; list cleared."` — in the same call, no follow-up needed.

**Why here and not in a post-turn hook?** Doing it inside the write makes "no resolved list is persisted" true by construction. A post-turn hook would still be racing against aborts; putting it in the write folds the cleanup into the same atomic action the agent is already performing.

**Why keep `todo_clear` at all?** Its one remaining job is abandoning a task when items are still incomplete (`remaining > 0`). The auto-clear guard fires only on fully-resolved lists, so the two tools no longer overlap. The description is updated to reflect the narrowed scope.

The continuation nudge is updated from "call `todo_clear` when done" to "the list is cleared for you automatically once no incomplete items remain," which is now truthful.

## What this does NOT do

- Does not change how the continuation system counts incomplete items — `incompleteTodos()` is untouched.
- Does not remove `todo_clear`; it retains its role for explicit task abandonment.
- Does not affect lists that still have incomplete items — partial lists persist exactly as before.
- Does not change any persistence format or session state.

## Review notes

The load-bearing change in `todo_write` (in `src/agent/tools/todo/index.ts`):

```ts
if (remaining === 0 && list.length > 0) {
  await writeTodos(agentDir, [])
  return `All ${list.length} todo(s) done; list cleared.`
}
```

This fires only when the incoming write produces a non-empty list with zero incomplete items. An empty-list write (already `writeTodos([])`) skips the branch and hits the existing empty-list path below.

The two new tests in `index.test.ts`:
- All items done/cancelled → list written as `[]`, return value contains "cleared".
- At least one incomplete item → list persists as written (auto-clear does not fire).

Verification: `bun run typecheck` clean, `bun run lint` 0 errors (67 pre-existing warnings in `docs/`, unrelated), `bun run format:check` clean, `bun run test` 8982 pass / 0 fail / 1 skip.


devxoul:
Good catch — fixed in f980c8ee. The `## Tracking your work` section of the main system prompt (`src/agent/system-prompt.ts`) now says the runtime clears a fully-resolved list automatically once `todo_write` leaves no incomplete items, and scopes `todo_clear` to abandoning a task with items still incomplete — consistent with the `CONTINUATION_PROMPT` and tool-description edits. That was the third and last place carrying the stale instruction.

PR #869, opened 2026-06-14
perf(memory): dedupe memory-logger system prompt (~6.4k→4.2k tok)

devxoul:
## Background

`memory-logger`'s system prompt rides on **every** spawn, and the plugin spawns it often: on idle windows, on buffer-trip (`memory.bufferBytes` of transcript growth), and on `session.end` — coalesced per `agentDir`. A chatty channel agent trips this many times an hour.

That prompt had quietly grown to **~6,400 tokens (25,754 chars)** — 6.4× the `~1000 tok` figure the `debug:prompt` placeholder still cites. Profiling the sections showed the size came from **accretion, not content**: each past fix added a paragraph rather than editing the relevant one, so the same rule ended up restated in three or four places.

- The watermark/read loop (`find_entry` → `read` → advance `latestEntryId` → stop at `totalLines`) was taught across three sections: *Reading the transcript*, *Watermark contract*, and *Stopping*.
- "Passive context, not authorization" was stated four times.
- *Capture philosophy* overlapped *What to capture* / *What to skip* ("most runs 0–1 fragments", "bar is high", "over-writing is the common mistake").
- The "Content the user explicitly taught the agent" subsection carried 5–6 multilingual examples per intent family.

Since this prompt is in the per-spawn cost path, redundancy here is the single highest-leverage token win in the memory subsystem.

## Summary

Consolidate the restatements into one canonical statement each, and trim the teaching examples from 5–6 per family to 2–3 representative ones. **No rule is removed — only redundancy.**

The safety net is the existing test suite: `memory-logger.test.ts` pins an exact substring/regex for each distinct rule (secret-handling, dedup contract, watermark contract, per-fragment provenance, the `find_entry`/`totalLines` stop trigger, the anti-patterns list, etc.). A dropped constraint fails CI, so "behavior-preserving" is machine-verified, not asserted by prose.

Result: **25,754 → 16,598 chars (~6,439 → ~4,150 tok, −35.6%)** off every `memory-logger` spawn, with byte-identical behavior.

Why prose-merge and not a structural rewrite: the prompt's *content* is correct and battle-tested (each paragraph traces to a real past failure mode). The problem was duplication, so the minimal fix is deduplication — not re-deriving the rules.

## What this does NOT do

- Does not change the `memory-logger` handler, payload schema, tools, budgets, or any export — only the `MEMORY_LOGGER_SYSTEM_PROMPT` string body changed.
- Does not weaken the security section: every credential pattern and the Allowed/Forbidden `GH_TOKEN` examples are retained verbatim.
- Does not touch `dreaming`, `memory-retrieval`, or the injection path.
- Does not retune `bufferBytes` / `idleMs` / spawn frequency — that's a separate measurement-backed change.

## Review notes

- Load-bearing diff: the three merged watermark sections. Verify the stop-trigger contract survived — the two verbatim sentences the test pins (`find_entry gives you totalLines=T up front…` and `The hard stop is totalLines: …`) are intact, and the forbidden `more than a handful of reads/times` phrasing stays absent (both enforced by `memory-logger.test.ts`).
- Verification run locally: `bun run lint` (0 errors), `bun run format:check` (clean), `bun run typecheck` (clean), `bun run test` (9008 pass / 0 fail), including the `scripts/dump-system-prompt.test.ts` cache-order contract.

### DIFF
--- a/src/bundled-plugins/memory/memory-logger.ts
+++ b/src/bundled-plugins/memory/memory-logger.ts
@@ -66,177 +66,140 @@ export function isMemoryLoggerPayload(value: unknown): value is MemoryLoggerPayl
 
 export const MEMORY_LOGGER_SYSTEM_PROMPT = `You are typeclaw's memory-extraction subagent.
 
-Your job is to read a session transcript and capture, as fragments, only the durable operational facts a future agent in a future session would concretely need — explicit user instructions, stable identity/role/tool facts, decisions with reasoning, reproducible workarounds, and anything the user explicitly taught the agent or asked it to remember. You write zero or more fragments to today's memory stream file. Then you exit. Most runs produce zero or one fragment; that is the expected output, not a failure.
+Read the parent session transcript past the watermark and write zero or more durable memory fragments to today's stream, then exit. Capture only operational facts a future agent would concretely need: explicit user instructions, stable identity/role/tool facts, decisions with reasoning, reproducible workarounds, corrections, changed minds, and content the user explicitly taught the agent or asked it to remember. Most runs produce zero or one fragment; that is expected.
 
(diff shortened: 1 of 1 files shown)


devxoul:
@typeey please approve

PR #873, opened 2026-06-14
feat(provider): allow selecting models beyond the curated list

devxoul:
## Background

Model selection has been gated by a hardcoded allowlist. `KNOWN_PROVIDERS` (`src/config/providers.ts`) is the single source of truth, and its refs back a Zod enum (`z.enum(knownModelRefs)` in `src/config/config.ts`). Any `<provider>/<model>` ref outside that table fails schema validation the moment it is written to `typeclaw.json` — at `model set`/`add`, at init scaffolding, and at every config load. The init wizard and `model list --available` only ever surfaced curated models too, even though `src/init/models-dev.ts` already fetched the full models.dev catalog and then discarded everything not in the allowlist.

Net effect: a user could not pick a model a provider had just released (or any model we hadn't yet hand-added) without editing the source and cutting a release. This PR makes discovery dynamic and lets users actually select and run non-curated models.

## Summary

- **Two-tier ref types.** Keep `KnownModelRef` (the curated template-literal union) for curated-only APIs and editor autocomplete; add a broad `ModelRef` brand plus `isModelRef` (validates `<known-provider>/<model-id>` shape — the provider must still be one we know, so we have a `baseUrl`/`api`/auth surface).
- **Loosened gate, preserved autocomplete.** The config schema becomes a union of the curated enum **and** a validated custom-ref branch. Why a union and not a bare `z.string().refine`: the enum branch keeps the curated refs in the generated JSON Schema, so editors still autocomplete them.
- **Synchronous runtime synthesis.** `resolveModel` stays sync (it runs at session creation, no fetch). For a custom ref it synthesizes a pi-ai `Model<>`: inherits `api`/`baseUrl` from the provider's first curated model (the wire transport is per-provider, not per-model), and fills name/reasoning/input/context/maxTokens from a new reloadable `customModels` config map, written at selection time when the async catalog is in hand.
- **Cost is explicit zero when unknown** — never a guessed or copied price.
- **Discovery for all probed providers** via models.dev, surfaced in the wizard and `model list --available`, offline-safe via the existing curated fallback.
- **Both selection paths persist metadata.** Interactive picking and the explicit `model set/add <profile> <ref>` flow both write `customModels[ref]` from the catalog; the non-interactive path warns when a ref isn't in the catalog so the fallback-default synthesis is never silent.

Split into reviewable commits: the `ModelRef`/`isModelRef` primitive; the config gate + `customModels` + `resolveModel` (with the mechanical `ModelRef` type ripple to runtime call sites); the models.dev broadening; the persistence layer; the CLI/init wiring; and the review follow-up for the explicit-ref metadata path.

## Usage — per provider

`typeclaw model list --available` now lists the live models.dev catalog for every provider in `KNOWN_PROVIDERS`, not just the curated allowlist. The listing below shows the curated refs that ship today; when models.dev is reachable, any newer model each provider has published is appended under the same provider heading and is selectable with `typeclaw model set <profile> <ref>`:

```text
$ typeclaw model list --available
Use `typeclaw model set <profile> <ref>` to apply.

OpenAI
  openai/gpt-5.4-nano
  openai/gpt-5.4-mini
  openai/gpt-5.4
  openai/gpt-5.5

OpenAI Codex (ChatGPT Plus/Pro)
  openai-codex/gpt-5.4-mini
  openai-codex/gpt-5.4
  openai-codex/gpt-5.5

Anthropic
  anthropic/claude-haiku-4-5
  anthropic/claude-sonnet-4-6
  anthropic/claude-opus-4-7
  anthropic/claude-opus-4-8

Fireworks
  fireworks/accounts/fireworks/routers/kimi-k2p6-turbo

Z.AI
  zai/glm-4.5-air
  zai/glm-4.6
  zai/glm-4.7

Z.AI (GLM Coding Plan)
  zai-coding/glm-4.5-air
  zai-coding/glm-4.7
  zai-coding/glm-5
  zai-coding/glm-5-turbo
  zai-coding/glm-5.1

xAI (Grok)
  xai/grok-4.3
  xai/grok-4.20-0309-reasoning
  xai/grok-4.20-0309-non-reasoning
  xai/grok-build-0.1

MiniMax
  minimax/MiniMax-M3
  minimax/MiniMax-M2.7
  minimax/MiniMax-M2.5
  minimax/MiniMax-M2.1
  minimax/MiniMax-M2

DeepSeek
  deepseek/deepseek-v4-flash
  deepseek/deepseek-v4-pro

Moonshot (Kimi)
  moonshot/kimi-k2.7-code
  moonshot/kimi-k2.6
  moonshot/kimi-k2.5

Moonshot (Kimi Coding Plan)
  moonshot-coding/kimi-for-coding
```

Before this PR a user could only pick a ref from that curated set; now they can also pick any newer `<provider>/<model-id>` models.dev surfaces for a known provider (e.g. a fresh OpenAI or Anthropic release that hasn't been hand-added to `KNOWN_PROVIDERS` yet) without a source edit.

Offline (or when models.dev is unreachable) the command falls back to exactly the curated list above and prints a `Using built-in catalog (models.dev unavailable: …)` note. Each provider's env var: `OPENAI_API_KEY`, OAuth for `openai-codex`, `ANTHROPIC_API_KEY` (or OAuth), `FIREWORKS_API_KEY`, `ZAI_API_KEY`, `ZAI_CODING_API_KEY`, `XAI_API_KEY` (or OAuth), `MINIMAX_API_KEY`, `DEEPSEEK_API_KEY`, `MOONSHOT_API_KEY`, `MOONSHOT_CODING_API_KEY`.

Picking a non-curated ref records its catalog metadata into `typeclaw.json#customModels`; curated refs need no entry. Fallback chains mix curated and custom freely, e.g. `"default": ["anthropic/claude-opus-4-8", "anthropic/claude-sonnet-4-6"]`.

## What this does NOT do

- **Does not validate upstream availability.** A ref is checked for shape and a known provider only — never "does this provider actually serve this model". If it doesn't, the turn fails at runtime; fallback chains remain the safety net.
- **Does not add new providers.** The provider must already exist in `KNOWN_PROVIDERS` (so auth/baseUrl/api are known). Only the *model* is dynamic.
- **Does not infer pricing.** Unknown cost is zero, not estimated, so usage reporting under-reports rather than lies.
- **Does not reorder `KNOWN_PROVIDERS`** (order is load-bearing for the schema enum and `provider --help`).

## Review notes

- Load-bearing: `resolveModel` synthesis (`src/config/config.ts`) and `isModelRef` (`src/config/providers.ts`). Verify the synthesized `Model<>` carries every field pi-ai needs (id, name, api, provider, baseUrl, reasoning, input, cost, contextWindow, maxTokens).
- The explicit `model set/add <ref>` path routes through `resolveExplicitRef` (`src/cli/model.ts`): curated refs resolve from `KNOWN_PROVIDERS`; non-curated refs are looked up in the live catalog so `customModels[ref]` matches the interactive picker; a catalog miss still writes the ref but warns about the fallback. Covered by new unit tests in `src/cli/model.test.ts`.
- `customModels` is classified `applied` (reloadable) in `FIELD_EFFECTS` — the guard test in `src/config/reloadable.test.ts` enforces a classification exists.
- The `isModelRef` regex intentionally allows `/` in the model-id segment so Fireworks paths (`fireworks/accounts/fireworks/models/...`) validate; only the provider segment is constrained.
- `typeclaw.schema.json` regenerates with no further diff (already current on the branch). `bun run typecheck` / `lint` (0 new warnings) / `format` / `test` (9026 pass, 0 fail) all green.

### DIFF
--- a/src/agent/index.ts
+++ b/src/agent/index.ts
@@ -15,7 +15,7 @@ import { loadMemory } from '@/bundled-plugins/memory/load-memory'
 import type { ChannelRouter } from '@/channels/router'
 import type { ReactionRef } from '@/channels/types'
 import { getConfig, resolveModel, resolveProfile } from '@/config'
-import { defaultThinkingLevelForRef, providerForModelRef, type KnownModelRef } from '@/config/providers'
+import { defaultThinkingLevelForRef, providerForModelRef, type ModelRef } from '@/config/providers'
 import { renderMcpCatalog } from '@/mcp/catalog'
 import type { McpManager } from '@/mcp/manager'
 import { createMcpDispatcherTools, MCP_DISPATCHER_TOOL_NAMES } from '@/mcp/tools'
@@ -196,7 +196,7 @@ export type CreateSessionOptions = {
   // pinned to the next ref in the chain after the previous one failed. When
   // set, `profile` is still recorded for the fallback-warning bookkeeping;
   // the profile→refs resolution is skipped.
-  refOverride?: KnownModelRef
+  refOverride?: ModelRef
   // Defensive ceiling on cumulative bytes of tool-result text per session,
   // applied to the named tools only. See `src/agent/tool-result-budget.ts`
   // for the rationale. Intended for subagents that read large files
@@ -247,7 +247,7 @@ export async function createSessionWithDispose(options: CreateSessionOptions = {
   // exactly what they're doing.
   // `refOverride` lets the model-fallback helper pin a specific entry from
   // the chain when it recreates a session after the previous ref failed.
-  const activeRef: KnownModelRef = options.refOverride ?? resolved.ref
+  const activeRef: ModelRef = options.refOverride ?? resolved.ref
   const { authStorage, modelRegistry } = getAuthFor(providerForModelRef(activeRef))
 
   const materializedSkills =
--- a/src/agent/model-fallback.test.ts
+++ b/src/agent/model-fallback.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, test } from 'bun:test'
 
-import type { Models } from '@/config/config'
+import { configSchema, type Models } from '@/config/config'
 import type { KnownModelRef } from '@/config/providers'
 
 import type { AgentSession } from './index'
@@ -9,6 +9,10 @@ import { promptWithFallback, resolveFallbackChain } from './model-fallback'
 const REF_A = 'openai/gpt-5.4-nano' as KnownModelRef
 const REF_B = 'fireworks/accounts/fireworks/routers/kimi-k2p6-turbo' as KnownModelRef
 const REF_C = 'zai/glm-4.6' as KnownModelRef
+
+function parseModels(models: Record<string, string | string[]>): Models {
+  return configSchema.parse({ models }).models
+}
 
 type SessionFake = {
   ref: KnownModelRef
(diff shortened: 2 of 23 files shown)


devxoul:
Addressed in cf3c4c77. The non-interactive `model set/add <profile> <ref>` path now routes through `resolveExplicitRef`, which looks a non-curated ref up in the live catalog and persists the same `customModels[ref]` metadata the interactive picker would. On a catalog miss (offline / genuinely unknown id) it still writes the ref but emits a `log.warn` spelling out the fallback defaults (reasoning off, text-only input, zero cost, provider-default context), so the degradation is never silent. Added three unit tests covering curated (no meta), non-curated-in-catalog (meta carried), and non-curated-missing (ref-only + warn).

Issue #909, opened 2026-06-16
perf(memory): compact channel memory index to one line per topic


devxoul:
## Problem

The channel / headings-only memory index injects each topic as a 3-line block:

```
## Service deployment topology
slug: `service-deployment-topology`

```

The `slug:` line is largely a kebab-cased echo of the heading, and the `## ` header + `slug: ` framing + blank line is boilerplate repeated per topic. With ~10 topics that's a lot of structural tokens, every channel turn, for two short strings.

## Why not just drop one

- **Can't drop the slug** — it's the exact `memory_search({ topic })` key and is NOT a clean slugify of the heading (e.g. a heading like `EXT-1.5/REL-2.1 acceptance-criteria scope` maps to `ext-1-5-rel-2-1-ac-scope`), so the agent can't derive it.
- **Can't drop the heading** — it's the relevance signal (belief sentence) the model reads to decide whether to open the topic.

## Change

Collapse each entry to one line, keeping both heading and slug:

```
- Service deployment topology `service-deployment-topology`
- Release parity verification approach `release-parity-verification-approach`
```

3 lines to 1, no information loss; slug stays verbatim. Applies to the channel branch of `renderRetrievedMemorySection` and to `renderTopicIndexMemorySection` (`src/bundled-plugins/memory/load-memory.ts`). Stream items (no slug) keep a `_(recent observation)_` marker so the recovery directive still distinguishes the two `memory_search` paths.

## Preserved

`# Memory` framing, channel-bleed boundary, recovery directive, slugs verbatim, and the non-channel excerpt path (unchanged).

Est. ~40% lighter on the channel index block.

devxoul:
Shipped in #910 (merged).

PR #910, opened 2026-06-16
perf(memory): compact channel memory index to one line per topic

devxoul:
## What

Collapses the channel / headings-only memory index from a 3-line block per topic to one line.

Before:

```
## Service deployment topology
slug: `service-deployment-topology`
```

After:

```
- Service deployment topology `service-deployment-topology`
```

Applies to the channel branch of `renderRetrievedMemorySection` and to `renderTopicIndexMemorySection` (`src/bundled-plugins/memory/load-memory.ts`). Stream items (no slug) keep a `_(recent observation)_` marker.

## Why

The `## ` / `slug: ` / blank-line framing is boilerplate repeated per topic, and the slug is mostly a kebab echo of the heading — a lot of structural tokens on every channel turn. ~40% lighter on the index block, **no information lost**: heading and slug both stay, slug verbatim for `memory_search({ topic })`.

The slug can't be dropped (it's the lookup key and isn't a clean slugify of the heading), and the heading can't be dropped (it's the relevance signal) — so they're kept, just on one line without the framing.

## Preserved

`# Memory` framing, channel-bleed boundary, recovery directive, slugs verbatim, and the non-channel excerpt path (unchanged).

## Test plan

`typecheck` / `format` / `lint` clean; full suite 0 fail. Updated channel-format assertions in `load-memory.test.ts` and `index-vector.test.ts`.

Refs #909.

### DIFF
--- a/src/bundled-plugins/memory/index-vector.test.ts
+++ b/src/bundled-plugins/memory/index-vector.test.ts
@@ -96,10 +96,8 @@ describe('vector session.turn.start hook', () => {
 
     expect(emptySearch).toHaveBeenCalledWith('anything', expect.anything(), agentDir, 10, expect.any(Function))
     expect(retrievalContext.results).toContain('# Memory')
-    expect(retrievalContext.results).toContain('## First Topic')
-    expect(retrievalContext.results).toContain('slug: `first-topic`')
-    expect(retrievalContext.results).toContain('## Second Topic')
-    expect(retrievalContext.results).toContain('slug: `second-topic`')
+    expect(retrievalContext.results).toContain('- First Topic `first-topic`')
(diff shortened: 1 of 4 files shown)


devxoul:
Addressed in 06b6f216. The channel direct-mode (under-budget) path now routes through `renderTopicIndexMemorySection(plan.shards)` instead of `renderMemorySection`, so it emits the same one-line `- Heading `'`slug`'` bullets (verbatim slug, no `cites=/days=` metadata, channel boundary preserved) as the other channel paths. Dropped the now-unused `forceIndexForChannel`/`renderMemorySection` imports and updated the channel direct-mode test to assert the bullets and no metadata block. Thanks for the catch.

PR #951, opened 2026-06-18
feat(git): relocate agent git DB to .gitstore for monorepo fleet support

devxoul:
## Background

Today every typeclaw agent owns its own `.git` directory. That means N agents = N GitHub repos. A user running a fleet of domain-specific agents ends up with a cluttered GitHub account and no single place to browse the full agent tree.

The end goal is to let N agents live as plain subfolders inside one parent monorepo with one remote and a browseable directory tree. The obstacle: when Git sees a subdirectory that contains a `.git/` folder, it treats it as a gitlink (submodule) rather than plain files. To break that association the agent's git database must live somewhere other than `.git/`.

The chosen location is `.gitstore/`. A directory named `.gitstore` is invisible to a parent repo — no gitlink, no submodule — but it still lives inside the agent folder's Docker bind mount (`/agent`), so in-container committers can reach it without a host-side commit broker or a cross-container shared-index. Switching between layouts is a pure directory rename — no database surgery needed.

## Summary

**`resolveAgentGit`** in `src/git/resolve-agent-git.ts` is the single source of truth for an agent's git layout. It probes the agent directory at call time:

- `.git` present → returns empty args `[]` (no extra flags; identical to today's behavior)
- `.gitstore` present → returns `["--git-dir", ".gitstore", "--work-tree", "."]`
- neither → returns `null` (callers no-op, same as before)

`.git` takes precedence over `.gitstore`, so any existing agent is byte-for-byte identical to today.

`core.worktree` is intentionally NOT persisted. It is absolute and would break across the host path vs the container path (`/agent`). The `--work-tree` flag is injected per-call instead.

**Every git committer** now resolves layout and threads those args into each invocation:

- backup runner: commit, push, fetch, rebase
- dreaming memory snapshot: commit, log readers, `rev-parse --show-toplevel`
- `system-commit` (async + sync variants)
- `reconcile-ignored` untrack plumbing
- doctor auto-fix commit
- session-start git-nudge

`src/dreams/git.ts` previously ran `rev-parse --show-toplevel` unguarded. For a `.gitstore` agent inside a parent monorepo that call would walk UP and read the parent repo's history. It now returns the agent dir directly for `.gitstore` layouts and threads git args through all dream log/show readers.

A guard test in `src/git/no-bare-git-spawn.test.ts` asserts that each retrofitted committer file imports `resolveAgentGit`, making it a compile-time backstop against reintroducing a layout-blind bare git spawn.

**`.gitignore`** — `.gitstore/` is added to the generated `.gitignore` (local-only, never tracked by the agent itself).

**`compose.monorepo`** — a boolean added to the config schema as a host-stage scaffolding hint. Classified `'ignored'` in `FIELD_EFFECTS`: the container never reads it; the on-disk `.gitstore` directory is the runtime signal. It exists so `typeclaw compose` commands (follow-up PRs) have a typed field to set when adopting an agent into a monorepo.

## What this does NOT do

- `typeclaw compose init` / `adopt` / `eject` commands — follow-up.
- Parent-monorepo `.gitignore` generation — follow-up.
- `discoverAgents` scanning for nested `agents/*` layout — follow-up.
- `compose doctor` layout cross-checks — follow-up.
- Container env-var injection for explicit fleet awareness — follow-up.
- Any behavior change for standalone `.git` agents — resolver returns empty args; all git calls are byte-identical.

## Review notes

**Load-bearing invariant:** `resolveAgentGit` returns `[]` for `.git` and `null` for neither. The empty-args case means the caller's git invocation gets no extra flags — the only change in the hot path is one `fs.existsSync` probe per call. The `null` case short-circuits the same way callers already handled the "no git" case.

**`rev-parse --show-toplevel` in `src/dreams/git.ts`:** This was the highest-risk unguarded call. For a `.gitstore` agent inside a parent monorepo, that command walks UP the tree and returns the parent repo root, causing dream log readers to find no relevant commits. The fix returns `cwd` directly for `.gitstore` layouts and threads the resolved git args through `git log` / `git show` in the dreaming path.

**Why `.gitstore` and not `.git-store` or `_git`?** `.gitstore` is dot-prefixed (hidden on Unix), unlikely to collide with real project files, and unambiguous in path-globs. It is easy to explain: "the store for git data."

**`compose.monorepo` as `'ignored'`:** The container must never branch on this field — doing so would couple the runtime to a host-stage concept. The `.gitstore` presence check is the runtime signal; `compose.monorepo` is the host-side scaffolding flag. Classified `'ignored'` so `getConfig()` live-reload cycles don't react to it.

**Guard test:** `no-bare-git-spawn.test.ts` checks that each committer module imports `resolveAgentGit`. This is a structural/import-level check — it exists solely to ensure a future refactor doesn't silently drop the resolver and reintroduce a layout-blind spawn.

**Multi-language test coverage:** committer tests include Korean memory and commit content per the repo's multi-language testing rule.

Full suite: 9342 pass / 0 fail. `bun run typecheck` clean. `bun run lint` 0 errors (65 pre-existing warnings, none in changed files). `bun run format:check` clean.

### DIFF
--- a/src/agent/git-nudge.test.ts
+++ b/src/agent/git-nudge.test.ts
@@ -106,6 +106,22 @@ describe('renderGitNudge', () => {
     expect(text).toContain('- src/bar.ts')
   })
 
+  test('passes gitstore args to status reader when .gitstore is present', async () => {
+    await mkdir(join(agentDir, '.gitstore'))
+    let captured: readonly string[] = []
+    const deps: GitNudgeDeps = {
+      readStatus: async (_agentDir, gitArgs) => {
+        captured = gitArgs
+        return ['src/foo.ts']
+      },
+    }
+
+    const text = await renderGitNudge(agentDir, deps)
+
+    expect(text).toContain('- src/foo.ts')
+    expect(captured).toEqual(['--git-dir', join(agentDir, '.gitstore'), '--work-tree', agentDir])
+  })
+
   test('drops paths under sessions/ and memory/ from the nudge', async () => {
     await mkdir(join(agentDir, '.git'))
     const deps: GitNudgeDeps = {
--- a/src/agent/git-nudge.ts
+++ b/src/agent/git-nudge.ts
@@ -1,5 +1,4 @@
-import { existsSync } from 'node:fs'
-import { join } from 'node:path'
+import { resolveAgentGit } from '@/git/resolve-agent-git'
 
 const MAX_LISTED_PATHS = 10
 
@@ -9,15 +8,16 @@ const MAX_LISTED_PATHS = 10
 const RUNTIME_OWNED_PREFIXES = ['sessions/', 'memory/']
 
 export type GitNudgeDeps = {
-  readStatus: (agentDir: string) => Promise<readonly string[] | null>
+  readStatus: (agentDir: string, gitArgs: readonly string[]) => Promise<readonly string[] | null>
 }
 
 // Returns "" (not a placeholder string) when there is nothing to nudge about.
 // The empty case must add zero bytes to the system prompt so cache prefixes
 // stay identical to a clean-worktree agent folder.
 export async function renderGitNudge(agentDir: string, deps: GitNudgeDeps = defaultDeps): Promise<string> {
-  if (!existsSync(join(agentDir, '.git'))) return ''
-  const status = await deps.readStatus(agentDir)
+  const repo = resolveAgentGit(agentDir)
+  if (!repo) return ''
+  const status = await deps.readStatus(agentDir, repo.gitArgs)
   if (status === null) return ''
   const dirty = filterAgentOwned(status)
   if (dirty.length === 0) return ''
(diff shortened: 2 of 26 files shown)


devxoul:
Both findings addressed in cb7ad909.

- **`backup/index.ts` push-url resolver** — `resolveOriginPushUrl` now resolves the layout via `resolveAgentGit(cwd)` and threads `--git-dir`/`--work-tree` into `git remote get-url --push origin`, so a relocated `.gitstore` agent inside a parent monorepo reads its own `origin` (and `resolveBackupPushAuthEnv` mints for the agent's slug), not the parent's. Added a real-git integration test that sets up a parent repo (`origin = .../fleet.git`) with a `.gitstore` child (`origin = .../alice.git`) and asserts the resolved URL is `alice.git`.
- **`guard/policies/uncommitted-changes.ts`** — replaced the `.git`-only guard with `resolveAgentGit(agentDir)` and threaded `gitArgs` into `readStatus`'s `git status`, so `.gitstore` agents keep warning on dirty user-owned files. Added a `.gitstore` test mirroring the git-nudge coverage.

Also extended the `no-bare-git-spawn` guard test to include both files so a future change can't reintroduce a layout-blind spawn here.

All checks green (typecheck, lint 0 errors, format, full suite 9346 pass / 0 fail).

PR #955, opened 2026-06-18
fix(init): pin bson for Bun hatching

devxoul:
## Background

Fresh `typeclaw init` hatching can build the agent image successfully but then fail before the websocket server binds. The attached failure shows the container exiting during startup with Bun throwing `ERR_NOT_IMPLEMENTED` from `node:v8` when `bson@7.2.0` imports and calls `startupSnapshot.isBuildingSnapshot()` at module initialization time. That `bson` version arrives transitively through `agent-messenger`, so a newly-scaffolded Discord agent can crash before TypeClaw has a chance to report anything beyond container logs.

## Summary

Pin `bson` to `6.10.4` through Bun `overrides` in scaffolded agent `package.json` files. That keeps freshly-created agents on the last BSON 6.x line that does not hit Bun's unimplemented `node:v8` snapshot probe.

The package refresh path now also adds the same override to existing agents, preserving any user-defined overrides, so `typeclaw start` can repair already-scaffolded folders after upgrading TypeClaw.

## What this does NOT do

This does not change `agent-messenger` itself or remove the override automatically later. Once upstream no longer imports the Bun-incompatible BSON startup probe, this pin can be removed in a follow-up.

## Review notes

Load-bearing invariant: both greenfield scaffold and existing-agent package refresh must include `overrides.bson = "6.10.4"` without clobbering existing `overrides` entries.

Validation run locally:

- `./node_modules/.bin/oxfmt --check src/init/index.ts src/init/packagejson.ts src/init/index.test.ts src/init/packagejson.test.ts package.json`
- `bun test --parallel src/init/packagejson.test.ts src/init/index.test.ts --test-name-pattern 'overrides|Bun-compatible'`
- `bun test --parallel src/init/packagejson.test.ts`
- `bun run typecheck`
- `bun run lint` exits 0, with pre-existing warnings unrelated to this change

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,9 @@
   "peerDependencies": {
     "typescript": "^5"
   },
+  "overrides": {
+    "bson": "6.10.4"
+  },
   "engines": {
     "bun": ">=1.1.0"
   }
--- a/src/channels/router.test.ts
+++ b/src/channels/router.test.ts
@@ -7099,20 +7099,17 @@ describe('ChannelRouter plugin lifecycle hooks', () => {
 
     // when a first message engages the bot and a second arrives after the
     // idle-hook watchdog should have fired
-    const start = Date.now()
     await router.route(inbound({ text: 'first' }))
     await router.__testing!.flushDebounce(KEY)
     await router.route(inbound({ externalMessageId: 'm2', text: 'second' }))
     await router.__testing!.flushDebounce(KEY)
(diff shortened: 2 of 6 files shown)


devxoul:
Closing this TypeClaw-side BSON override workaround in favor of the cleaner dependency bump in #958. agent-messenger@2.20.5 now owns the BSON pin, so TypeClaw only needs to consume the fixed release.

PR #967, opened 2026-06-20
secrets: emit codex auth.json id_token so the CLI loads

devxoul:
## Background

Every agent configured with `docker.file.codexCli: true` and an `openai-codex` OAuth model backend has a **broken Codex CLI**. Any `codex` invocation in the container dies before the model call with:

```
missing field `id_token` at line 6 column 3
```

Reproduced on two live agents — codex `0.140.0` and `0.141.0` — both with the typeclaw-emitted `~/.codex/auth.json`. `codex exec` and even `codex login status` fail identically, so this is an auth-file load failure, not a model/runtime issue.

**Root cause.** Codex's auth.json deserializer (`codex-rs/login/src/token_data.rs`) declares:

```rust
pub struct TokenData {
    #[serde(deserialize_with = "deserialize_id_token", ...)]
    pub id_token: IdTokenInfo,   // non-Option, no #[serde(default)]
    pub access_token: String,
    pub refresh_token: String,
    pub account_id: Option<String>,
}
```

`id_token` is required. typeclaw's auto-export (`emitCodexAuthJson`) wrote a `tokens` object containing only `access_token` / `refresh_token` / `account_id`, so serde rejects the whole file before codex ever reads a token.

The credential never carried an `id_token` to emit: pi-ai's OpenAI Codex OAuth flow requests `openid` scope and `id_token_add_organizations=true`, but its `readTokenResponse` reads only `access_token` / `refresh_token` / `expires_in` and discards the `id_token` OpenAI returns. Confirmed still absent through `@mariozechner/pi-ai@0.73.1` and the renamed `@earendil-works/pi-ai@0.79.8` — **a dependency bump does not fix this.**

## Summary

`emitCodexAuthJson` now writes `tokens.id_token`, sourced in this order:

1. **A real captured id_token** — `idToken` / `id_token` on the stored credential. Forward-compatible: if pi-ai (or our own capture) starts persisting it, we emit the real value with zero further changes. The credential schema already accepts it (`oauthProviderSchema` is `.catchall(z.unknown())`).
2. **Fallback: the `access` token.** In this OAuth flow the access token is itself a ChatGPT-claims JWT (it carries the `https://api.openai.com/auth` block). Codex parses `id_token` via `parse_chatgpt_jwt_claims`, which accepts **any** decodable 3-part JWT and treats every claim as `Option`/`#[serde(default)]` — so the access JWT satisfies the deserializer and even populates the same plan/account fields.

**Why the fallback and not a re-login flow:** the fallback self-heals every existing agent on the next container start — no re-paste, no `codex login`, no pi-ai bump. It stays within codex's own accepted contract (codex never validates that `id_token` claims are present), so it is not a hack that a codex update would break.

## What this does NOT do

- **Does not patch pi-ai.** The upstream omission is real but lives in a third-party package; recovering the true `id_token` upstream is a separate change. This fix makes typeclaw correct regardless.
- **Does not touch the newer-wins exporter, the symlink/atomic-write path, or the guard chain.** Only the emitted `tokens` shape changes.
- **No schema migration.** No stored credential is rewritten; the field is added at emit time.
- **No api-key path changes.**

## Review notes

- Load-bearing change is the `id_token` source resolution in `emitCodexAuthJson`. The invariant to verify: the emitted `tokens` always contains a non-empty `id_token` that is a decodable JWT (guaranteed because `access` is validated non-empty and is a JWT in this flow).
- The `firstNonEmptyString(idToken, id_token)` helper handles both camelCase and snake_case so a future pi-ai using either key Just Works.
- Tests cover: default emit includes `id_token`; real captured id_token wins (camelCase and snake_case); empty captured id_token falls back to access; the existing B1 export test asserts the new shape.

### DIFF
--- a/src/secrets/codex-auth-json.test.ts
+++ b/src/secrets/codex-auth-json.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, test } from 'bun:test'
 
-import { decodeCodexAccessTokenExpiryMs, emitCodexAuthJson } from './codex-auth-json'
+import { decodeCodexAccessTokenExpiryMs, emitCodexAuthJson, isDecodableJwt } from './codex-auth-json'
 
 function makeJwt(payload: Record<string, unknown>): string {
   const header = Buffer.from(JSON.stringify({ alg: 'RS256', typ: 'JWT' })).toString('base64url')
@@ -9,7 +9,7 @@ function makeJwt(payload: Record<string, unknown>): string {
 }
 
 describe('emitCodexAuthJson', () => {
-  test('emits the modern tokens-only shape with trailing newline', () => {
+  test('emits the tokens shape with id_token and a trailing newline', () => {
     const out = emitCodexAuthJson({
       type: 'oauth',
       access: 'access-1',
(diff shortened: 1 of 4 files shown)


devxoul:
Addressed in the latest push (squashed into the single commit per repo convention).

The blocking issue is fixed: `shouldOverwrite` now treats an on-disk `tokens` object without a non-empty `id_token` as unusable and forces the rewrite, gated **before** the newer-wins expiry compare. So a pre-fix TypeClaw-emitted `auth.json` (valid/fresher access token, no `id_token`) no longer returns `on-disk-is-fresher` — `emitCodexAuthJson` runs and codex is repaired on restart.

Added the requested regression test: a tokens-only `auth.json` whose access expiry beats the stored credential but lacks `id_token` → asserts `wrote` and a non-empty emitted `id_token`. Existing newer-wins skip fixtures were updated to include `id_token` so they exercise the real codex-loadable path.

devxoul:
Fixed in the latest push (squashed into the single commit).

`onDiskHasUsableIdToken` now requires the on-disk `tokens.id_token` to be a **decodable JWT**, not just a non-empty string. Factored the JWT-payload decode out of `decodeCodexAccessTokenExpiryMs` into a shared `decodeJwtPayload` helper and exposed `isDecodableJwt` — the same check (3 non-empty parts, base64url payload that is a JSON object) codex's `decode_jwt_payload` runs on `id_token`. So a file like `{ tokens: { id_token: 'not-a-jwt', access_token: <fresh JWT>, ... } }` is now treated as stale and rewritten before the newer-wins compare.

Regression tests added: (1) exporter — fresh/tied access token + non-empty non-JWT `id_token` → `wrote`; (2) unit — `isDecodableJwt` true for a JWT without `exp`, false for non-JWT / wrong part count / non-base64 / non-object payload.

PR #975, opened 2026-06-21
channels: decode webex ids in the router log lines

devxoul:
## Background

Follow-up to #969/#972/#973. Those PRs made Webex's opaque base64 REST ids readable on human-facing surfaces — the adapter log lines (#972, `webex.ts`/`webex-bot.ts`), the inspect renderer, and the system-prompt conversation line (#973). But a whole class of operator log output was missed.

The **adapter-agnostic router** (`src/channels/router.ts`) logs `channelKeyId(key)` = `${adapter}:${workspace}:${chat}:${thread}`. For Webex, `chat` is the raw base64 `ciscospark://` blob. The adapter PRs only touched the webex-specific adapter files; the router never decoded because it has no adapter-specific code path. So `typeclaw logs` still showed:

```
[channels] webex:@dm:Y2lzY29zcGFyazovL3VzL1JPT00vZmY3YTI1MjAtNmMwNC0xMWYxLThkZTEtOTdkMDcwMGM5NDU1:: ensureLive begin (cold-start)
[channels] webex:@dm:Y2lz...:: prompting batch=1 text_len=1350
```

That blob decodes to `ciscospark://us/ROOM/<uuid>`. It leaks into ~85 router log lines (every `ensureLive` phase, `prompting`/`prompted`, `send`, the empty-turn/suppression/recovery family, `stale-rollover`, command denials, restart-resume, sticky-clear) plus the `membership-cache` resolver warn and two `engage-react` lines.

## Summary

A **display-only** decode, mirroring the established `toRef`/`readableId` pattern:

- New `readableChannelId(adapter, id)` in `webex-id-ref.ts` — gates the decode to `webex`/`webex-bot` (the only adapters whose ids are opaque blobs) and falls open for everyone else. This centralizes the `WEBEX_ADAPTERS` + `toRef` pattern that already existed inline in `inspect/label.ts` and `agent/session-origin.ts`.
- New `channelKeyLabel(key)` in `types.ts` — a display sibling of `channelKeyId` with identical shape, but the `chat` component decoded.
- `LiveSession` gets a `keyLabel` field computed **once** at session creation (no per-log-line re-decode). Non-live scopes (command / restart-resume / clearSticky / membership-cache) compute a local `keyLabel` alongside their existing `keyId`.
- Every log interpolation swapped to the label.

## What this does NOT do

Does not change `keyId` semantics, session lookup, persistence, or any return value from `clearSticky` / `delivered` / `recorded`. Does not affect non-Webex adapters (Slack, Discord, Telegram, KakaoTalk ids pass through unchanged).

## Review notes

Start with the test `webex log lines decode the base64 room id while keyId stays the raw rest id` — it asserts both halves in one shot: the decoded room ref appears in the logs, the raw base64 blob does **not**, and `clearSticky(...).keyId` is still the canonical raw REST id.

`bun run lint` (0 errors; 65 pre-existing warnings in untouched test files), `format:check` (clean), `typecheck` (clean), `bun run test` (9525 pass / 0 fail).

### DIFF
--- a/src/channels/adapters/webex-id-ref.ts
+++ b/src/channels/adapters/webex-id-ref.ts
@@ -1,6 +1,8 @@
 import { decodeWebexId } from 'agent-messenger/webex'
 import type { WebexRestIdType } from 'agent-messenger/webex'
 
+import type { AdapterId } from '../schema'
+
 // Webex REST ids are base64url of `ciscospark://<cluster>/<TYPE>/<uuid>`. The
 // opaque blob is unreadable in `inspect` and — worst — in hand-authored
 // permission rules (`webex:* author:Y2lzY29zcGFyazov...`). A `ref` is the
@@ -29,4 +31,12 @@ export function toRef(id: string): string {
 export function isWebexIdOfType(id: string, type: WebexRestIdType): boolean {
   const decoded = decodeWebexId(id)
   return decoded !== null && decoded.type === type
+}
+
+// Gate the decode on adapter: only Webex ids are opaque base64 blobs, and
+// `toRef` would otherwise run its fall-open path over ids that can never decode.
+const WEBEX_ADAPTERS = new Set<AdapterId>(['webex', 'webex-bot'])
+
+export function readableChannelId(adapter: AdapterId, id: string): string {
+  return WEBEX_ADAPTERS.has(adapter) ? toRef(id) : id
 }
--- a/src/channels/membership-cache.ts
+++ b/src/channels/membership-cache.ts
@@ -8,7 +8,7 @@ import {
   type MembershipResolverResult,
 } from './membership'
 import type { ChannelKey } from './types'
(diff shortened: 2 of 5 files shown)


devxoul:
Superseded by a deeper fix. #975 decoded the base64 Webex room id only at the router log layer (display-only), leaving the opaque `ciscospark://` blob as the stored ChannelKey identity. Discussion on this PR concluded the blob-as-stored-id is itself the root cause — it forces ~95 scattered hand-decode sites across logs, inspect, the system prompt, permissions, and role-claim.

The replacement stores the decoded UUID ref as the canonical id end-to-end (agent-messenger now provides `ref` siblings on every id-bearing surface and already re-encodes bare uuids on outbound calls), with a v4→v5 persisted-session migration and a kept fail-closed permission compat layer. New PR incoming on branch `channels/webex-store-uuid-refs`.

PR #992, opened 2026-06-22
feat(subagent): add write-capable `deep` worker on the deep model profile

devxoul:
## Why

The user wants **fast conversational responsiveness** from a lighter default model, while still getting strong reasoning on hard execution work. The lever for the heavy half is a worker that runs on the `deep` model profile.

Investigation (two parallel explores over the subagent system) confirmed: **`operator` already is the build/run/fix/verify worker** — full write+bash tools, "verify after each step / recover from failures" prompt. The only gap vs. the goal is the **model profile**: `operator` runs `default`; heavy work wants `deep`. So this adds a sibling, not a redundant agent.

## What

A new public bundled subagent **`deep`** — the quality-over-speed counterpart to `operator`:

- Same write-capable tool set (`read/grep/find/ls/bash/write/edit`).
- Binds `profile: 'deep'` instead of `'default'`.
- Prompt is tuned for *careful* work ("understand before acting", "diagnose root cause, don't paper over the symptom") rather than operator's fast-execution framing.

The intended deployment shape: point `models.default` at a **lighter/faster** model (keeps the chat responsive) and let the agent **delegate gnarly build/run/fix/verify work to `deep`** on a stronger model. The orchestration prompt's **Mode B** now names that routing rule explicitly (operator = fast/default, deep = quality/deep).

## Security (load-bearing)

`deep` is write-capable, so it reuses `operator`'s gating model exactly:

- `requiresSpecificPermission: true` → gated on `subagent.spawn.deep` (derived from the registry name).
- `subagent.spawn.deep` granted to **owner + trusted only**, never `member` — mirroring `subagent.spawn.operator`.

Replicating operator's tool power without its permission gate would be a privilege-escalation hole; this avoids it, and a guard test asserts `member` does **not** carry the permission.

## Scope notes

- **No model config change.** `models.default` (the lighter model) is a user choice in `typeclaw.json` and is intentionally **not** touched here — the repo doesn't hardcode users' model picks. Recommended config is documented here, not shipped.
- Built on existing machinery: `profile: 'deep'` already exists (used by `reviewer`/`researcher`/`planner`); subagents are bundled plugins registered via `definePlugin`.

## Commits (atomic, split per the repo's commit guard)

1. `feat(subagent): define a write-capable deep worker on the deep model profile` — the module + plugin wrapper + tests, self-contained.
2. `feat(permissions): grant subagent.spawn.deep to owner and trusted` — permission policy.
3. `feat(subagent): register deep in the roster and route to it from Mode B` — registration, roster/order guards, prompt routing note.

## Tests (TDD, written failing first)

- `src/bundled-plugins/deep/deep.test.ts` — profile=`deep`, public, gated, write-capable tool set, roster description, payload schema, prompt phrases.
- Extended guards: roster drift fixture, `BUILTIN_ROLES` owner/trusted/member assertions, bundled-plugins order guard.

```
bun run typecheck     # clean
bun run lint          # 0 errors (pre-existing warnings only)
bun run format:check  # clean
bun run test          # 9578 pass / 0 fail
```

## Relationship to #989

Orthogonal. #989 is universal prompt steering (completion / anti-fabrication / parallel tool calls); this is a new delegation target. Both stay open on separate branches off `main`.


devxoul:
Superseded: instead of a separate `deep` subagent (which duplicates operator's prompt + tools to vary only the model tier), the better factoring is a per-spawn model-tier override on `operator` — model tier is a routing decision that belongs at the call site, not baked into a second agent's identity. Closing in favor of that approach. Branch left intact for reference.

Issue #899, opened 2026-06-15
Native Windows support


devxoul:
## Goal

Make TypeClaw fully usable on **native Windows** — PowerShell/cmd + Docker Desktop, **without WSL2**. This issue captures a code-grounded feasibility audit + an architecture review so the work can be picked up in deliberate phases.

## TL;DR

- **No HARD blockers.** Every Windows limitation has an established workaround.
- This is a **host-stage-only** problem. The container is always Linux, so the scary Linux machinery (bwrap, seccomp, Xvfb, `/proc`, enforced `chmod`) is **out of scope** — it never runs on the host.
- The container→host control leg **already uses TCP** (`host.docker.internal`), which works on Docker Desktop for Windows today. There is **no** Windows-host→Linux-container Unix-socket mount to untangle.
- **WSL2 + Docker Desktop already works today** (`src/shared/wsl.ts` exists). Native Windows should be framed as a **phased, "experimental until proven"** project, not a quick port.
- Effort: **Large** — not from container internals, but from coordinated platform work on hostd lifecycle, IPC, Docker Desktop path semantics, and Windows process/spawn behavior.

## Why this is host-stage only

TypeClaw runs in three stages; only one touches Windows:

| Stage | OS | In scope? |
|---|---|---|
| **dev** (build/test this repo) | dev's choice | Bun supports Windows x64/arm64 — trivial |
| **container** (`typeclaw run`) | **always Linux** (Docker) | **No** — all Linux-only sandbox machinery lives here |
| **host** (`init`/`start`/`stop`/`restart`/`logs`/`tui`/`compose`/`_hostd`) | **the user's machine** | **Yes — the entire porting surface** |

## What already works on native Windows (verified, no change needed)

- **Container→host control is TCP, not a mounted Unix socket.** `src/container/start.ts` injects `TYPECLAW_HOSTD_URL=http://host.docker.internal:<port>` + `TYPECLAW_HOSTD_TOKEN` and adds `--add-host host.docker.internal:host-gateway`. The container-side client (`src/agent/restart/index.ts`) prefers `sendHttp()`. There is **no `/run/typeclaw-host` bind mount** in the run args — `containerSocketPath()` / `containerHostRunDir()` in `src/hostd/paths.ts` are **vestigial dead code**. Docker Desktop for Windows supports `--add-host ...:host-gateway`, so this leg is already compatible.
[...]

PR #1010, opened 2026-06-22
fix(init): use --backend=copyfile for bun install on Windows

devxoul:
## Background

`typeclaw init` fails on native Windows during the dependency install step with:

```
EPERM: failed copying files from cache to destination for package typeclaw
```

Verbose `bun install` output pins the mechanism:

```
warn: CreateHardLinkW failed, falling back to CopyFileW:
  \\?\C:\Users\<user>\workspace\typeclaw\.git\config
  - \\?\C:\Users\<user>\typeclaw\<agent>\node_modules\typeclaw\.git\config
EPERM: failed copying files from cache to destination for package typeclaw
```

Root cause is in Bun, not typeclaw. Bun's default install backend on Windows is `hardlink` (`CreateHardLinkW`). When a hardlink fails — locked files, antivirus/Defender real-time scanning of the cache or destination, cross-context copies — Bun's per-file path falls back to `CopyFileW`; but when that *also* fails it aborts the whole install. Bun's outer hardlink→copyfile fallback is gated behind `#[cfg(not(windows))]`, so Windows has no recovery path and surfaces the EPERM.

This hits the `bun link` dev workflow hardest: in that mode the agent's `package.json` declares a `file:` typeclaw dep, so Bun copies the entire source tree — including `.git/` — into `<agent>/node_modules/typeclaw`, and the `.git` internals are exactly what Defender tends to lock. End users installing from npm (`^X.Y.Z` registry spec) hit it less often but the same hardlink-permission failure mode applies to any package.

## Summary

Pass `--backend=copyfile` to `bun install` / `bun update` on Windows so Bun skips `CreateHardLinkW` and copies directly via `CopyFileW`, sidestepping the hardlink-permission failure that has no fallback on Windows.

- **Why `copyfile` and not "run as admin" / antivirus exclusions:** those are per-machine manual workarounds that mask the issue; selecting a backend that doesn't depend on hardlinks fixes it for every Windows user with no setup.
- **Why gate on platform:** POSIX keeps Bun's faster defaults (hardlink on Linux, clonefile on macOS) — there's no reason to slow those down.
- The backend selection lives in one helper (`bunInstallBackendArgs`) threaded through both `runBunInstall` and `runBunUpdate`, with a `platform` test seam mirroring the existing `isWindows(platform)` convention in `src/shared/platform.ts`.

This continues the recent Windows-init fix series (`resolve docker via Bun.which`, `fall back to Docker Desktop install paths`).

## What this does NOT do

- Does not change install behavior on Linux or macOS — argv is byte-identical there.
- Does not add retry-on-EPERM, cache clearing (`bun pm cache rm`), or `BUN_INSTALL_CACHE_DIR` redirection. Those are complementary mitigations but out of scope; `--backend=copyfile` addresses the specific hardlink-fallback gap.
- Does not address the deferred native-Windows dev-mode symlink/junction work tracked in #899; it only makes the install copy succeed.

## Review notes

- Load-bearing change: `bunInstallBackendArgs(platform)` returns `['--backend=copyfile']` only on `win32`, `[]` elsewhere. Verify argv ordering — on Windows with `force`, the backend flag must precede `--force` (asserted in the new tests).
- The new test helper `capturingSpawn()` records the argv handed to `spawn` but still returns a real fast-exiting Bun process, so the production `runTimedBunProcess` can `await proc.exited` / read `proc.stderr` without hanging.

### DIFF
--- a/src/init/run-bun-install.test.ts
+++ b/src/init/run-bun-install.test.ts
@@ -5,6 +5,17 @@ import { join } from 'node:path'
 
 import { runBunInstall, runBunUpdate } from './run-bun-install'
 
+// Capture the argv a runner hands to `spawn` while still returning a real,
+// fast-exiting process so `runTimedBunProcess` can await `exited`/`stderr`.
+function capturingSpawn(): { spawn: typeof Bun.spawn; cmds: string[][] } {
+  const cmds: string[][] = []
+  const spawn: typeof Bun.spawn = (options) => {
+    cmds.push([...((options as { cmd: string[] }).cmd ?? [])])
+    return Bun.spawn({ cmd: ['bun', '-e', ''], stdout: 'pipe', stderr: 'pipe' })
+  }
+  return { spawn, cmds }
+}
+
 describe('runBunInstall', () => {
   test('times out a hung install process', async () => {
     const cwd = await mkdtemp(join(tmpdir(), 'tc-bun-install-timeout-'))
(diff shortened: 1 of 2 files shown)


devxoul:
Closing: this fix doesn't actually resolve the reported Windows EPERM. Bun's hardlink-backend per-file fallback and the copyfile backend call the identical `CopyFileW(src, dest, 0)` syscall (src/install/PackageInstall.rs) — and the user's verbose log shows that same CopyFileW already failing on `.git/config`. Forcing `--backend=copyfile` only skips the CreateHardLinkW attempt; it then calls the same CopyFileW that already threw EPERM. The real root cause is that the `file:` typeclaw dep is going through Bun's cache-copy path (copying the whole source tree including `.git/`) instead of resolving as a Folder->symlink. Will reopen with a correct fix once the right approach is verified locally.

devxoul:
Superseding this. `--backend=copyfile` does not fix the reported EPERM: Bun's hardlink-fallback path and the copyfile backend call the identical `CopyFileW(src,dest,0)` syscall, and the user's verbose log shows that same CopyFileW already failing on `.git/config`. Root cause is dev-mode `file:` resolution copying the whole checkout (incl `.git/`) into node_modules on native Windows. The real fix is a typeclaw-controlled native-Windows dev-mode path (host-side junction for `node_modules/typeclaw` + skip `bun install` + container bind-mount over `/agent/node_modules/typeclaw`), landing in a new PR that also addresses #899. Closing in favor of that.

PR #1017, opened 2026-06-22
fix(container): resolve docker binary for `logs` like start/init

devxoul:
## Background

`typeclaw logs -f` reported "Executable not found in \$PATH: docker" on a Windows machine where Docker Desktop was installed and the engine was running. The observed symptom:

```
typeclaw logs -f
Streaming container logs...
✖ Executable not found in $PATH: "docker"
```

This is the same stale-PATH failure mode that #1005 / #1007 already fixed for `init` and `start`. Docker Desktop's installer registers `docker.exe` on the SYSTEM PATH, so a shell or process started before the install (or with a customized PATH) inherits a PATH where `Bun.which` — and therefore a bare `'docker'` head passed to `Bun.spawn()` — cannot see the binary, even though docker exists at an absolute path and the daemon is up.

The `logs` path never received the fix: `buildDockerLogsCmd` built its argv with a literal `'docker'` head, and both `container/logs.ts` and `compose/logs.ts` spawned it directly, bypassing the `resolveDockerBinary()` finder that `start`, `init`, and `shell` already use.

## Summary

Route `logs` and `compose logs` through the same `resolveDockerBinary()` finder that #1007 added the Windows Docker Desktop install-location fallback to. `buildDockerLogsCmd` now takes the already-resolved absolute binary path as its argv head instead of a literal `'docker'`.

**Why keep it a pure argv-builder rather than have it call the resolver internally:** it preserves precise `Bun.spawn` return-type inference at the compose call site and lets the unit tests assert the argv shape without a real docker install.

Both call sites resolve the binary up front and surface a clear "Docker is not installed" error when docker is genuinely absent — mirroring `shell.ts`. Resolution happens before the `containerExists` probe on purpose: `containerExists` collapses a missing docker binary into `false`, so resolving later would mask the real cause behind the misleading "Container … not found".

## What this does NOT do

- Does not change `resolveDockerBinary` or the Windows fallback locations (that is #1007).
- Does not change the `docker logs` argv shape, timestamp reformatting, or the esc-to-exit abort handling.
- Does not change POSIX behavior: when `Bun.which` finds docker (the normal Linux/macOS case), the resolved path is exactly what the bare name would have spawned.

## Review notes

- Load-bearing ordering in `container/logs.ts`: resolve docker → null check → `containerExists`. Reversing it reintroduces the "Container not found" masking bug.
- In `compose/logs.ts` the `dockerBinary === null` branch is effectively unreachable (every attached agent already passed `containerExists`, which is impossible without a working docker), but it is handled as a hard error rather than spawning a bare `'docker'`. A comment documents this so it is not removed as dead code.
- Pre-existing schema-drift test failures on `main` (typeclaw.schema.json, cron.schema.json, secrets.schema.json, scaffold) are unrelated to this change. All checks relevant to this diff pass: typecheck, lint (0 errors), format:check, and the logs/compose/shell test suites.

### DIFF
--- a/src/compose/logs.ts
+++ b/src/compose/logs.ts
@@ -1,7 +1,7 @@
 import { buildDockerLogsCmd, containerExists } from '@/container'
 import { supportsColor } from '@/container/log-colors'
 import { makeLogTimestampReformatter, type TimestampReformatter } from '@/container/log-timestamps'
-import { getBun } from '@/container/shared'
+import { getBun, resolveDockerBinary } from '@/container/shared'
 
 import { discoverAgents, type AgentEntry } from './discover'
 
@@ -74,6 +74,17 @@ export async function composeLogs({
 }: ComposeLogsOptions): Promise<ComposeLogsResult> {
   const agents = discoverAgents(rootCwd)
 
+  // Resolve docker BEFORE probing containers: containerExists collapses a
+  // missing docker binary into `false`, so every agent would be classified
+  // "container not running" and the empty-attached early return would exit 0 —
+  // masking the real "docker not on PATH" cause (the stale-Windows-PATH case
+  // #1007 covers for start/init). Resolving first surfaces it correctly.
+  const dockerBinary = resolveDockerBinary()
+  if (dockerBinary === null) {
+    err.write('compose: docker not found on PATH\n')
+    return { agents, attached: [], missing: agents, exitCode: 1 }
+  }
+
   const liveness = await Promise.all(
     agents.map(async (a) => ({ agent: a, exists: await containerExists(a.containerName) })),
   )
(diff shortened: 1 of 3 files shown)


devxoul:
Good catch — fixed. `composeLogs` now resolves docker **before** the `containerExists` liveness probe, so a missing docker binary surfaces as `compose: docker not found on PATH` with exit 1 instead of being masked: previously every agent's `containerExists` returned `false`, classified them all as "container not running", and the empty-`attached` early-return exited 0 — the exact masking this PR is about.

The single-container and compose paths now share the same invariant (resolve → null-check → probe), documented in a comment so a future reorder doesn't regress it.

PR #1060, opened 2026-06-24
fix(channels): recover empty stop after tool work instead of going silent

devxoul:
## Background

A channel turn where the model did tool work (e.g. `web_search`), then emitted a final assistant message with empty text and `stopReason: stop`, was silently discarded. The user sent a direct question in a group channel; the agent ran its research tools, then went quiet — dead air where an answer was expected.

The path through the router:

1. `recoverableAssistantText` returns `{ text: '', source: 'leaf' }` (not `null`) because the leaf block exists; it just has no text.
2. The `candidate === null` retry branch is skipped.
3. `endsWithNoReplySignal('')` returns `true` — an empty string is treated as a NO_REPLY signal.
4. All four existing empty-turn recovery guards miss this case (see Summary).
5. The turn falls through to a silent `no_reply`.

The question was a Korean-language message asking about research the agent had just looked up. It received no response.

## Summary

**Why the four existing guards all miss this case.**

- `isEmptyStopAfterWillingnessAck` — only fires when a send already landed this turn (it guards a stale willingness state after a visible reply). No send landed here.
- `leafIsStrandedToolUse` — only fires when a send already landed this turn for the same reason.
- `candidate === null` retry — the candidate is non-null (`{ text: '', source: 'leaf' }`), so this branch never enters.
- Cold-start-solo guard — only arms on the very first one-on-one turn; this was a group channel mid-conversation.

**The fix.**

A new guard is inserted in the `endsWithNoReplySignal` block, after the cold-start-solo guard and before the terminal `no_reply`. It fires when the current prompt-attempt made at least one tool call — the evidence that the model intended to gather and return information, not stay silent. On each retry the router posts `EMPTY_STOP_AFTER_TOOL_WORK_NUDGE`, which instructs the model to summarize the results it already gathered rather than re-run its tools. On exhaustion the visible fallback is posted so the user gets a response.

**`attemptMadeToolCall`** walks the `parentId` chain back to the first user message of the current attempt (which includes the injected retry nudge as a new user message, so each retry is scored on its own tool work independently). Tool-call detection uses the existing `AssistantMessage.toolCalls` array; no new I/O.

**Why a separate nudge constant and not `EMPTY_TURN_RETRY_NUDGE`.** `EMPTY_TURN_RETRY_NUDGE` was designed for the budget-exhaustion / no-tool-use path; reusing it here would tell the model to re-investigate (re-run tools) when it already ran them, and would misdiagnose a clean stop as budget exhaustion. The new `EMPTY_STOP_AFTER_TOOL_WORK_NUDGE` explicitly says "you've gathered results — please summarize them." The trailing `NO_REPLY` escape is preserved so a genuine research-then-decline can self-correct on the first retry rather than being forced to reply.

**The tool-work requirement is the decision threshold.** A bare-empty stop with no tool work stays silent (group-chat default, already handled by other paths). Tool work is the observable evidence that overrides that default.

Design was reviewed by Oracle before implementation.

## What this does NOT do

- Does not change how `recoverableAssistantText` classifies empty leaves — the source of truth stays `{ text: '', source: 'leaf' }`.
- Does not affect turns where the model returned non-empty text after tool work (already handled correctly).
- Does not affect the `candidate === null` retry path or the willingness-ack / stranded-toolUse degeneration paths.
- Does not change behavior for bare-empty stops with no preceding tool work (still silent — the group-chat default is intentional).
- Does not add new config surface; the guard is always-on within the `endsWithNoReplySignal` block.

## Review notes

- **Load-bearing location:** the new guard sits *after* the cold-start-solo guard and *before* the terminal `no_reply` in the `endsWithNoReplySignal` block. Order matters — cold-start-solo must remain the earlier check.
- **`attemptMadeToolCall`** uses `parentId` chain traversal capped at the first user message of the current attempt. Verify it does not walk past a retry boundary into a prior attempt's tool calls.
- **Log causes to look for:** `empty_stop_after_tool_work` (guard fires, retry posted) and `empty_stop_after_tool_work_retries_exhausted` (fallback posted). Absence of these in logs for an affected turn means the guard condition was not satisfied.
- **5 new tests** in `router.test.ts` cover: guard fires after one tool call; guard fires after multiple tool calls; guard does NOT fire on a bare-empty stop with no tool work; retries are capped at `MAX_EMPTY_STOP_RETRIES`; exhaustion posts the visible fallback. All 5 were verified via mutation (removing the guard causes each to fail). Full suite: 9973 pass / 0 fail. Typecheck, lint, and format:check all clean.

### DIFF
--- a/src/channels/router.test.ts
+++ b/src/channels/router.test.ts
@@ -21,6 +21,7 @@ import {
   createChannelRouter,
   disengageReactionEmojiFor,
   DUPLICATE_SEND_ERROR,
+  EMPTY_STOP_AFTER_TOOL_WORK_NUDGE,
   EMPTY_TURN_FALLBACK_TEXT,
   EMPTY_TURN_RETRY_NUDGE,
   extractPlainTextChannelToolCallText,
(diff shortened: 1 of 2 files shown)


devxoul:
Addressed in the latest push (rebased onto `main`).

You were right — the original guard required fresh tool work on *every* attempt, but the recovery nudge says "Do NOT re-run your tools," so a compliant retry lands a bare-empty `stop` with no new tool call and fell straight back to silent `no_reply` with the budget unspent. Self-defeating.

Fix: the classification now **latches** for the logical turn via a new `emptyStopAfterToolWorkArmed` flag (set the first time the shape is identified, reset alongside `emptyTurnRetries` on a real user batch). Once armed, subsequent bare-empty stops keep consuming the shared retry budget and eventually post `empty_stop_after_tool_work_retries_exhausted` instead of going silent. Explicit non-empty `NO_REPLY` still escapes (its `trim()` is non-empty, so the guard's first condition fails and it falls through to silence) — kept as the deliberate-decline hatch.

Tests updated:
- The test that previously codified the silence hole now asserts the armed turn spends the full budget to the visible fallback (mutation-verified: disabling the latch fails it).
- Added an explicit "armed turn + `NO_REPLY` escape → silent, no fallback" test.

`typecheck` / `lint` / `format:check` / full `bun run test` all green.

PR #1071, opened 2026-06-24
feat(cli): recommend Actions read/write in github channel setup

devxoul:
## Background

The interactive `typeclaw channel add github` flow prints a "Required permissions" checklist so operators provision the right GitHub App / fine-grained PAT scopes before pasting credentials. The list omitted the **Actions** permission. Without `Actions: Read and write`, an agent operating on a repo can't read workflow run status or re-run/dispatch workflows from the channel — a gap operators only discover at runtime when a call 403s, after setup is already "done".

## Summary

Added `Actions read/write` to the recommended-permissions guidance in all four places the string appears in `src/cli/channel.ts`:

- the shared "Get GitHub credentials" prompt (initial add, PAT or App),
- "Switch to GitHub App auth",
- "Rotate the GitHub PAT",
- "Switch to GitHub PAT auth".

Placed it as the first item after the event-family permissions (Issues / Pull requests / Discussions) and before the infrastructure ones (Metadata, Webhooks), keeping the existing ordering intent. Applied to the PAT blocks as well as the App blocks because `Actions` is a valid family for both fine-grained PATs and GitHub Apps, so the checklist stays consistent regardless of the auth type the operator picks.

## What this does NOT do

- Does not change runtime behavior, permission *enforcement*, or the 403 remediation guidance in `permission-guidance.ts` — this is operator-facing setup copy only.
- Does not add Actions to the `EVENT_PERMISSION_KEY` map or the auto-preflight gap check (no Actions-gated webhook event is in the default allowlist).
- Does not touch the docs site copy.

## Review notes

These are plain `note(...)` strings with no test coverage asserting on their content, so no tests changed. Pre-commit checks all pass: `typecheck` (0 errors), `lint` (0 errors), `format:check` (clean), full suite (10109 pass / 0 fail).

### DIFF
--- a/src/channels/adapters/github/permission-guidance.test.ts
+++ b/src/channels/adapters/github/permission-guidance.test.ts
@@ -4,6 +4,7 @@ import {
   buildAppPermissionPreflightGuidance,
   buildOutboundPermissionGuidance,
   buildPermissionGuidance,
+  githubRequiredPermissionsNote,
   isOutboundPermissionDenial,
   parseListHooksPermissionStatus,
 } from './permission-guidance'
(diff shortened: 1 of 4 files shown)


devxoul:
Good catch — addressed both points in the amended commit:

- **`src/cli/init.ts` `runGithubFlow`** now includes `Actions read/write` in the required-permissions note.
- **Extracted a single shared source of truth** to prevent future drift: `githubRequiredPermissionsNote(includeWebhookNote)` in `src/channels/adapters/github/permission-guidance.ts` (the existing owner of GitHub permission copy). All four `channel.ts` prompts and the one `init.ts` prompt now call it; `includeWebhookNote` toggles the "TypeClaw will create and manage the repository webhooks for you" reassurance shown on first-setup prompts only.
- Added two regression tests in `permission-guidance.test.ts` asserting both variants contain `Actions read/write` and that the webhook note toggles correctly.

Folded into the original commit and force-pushed.

PR #1095, opened 2026-06-26
feat(plugin): let plugins contribute MCP servers

devxoul:
## What

Adds `mcpServers` to the plugin contribution surface so plugins (bundled or third-party) can register **stdio/http MCP servers** that merge with `typeclaw.json#mcpServers` and connect through the existing `McpManager`. This is the platform groundwork; a follow-up PR ships first-party GLM vision as the first consumer.

## Why

MCP is currently config-only. Letting plugins contribute servers lets a capability ship as one unit (server + skill + doctor check + typed config + conditional activation), matching the "plugins are just imports" + "MCP support" extensibility story.

## Contract

- `PluginExports.mcpServers?: Record<string, PluginMcpServer>` where `PluginMcpServer` is a tagged transport union:
  - `{ type: 'stdio', command, args?, env? }` or `{ type: 'http', url, env? }`
- Server names must be `__`-free (the `server__tool` namespace splits on the **first** `__`, see `parseNamespacedTool`) and globally unique across config + all plugins; collisions are rejected.
- Exporting servers **requires declaring the `'mcp'` permission** (validated at registration: fatal for bundled, tolerated/skipped for user plugins).
- `PluginContext` gains `defaultProviderId` and `hasSecret(envName)` (presence-only — never exposes the value) so plugin factories can do provider/credential-conditional activation.

## Boot & failure semantics

- `McpManager` is now constructed **after** `loadPlugins`, merging `config.mcpServers` (which win on name collision) with normalized plugin servers.
- Spawn/connect failures stay **non-fatal** (warn only) — a bundled plugin whose MCP server can't start never aborts boot.
- Cleanup on boot failure moved into the process-globals disposer.

## Tests

New unit coverage: registry registration + name/`__`/collision rejection, the permission gate (with/without `'mcp'`), the `pluginMcpServersToConfig` mapper (stdio/http/defaults/env), the config-wins merge, and `hasSecret` presence semantics.

## Checks

`bun run typecheck` ✓ · `bun run lint` (0 errors, no new warnings) ✓ · `bun run format` ✓ · `bun run test` (10197 pass / 0 fail) ✓

No change to the existing `typeclaw.json#mcpServers` config schema or behavior.


devxoul:
@typeey review again please

PR #1123, opened 2026-07-02
Upgrade pi-coding-agent to 0.73 and own pi builtins

devxoul:
## Background

Our `@mariozechner/pi-coding-agent` / `@mariozechner/pi-tui` pin was stuck at `^0.67.3` while upstream shipped `0.73.1` — six minor releases behind, with the caret locking us to `0.67.x` so a plain install never advanced. Three of those releases were breaking in ways that touch our load-bearing tool wiring:

- **0.68.0** removed the bare builtin tool exports (`readTool`, `bashTool`, …) and, more importantly, changed `createAgentSession({ tools })` from "swap the builtin implementation" to "name allowlist only". Our whole hook/guard/sandbox coverage of `read/bash/edit/write/grep/find/ls` was built on the old semantics, routing wrapped implementations through `customTools` and then re-narrowing the active set after the fact with `setActiveToolsByName`.
- **0.69.0** migrated `@sinclair/typebox` 0.34 → `typebox` 1.x (a different package name).
- pi-tui **0.73** added members to the `Terminal` interface.

The upgrade is therefore not a version bump — it's a migration of the tool-selection layer that also happens to be a security boundary (per-subagent tool narrowing, the bwrap bash sandbox).

## Summary

The core move: **TypeClaw now owns every pi builtin as a wrapped `ToolDefinition`.** pi 0.68 ships `create*ToolDefinition(cwd, opts?)` factories, so the old `AgentTool → ToolDefinition` bridge collapses — `wrapSystemAgentTool` + `wrapAgentToolAsCustomToolDefinition` merge into a single `wrapBuiltinToolDefinition`, and `resolveBuiltinToolRefs` returns a flat `ToolDefinition[]` instead of the `{ agentTools, toolDefinitions }` split.

Key decisions a reviewer would otherwise have to ask about:

- **Why wrap builtins unconditionally now (not gated on plugin hooks)?** The bwrap sandbox and per-subagent bash policy are security behavior independent of whether any plugin registered a `tool.before/after` hook. Under the old code a hookless session ran pi's *raw* bash; gating the wrap on hooks would leave an unsandboxed bash path. So the wrapper is always the execution path; its optional stages (plugin hooks, permissions, bashPolicy) no-op when absent.
- **Why `noTools: "builtin"` + an explicit `tools:` allowlist instead of the post-hoc `setActiveToolsByName` re-narrow?** In 0.73, `tools:` *is* the active-set gate (`allowedToolNames` in pi's `_refreshToolRegistry`). Passing the intended active names at construction makes the widening branch dead, so the `setActiveToolsByName` workaround is provably redundant and is deleted. `noTools: "builtin"` disables pi's own unwrapped `read/bash/edit/write` so we don't rely on customTools override ordering for a security property. Verified against pi's shipped `sdk.js` / `agent-session.js`, not just the `.d.ts`.
- **Why does a TUI session now expose all 7 builtins (was 4)?** `grep/find/ls` are TypeClaw-owned tools we deliberately expose; a non-subagent session is not narrowed. This is an intentional widening of the *non-subagent* surface, not a regression — the subagent narrowing (the actual security boundary) is unchanged.

`@sinclair/typebox` imports become `typebox` (5 files); both `DefaultResourceLoader` constructors get the now-required `cwd`/`agentDir`; the pi-tui `FakeTerminal` test doubles gain the new `Terminal` members.

## What this does NOT do

- Does **not** adopt pi's new `Extension`/`ExtensionFactory` hook system (0.69) — TypeClaw still integrates by wrapping tools and monkey-patching `agent.afterToolCall`/`convertToLlm`. The session-replacement `ctx`-invalidation breaking change does not affect us (we call none of `newSession`/`fork`/`switchSession`/`withSession`).
- Does **not** wire up the new built-in providers that landed in this range (Fireworks 0.68.1, DeepSeek 0.70.1, Cloudflare 0.70.6) — separate follow-ups.
- Does **not** change runtime tool behavior for the common (hooked) session: the wrapped pipeline was already the de-facto path whenever hooks existed.

## Review notes

- **Load-bearing change:** `src/agent/plugin-tools.ts` (`wrapBuiltinToolDefinition`, `resolveBuiltinToolRefs`, the builtin map) and the `createAgentSession` call site in `src/agent/index.ts`.
- **Invariant to verify:** subagent tool narrowing is a privilege-escalation boundary. The `setupSession integration` suite pins it — a subagent declaring `[edit]` exposes **only** `edit` (not the other six wrapped builtins that ride in `customTools`), bash **fails closed** without bwrap, and builtins are always the wrapped (`sdk`) implementation, never pi's unwrapped `builtin`. If you touch the `tools:`/`noTools` wiring, those tests are the guardrail.
- Two `setupSession` tests changed their *expected* values to match the intended new behavior (edit is always `sdk`-sourced; TUI activates all 7). Both changes are annotated inline with the security rationale so they can't be silently reverted.
- Full `typecheck` / `lint` / `format` / test suite green (10308 pass). Design was reviewed with Oracle against pi's actual 0.73.1 source; the `noTools`+allowlist precedence was confirmed in `sdk.js`/`agent-session.js`, not inferred from docs.


shjeong92:
Could you add a TODO for the next pi migration?

Once we move beyond the `@mariozechner/pi-*` 0.73.x line to `@earendil-works/pi-* >= 0.80`, we should revisit and remove the TypeClaw-side adaptive-thinking compat shim added in #1122.

I checked `@mariozechner/pi-ai@0.73.1`, and it still does not force adaptive thinking for `claude-sonnet-5` / `claude-fable-5`, so the shim is still needed on top of this PR. The later `@earendil-works/pi-ai@0.80.3` line appears to handle this via `compat.forceAdaptiveThinking`, which should make the shim removable during that migration.

devxoul:
Done — tracked in #1126.

Added a grep-able `TODO(pi-migration)` marker to `src/agent/adaptive-thinking-compat.ts` with the concrete removal condition and the upstream mechanism that replaces it. I verified your claim against the published packages: `@earendil-works/pi-*` is the same maintainers' scope rename (`@mariozechner` ends at 0.73.1, `@earendil-works` starts at 0.74.0), and as of `@earendil-works/pi-ai@0.80.3` the `claude-sonnet-5`/`claude-fable-5` records carry `compat: { forceAdaptiveThinking: true }` (`AnthropicMessagesCompat.forceAdaptiveThinking`), which the Anthropic transport reads to emit `thinking: {type: "adaptive"}` itself — so the shim (and its `onPayload` wiring) become removable on that stack.

Also confirmed the shim is still needed on 0.73.1 as you noted, and refreshed the stale `0.67.3` pinned-version references in the shim header, its wiring in `src/agent/index.ts`, and the sonnet-5/fable-5 model records in `src/config/providers.ts`.

PR #1137, opened 2026-07-03
fix(channels): react only to the last message of a coalesced turn

devxoul:
## Background

When TypeClaw engages with a channel message on a typing-less adapter (GitHub, KakaoTalk), it drops an `:eyes:` reaction on the triggering message as an acknowledgment, then removes it once it replies. That ack fired **eagerly, per inbound**, inside `route()` — the moment each message passed the engagement gate, before any batching.

But messages that arrive within the debounce window (`INITIAL_DEBOUNCE_MS` / `HOT_DEBOUNCE_MS`) coalesce into a **single turn** at `drain()`. So a burst that all engages — sticky follow-ups, a user firing three quick messages — produced one reply but `:eyes:` on **every** message in the batch. The reaction is meant to mark "the message I'm answering", and a coalesced turn answers one logical thing, so the extra reactions are noise.

## Summary

Keep the eager fire (it preserves fast-ack feedback during the debounce window, which is exactly where it matters for slow async channels), but make it **rolling within `promptQueue`**: `enqueue()` now strips the eager reaction off earlier queued-but-not-yet-drained inbounds before pushing the new one. At any moment only the latest engaging message in the pending batch carries `:eyes:`, so the turn's single reply lines up with a single reaction.

**Why roll in `enqueue()` and not defer to `drain()`:** deferring until the batch is known would lose the sub-second ack during the debounce window. `promptQueue` *is* the "not-yet-drained batch" boundary, so rolling within it selects the future batch's last message without new state machinery. `drain()` already treats the last queued inbound as the turn trigger.

**Why `delete` and not overwrite** the `engageReaction` promise: a newer engaging inbound that has *no* `reactionRef` (unreactable) must still strip the previous message's `:eyes:` rather than leave it stranded on a non-triggering message.

Teardown also cleans up: `dropEngageReactions` only runs after a successful send, so a session destroyed mid-turn would otherwise strand the eager ack forever. `tearDownLive()` now drops both the queued and current-turn acks.

## What this does NOT do

- Does not change typing-capable adapters (Slack/Discord/Telegram) — they never got the eager `:eyes:` in the first place (the typing indicator is their ack).
- Does not touch the model-initiated `channel_react` tool or the disengage reaction, which were already scoped to the single triggering message.
- Does not change the engage/observe decision itself — only which engaged message keeps the acknowledgment.

## Review notes

- Load-bearing change: `clearQueuedEngageReactions()` wired into `enqueue()` (rolling supersede) and `tearDownLive()` (stranded-ack cleanup).
- Invariant to verify: after a coalesced batch, exactly one message retains `:eyes:` and it is the last engaging inbound; the superseded acks are removed *before* the turn produces a reply.
- The rewritten test `rolls the eager eyes onto only the last inbound...` is a real guard — it times out (fails) if the `enqueue()` supersede is removed. Added coverage for the `reactionRef`-less-supersede edge case and the teardown cleanup path.

### DIFF
--- a/src/channels/router.test.ts
+++ b/src/channels/router.test.ts
@@ -2424,13 +2424,14 @@ describe('ChannelRouter drop-eyes-after-reply', () => {
     expect(removed[0]).toMatchObject({ adapter: 'discord-bot', chat: 'c1', reactionRef: INSTANCE_REF })
   })
 
-  test('removes every engage reaction when multiple inbounds coalesce into one turn', async () => {
-    // given two inbounds debounced into a single turn, each with its own :eyes:
+  test('rolls the eager eyes onto only the last inbound when several coalesce into one turn', async () => {
+    // given three inbounds debounced into a single turn, each eagerly acked
     const dir = await tempDir()
     const { router, sessions } = makeRouter(dir)
     const instanceFor: Record<string, ReactionRef> = {
       'msg-a': { adapter: 'discord-bot', value: 'instance-a' },
       'msg-b': { adapter: 'discord-bot', value: 'instance-b' },
+      'msg-c': { adapter: 'discord-bot', value: 'instance-c' },
     }
     const removed: RemoveReactionRequest[] = []
     router.registerReaction('discord-bot', async (req) => ({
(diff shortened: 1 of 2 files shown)


devxoul:
Closing: this fix addressed the wrong problem. It scoped the eager engage-ack (:eyes:) to the last message of a coalesced batch, but that eager reaction is skipped entirely on typing-capable adapters (Discord/Slack/Telegram). The actual reported issue is on Discord, where the :eyes: comes from the model calling channel_react on messages it merely observes (e.g. the cron 'look around channels' job) but never replies to — and those model-added reactions are never cleaned up. Replacing with a proper fix: defer channel_react reactions and only apply them when the agent actually replies to the message in the same turn. New PR incoming.

PR #1152, opened 2026-07-06
fix(channels): use Discord native reply for anchored replies (bot + user adapters)

devxoul:
## Background

Reported in Discord (https://discord.com/channels/687271752224735233/1483291874964803614/1523602479244378133): when the agent replied to a message that had other messages in between (not the most recent one), Discord rendered a standalone `> @user:` inline blockquote instead of its native reply feature (the reply-arrow, backed by `message_reference`). The user witnessed this on the user-token `discord` adapter, which had no native reply support at all. Text-only anchored replies already used native reply; the remaining gap was attachment-only / textless sends.

This PR started scoped to the bot adapter only. It grew to cover **both** Discord adapters once the same defect was confirmed on the user-token side — that half was gated on an upstream `agent-messenger` release, which has since published, so the branch is now feature-complete and green.

Root cause, two adapters, two mechanisms:

1. **Router** — `resolveReplyRenderMode` gated Discord's native mode on text presence (`NATIVE_REPLY_TEXT_ADAPTERS`), and the user-token `discord` adapter wasn't in either native set. Anchored replies degraded to a blockquote whenever there was no text.
2. **discord-bot** — agent-messenger's SDK `uploadFile` posts a multipart-only create-message with no room for `message_reference`, so a textless bot reply had no wire path to a reply-arrow.
3. **discord (user-token)** — the adapter never threaded any reply reference through `sendMessage`/`uploadFile` at all, until `agent-messenger#290` added a `{ reply_to }` option (released in 2.29.0).

## Summary

- `src/channels/adapters/discord-bot.ts` — new `uploadFirstFileWithReply` does a raw multipart POST with `payload_json.message_reference`, since the SDK's `uploadFile` can't carry a reference. Only the first file in a textless, multi-attachment reply carries it; a text+attachment send keeps the reference on the text `sendMessage` instead.
- `src/channels/adapters/discord.ts` — threads `msg.replyTo` through `sendMessage`/`uploadFile` using the new `{ reply_to }` option from `agent-messenger` 2.29.0. Same one-reference invariant: text carries it when present, otherwise the first file.
- `src/channels/router.ts` — both `discord` and `discord-bot` move into `NATIVE_REPLY_EVERY_SHAPE_ADAPTERS`, so an attachment-only anchored reply on either adapter sets `replyTo` instead of prepending a blockquote.
- `package.json` / `bun.lock` — `agent-messenger` bumped to `^2.29.0`, required for the user-token half. That release is now published, which unblocked CI (see the earlier comment on this PR describing the interim release-gated red state).

## What this does NOT do

- Slack, Telegram, KakaoTalk, and GitHub are unchanged — none has a per-message reply primitive for file sends, so attachment-only replies on those adapters still degrade to the blockquote fallback (a guard test asserts Slack keeps doing this).
- Telegram and KakaoTalk stay text-gated in `NATIVE_REPLY_TEXT_ADAPTERS` — correct given their SDKs, not an oversight.
- No change to the already-native text-only reply path or to how the anchor candidate is selected/consumed.

## Review notes

- Load-bearing on both adapters: the reply reference must land on exactly one outbound message — the first text chunk when there's text, otherwise the first file. Verify the `replyOnFirstFile` / `index === 0` guards in both `discord.ts` and `discord-bot.ts`.
- `discord.ts`'s `replyOption(...)` helper and `discord-bot.ts`'s `uploadFirstFileWithReply` are two independent mechanisms (the agent-messenger-native option vs. a raw multipart POST) because the underlying SDKs differ for the two adapters — not meant to be unified.
- Tests: `discord.test.ts` (3 user-token reply-forwarding cases: text uses `reply_to`, attachment-only uses `reply_to` on the first file, text+attachment keeps the reference on text), `discord-bot.test.ts` (3 bot-adapter raw-multipart reply cases), `router.quote-anchor.test.ts` (native for both `discord` and `discord-bot`, text and attachment-only), `router.test.ts` (rewrote the old bug-encoding test to assert native `replyTo`; added the Slack blockquote guard). Full suite: 10501 pass / 0 fail; typecheck, lint, and format all clean.
- Dependency: requires `agent-messenger >= 2.29.0` (`agent-messenger#290`), now published and pinned in `ca5a6939` — the "CI is intentionally red — release-gated" condition from an earlier comment on this PR is resolved.


devxoul:
## ⚠️ CI is intentionally red — release-gated, not a defect

This branch now covers **both** Discord adapters:
- `discord-bot` (bot token) — commits `d2ae4c58`, `1b002e3d` — self-contained, was green.
- `discord` (user token) — commits `3d9f5867`, `431e8620` — **depends on [agent-messenger#290](https://github.com/agent-messenger/agent-messenger/pull/290)** (native `{ reply_to }` on `DiscordClient.sendMessage`/`uploadFile`), which is **merged to `main` but not yet released** (npm latest is `2.28.0`).

**Why typecheck fails:** exactly 2 errors in `src/channels/adapters/discord.ts` — the third `reply_to` argument doesn't exist on the pinned `2.28.0` signature. They disappear the moment `2.29.0` is installed. Lint and format are clean; all SDK-independent tests (incl. `router.quote-anchor.test.ts`) pass.

**Unblock steps (after agent-messenger `2.29.0` publishes):**
1. Bump `agent-messenger` → `^2.29.0`, `bun install`
2. `bun run typecheck` / `lint` / `format:check` / `test` (the 3 new `discord.test.ts` reply tests run then)
3. CI goes green → mergeable

Holding for the upstream release.

devxoul:
## Update — CI is now green, release landed

`agent-messenger@2.29.0` has published (agent-messenger#290 is out), so the dep bump in this branch (`^2.28.0` → `^2.29.0`) is no longer speculative. `bun install` picks up the real release, `typecheck`/`lint`/`format:check`/`test` all pass against it (10501 tests), and the scope has grown to cover the user-token `discord` adapter alongside `discord-bot`, per the updated PR description.

The "CI is intentionally red — release-gated" note above is resolved; nothing further is blocking merge.

PR #1199, opened 2026-07-14
feat(kakaotalk): add typing indicator via sub-device sendTyping

Gijuno:
## Summary

Registers a **typing capability** on the KakaoTalk adapter so the recipient sees the animated "…" composing dots while the agent generates a reply. Emits the LOCO `ACTION` packet (`type=1`) directly from the sub-device session via `KakaoTalkClient.sendTyping` ([agent-messenger#302](https://github.com/agent-messenger/agent-messenger/pull/302), shipped in `agent-messenger@2.31.0`), so **no external infra is needed** — the capability is always on, mirroring `webex`/`slack-bot`.

## Motivation

The one thing that very obviously read as "this is a bot" from the recipient's side was the missing typing indicator — the agent would go silent for 10-20s while generating, then dump the whole message at once. Every other realtime adapter (`webex`, `discord-bot`, `telegram-bot`, `slack-bot`) already surfaces typing; KakaoTalk was the gap. Now that the SDK exposes `sendTyping`, the adapter can register the capability like the rest.

## KakaoTalk's typing model

KakaoTalk's composing indicator is **fire-and-forget with no explicit clear**:

- One `sendTyping` call keeps the indicator up for **~5s**, then it auto-expires.
- Sending again **resets** the 5s window from that moment.
- A delivered message clears it **client-side** — there is no "stop typing" API.

This is the same shape as `telegram-bot`'s `sendChatAction` (also ~5s auto-expiry, no clear), and unlike `webex`'s explicit on/off `setTyping`.

### `createKakaoTypingCallback` — `src/channels/adapters/kakaotalk-typing.ts` (new)

Follows the shared adapter typing shape (per-chat FIFO queue, swallow-and-warn on failure, `adapter`/`phase` guards), plus one KakaoTalk-specific piece:

- **`phase: 'tick'`** → `sendTyping(chatId)` immediately, and arm a **self-refresh timer** (`KAKAO_TYPING_REFRESH_MS = 4000`).
- **`phase: 'stop'`** → clear the refresh timer only (no packet — KakaoTalk has no stop API; a delivered message clears the indicator).

**Why the self-refresh timer (the one deviation from `telegram-bot`):** the router heartbeat fires every `TYPING_HEARTBEAT_MS = 8000`, but KakaoTalk expires the indicator after ~5s — a 3s gap that would visibly blink the indicator off-and-on every heartbeat. `discord-bot` doesn't hit this (its indicator lasts ~10s, longer than the 8s heartbeat) and `telegram-bot` has the identical 5s/8s mismatch but deliberately accepts the gap (`// a missed beat just gaps the indicator`). KakaoTalk refreshes every 4s (~20% below the 5s expiry, using the "resend resets the window" property) to keep the indicator continuously visible. The timer is fully injectable (`setInterval`/`clearInterval` seams) for deterministic tests.

The per-chat FIFO mirrors `webex`/`slack-bot`: chaining every request through the per-chat tail keeps on-the-wire order matching enqueue order even under network jitter, so a slow start can't resolve after a follow-up and strand the indicator.

### `src/channels/adapters/kakaotalk.ts`

- Add `sendTyping(chatId, opts?: { linkId? }): Promise<KakaoTypingResult>` to the structural `KakaoTalkClient` interface.
- Register/unregister `typing.callback` + `setTypingCapability('kakaotalk', …)` in `start()`/`stop()` (inverse order, mirroring the other registrations), and call `typing.shutdown()` on stop to clear outstanding refresh timers.

### `agent-messenger` → `^2.31.0`

Bumped for the `sendTyping` surface (added in 2.31.0).

### Docs — `docs/content/docs/reference/kakaotalk.mdx`

Capability table `Typing indicator: no → yes`, plus a Quirks entry documenting the LOCO `ACTION` packet, the 5s auto-expiry / 4s self-refresh, and client-side clear on message delivery.

### Tests — `kakaotalk-typing.test.ts` (10, new) + `kakaotalk.test.ts` (FakeClient `sendTyping`)

- `tick` sends an ACTION packet; `stop` sends nothing.
- Ignores other adapters' targets.
- Swallows transport failures and non-zero `status_code` rejections as warnings (never throws).
- `formatChannelTag` used for the warning label.
- Per-chat FIFO ordering; no serialization across distinct chats.
- Self-refresh timer re-fires ACTION between heartbeats; `stop` clears it and a fired-but-cleared timer does not re-enqueue.
- `shutdown()` clears every outstanding refresh timer.

## Verification

`bun run check` (typecheck + lint + format) clean; the typing suite (10) and full kakaotalk adapter suite (39) pass.


devxoul:
@typeey which pr is better #1198 and this (#1199)

devxoul:
> **#1199** is the stronger direction: it closes the 5s-vs-8s heartbeat gap with a 4s refresh and gives the timer lifecycle explicit tests. #1198 still has the blink issue I blocked on, plus missing OpenChat suppression coverage. I’ve got a full review of #1199 running now, but architecturally I’d choose #1199. (๑˃̵ᴗ˂̵)و

alright then please close #1198 with the context

Gijuno:
Thanks for the review — all three addressed in 42ea689.

**1. Queued refreshes surviving `stop` (`kakaotalk-typing.ts`).** Added a per-chat generation token: each pulse captures the chat's generation at enqueue time and re-checks it immediately before hitting the wire. `clearRefresh` (fired by `stop`, `shutdown`, or teardown) deletes the generation, so a pulse already queued behind a slow send is dropped instead of re-raising the indicator after a delivered message cleared it. Regression test added: block the first send, fire the refresh (queuing a second pulse), land `stop`, then release the first send — only the already-in-flight packet ships, the queued one is dropped.

**2. OpenChat `linkId` (`kakaotalk-typing.ts:61`).** Went with the skip-and-document path you suggested, mirroring `markReadIfSupported`: typing is skipped for `@kakao-open` with a once-per-chat log (`reason=open_chat_link_id_unsupported`), since the resolver doesn't surface the link ID today. Wiring `linkId` through the resolver is a follow-up shared with mark-read. Tests assert `@kakao-open` sends nothing (logged once) while `@kakao-group`/`@kakao-dm` still emit.

**3. Adapter lifecycle test (`kakaotalk.ts:681`).** Added a start/stop test on the real router + fake client that captures the callback the adapter registers, drives a `tick` through it to `client.sendTyping`, then asserts `stop()` unregisters the callback, disables the capability, and that a post-stop `fireTypingHeartbeat` no longer reaches the client.

`bun run check` clean; typing suite (13) + kakaotalk adapter suite (40) pass.

Gijuno:
Pushed c09759f — plus an architectural change worth calling out.

## Root-cause fix: the adapter no longer owns a timer

Both prior rounds were really about one thing: KakaoTalk was the only adapter running its own `setInterval` self-refresh loop (to bridge the 3s gap between its ~5s indicator expiry and the router's 8s heartbeat). That independent loop is what created the stop-race and the shutdown-lifecycle surface you flagged.

Rather than keep hardening the adapter-owned timer, this commit removes it. The router already owns a per-session typing heartbeat; I made its interval per-adapter via a new `setTypingHeartbeatInterval(adapter, ms)`. KakaoTalk registers a 5s heartbeat (faster than the 5s expiry) and the router paces the refresh — so the callback becomes **stateless**, exactly like `telegram-bot`/`webex`:

- `tick` → one `sendTyping` packet; `stop` → no-op (no stop API, delivered message clears it client-side).
- **No independent timer → no stop-race → the generation gate is gone.**
- **No timer to leak → `typing.shutdown()` and the timer test seams are gone.**
- KakaoTalk is no longer the only adapter with its own typing loop.

Net −121 LOC.

## Backwards compatibility

`setTypingHeartbeatInterval` is opt-in and defaults to `TYPING_HEARTBEAT_MS` (8s). Every other adapter is unaffected — asserted directly in `router.test.ts` (default-vs-override + per-adapter isolation).

## The two blockers from re-review

Both were addressed before this refactor and remain covered:

- **OpenChat docs** — `docs/reference/kakaotalk.mdx` now marks the capability `DM/group` (not universal) and documents that `@kakao-open` is skipped until `linkId` is exposed.
- **Lifecycle test** — the old test's weakness (it cleared refresh state before `adapter.stop()`) is moot now that there's no refresh state. The rewritten test asserts the adapter registers the 5s heartbeat and that a post-stop heartbeat dispatched through the **real router path** no longer reaches the client.

`bun run check` clean; kakaotalk typing (11) + adapter (40) + router (full) suites pass, and the 8 tool-test router mocks were updated for the new interface method.

Gijuno:
Pushed 4f08334 — all three re-review points addressed.

**1. Queued pulse could still send after `stop`.** You were right: removing the adapter timer also removed the generation gate, but the race lived in the per-chat FIFO queue, not the timer. Restored a **timer-independent** per-chat generation token — captured when a tick is accepted, re-checked immediately before `sendTyping`, and deleted on `'stop'` and on `reset()`. The callback handle now exposes `reset()`, which `adapter.stop()` calls to invalidate every queued pulse. Regression tests cover both the `'stop'`-invalidates-queued and `reset()`-invalidates-queued paths (blocked first tick → queued second tick → stop/reset → release → only the in-flight packet ships).

**2. 5s heartbeat had no margin.** Agreed — dropped to **4s** (the interval upstream live-verified), ~20% below the ~5s expiry so a replacement packet lands before expiry under scheduler/network jitter. Updated the assertion and docs.

**3. Lifecycle test post-stop assertion was vacuous.** Correct — `fireTypingHeartbeat` no-ops without a live session. Replaced it with a **blocked-tick → queued-second-tick → `adapter.stop()` → release** scenario driven through the exact callback the router holds, asserting only the in-flight packet ships. I verified it's **non-vacuous by mutation**: removing `typing.reset()` from `stop()` makes it fail (expected 2, received 3), then passes again once restored.

`bun run check` clean; typing (14) + adapter (41) + router (full) suites pass — 662 across the three files.

Gijuno:
Pushed d15d9bd — the provisional-classification gap is fixed.

**Good catch.** The resolver caches a room absent from `getChats` as a provisional `@kakao-group` placeholder, so a not-yet-confirmed OpenChat slipped past the `@kakao-open`-only skip and would get `sendTyping(chatId)` without its `linkId` on every heartbeat.

**Fix — classify live, track authority:**

- **Resolver**: added a `provisional` flag to `Entry`/`KakaoChatLookupValue`. Authoritative `getChats` entries are `false`; `ingestProvisional` entries are `true`.
- **Typing**: the callback no longer trusts the (stale) session-key workspace — it classifies each chat **live at tick time** via the resolver. `kakaoTypingClassFromLookup` returns `send | skip-open | skip-unresolved`; it sends only for an authoritative DM/group and **silently skips provisional/unknown** rooms until a refresh confirms the kind. So a room that resolves to OpenChat later is never sent a doomed packet in the meantime.
- **Adapter**: `classifyChat` wired to `channelResolver.lookupChat`.

**Regression test (as requested):** `getChats` omits an OpenChat room, an inbound push registers it provisionally, and a typing tick emits **no ACTION packet**. Verified non-vacuous by mutation — forcing `classifyChat` to `'send'` makes it fail (expected 0, received 5). Added unit coverage for the classifier (`null`/provisional/open/authoritative) and the resolver `provisional` flag, plus a "suppressed-while-provisional, sends once it resolves to group" test proving the live re-classification.

`bun run check` clean; 681 tests pass across typing + adapter + resolver + router suites.

devxoul:
Thanks for your contribution @Gijuno :tada:

PR #1204, opened 2026-07-14
[Superseded] feat(memory): wire provenance recall permissions

devxoul:
## Background

Extracted unchanged from #1200 to make the work reviewable. This is slice 2/6.

## Summary

Adds the global recall capability, live tool caller context, and origin-ancestry access resolution.

## Review notes

The base is the provenance-index slice, so this diff contains only access-control wiring.

### DIFF
--- a/src/bundled-plugins/memory/provenance-index.test.ts
+++ b/src/bundled-plugins/memory/provenance-index.test.ts
@@ -1,0 +1,768 @@
+import { afterEach, describe, expect, test } from 'bun:test'
+import { lstat, mkdir, mkdtemp, readFile, rm, symlink, writeFile } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+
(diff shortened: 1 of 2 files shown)


devxoul:
This PR is superseded. The proposed `memory.recall.global` permission and origin-ancestry-derived recall scoping were removed from the memory-provenance stack: provenance (workspace/chat/thread, captured names) is searchable context for lexical matching and attribution, not an authorization boundary, and gating recall by channel/role would have made memory silently narrower than intended by default.

The replacement approach lives in #1203: `memory_search`/`hybridSearch` accept **optional, explicit** `workspace`/`chat`/`thread` filters that a caller can supply when it wants a focused subset, while recall remains globally visible by default across all roles and origins, with no permission check and no ancestry walk.

This branch's head is now identical to #1206's head (no diff against the stack). GitHub won't let a merged PR be closed via the API/CLI (`can't be closed because it was already merged`), so it's left as-is with this note rather than closed; the branch is not deleted.

PR #1209, opened 2026-07-14
feat(github): submit formal scoped reviews

devxoul:
## Background

Extracted from #1201 to make the work reviewable. Slice 8/11.

## Summary

Scopes GitHub App tokens per repository and registers verified formal review submitters.

## Review notes

This slice contains the router and adapter review-submission boundary only.


devxoul:
Addressed all three current blockers at `c21c3295`: verification now requires the exact event-matching GitHub state; multiline range fields are accepted/serialized only as a pair; and anchor collection retries against a rechecked stable head with a bounded failure path. Added mismatch, partial-range, head-change retry, and unstable-head regressions.

PR #1211, opened 2026-07-14
feat(agent): pin tool file operands

devxoul:
## Background

Extracted from #1201 to make the work reviewable. Slice 4/11.

## Summary

Pins built-in, plugin, and MCP file operands across execution and anchors output writes.

## Review notes

This is test-heavy but limited to one security boundary and its direct integration tests.

### DIFF
--- a/src/agent/index.test.ts
+++ b/src/agent/index.test.ts
@@ -4,7 +4,8 @@ import { mkdir, mkdtemp, readFile, rm, writeFile } from 'node:fs/promises'
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
 
-import { SessionManager } from '@mariozechner/pi-coding-agent'
+import { defineTool as definePiTool, SessionManager } from '@mariozechner/pi-coding-agent'
+import { Type } from 'typebox'
 
 import { PLANNER_SYSTEM_PROMPT } from '@/bundled-plugins/planner/planner'
 import { SCOUT_SYSTEM_PROMPT } from '@/bundled-plugins/scout/scout'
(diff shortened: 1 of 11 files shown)


devxoul:
Addressed the fresh cleanup lifecycle blocker at `e28c9738`: plugin and custom-system wrappers now select execution/cleanup errors after cleanup, invoke `tool.after` exactly once, and propagate cleanup/output-verification failures. Added plugin output-verification coverage and stabilized the queued hardlink regression.

devxoul:
Fixed the Linux CI output-pinning regression at `6d43679b`. Plugin-declared `fileOperands.output` paths now enter the same inode-anchored destination verification as builtin write/edit tools, including arbitrary declared operand names and bounded collections. The deterministic Linux regression replaces the declared destination with a symlink during execution, verifies rejection, and confirms `tool.after` receives the error exactly once. Local focused: 120 pass / 20 platform skips; full: 11,228 pass / 26 skips. Threads remain unresolved for reviewer verification.

devxoul:
Follow-up at `2d1b5ae0`: preserved the established plugin `errorResult` shape in the lifecycle assertion (`{ error: true, message }`) and made the researcher writer validate the lexical parent recovered from the runtime-owned `/proc/self/fd/<dirfd>` anchor while still writing through that anchored path. Direct non-anchored callers retain the strict workspace/public lexical-parent rule. Focused: 138 pass / 19 platform skips; full: 11,229 pass / 25 skips. Threads remain unresolved.

devxoul:
CI correction folded into the owning file-pinning commit at 37cecd7d. The hardlink race test now asserts the operation never dispatches and accepts only the two valid secure denials: canonical private-surface rejection or hardlink/pinning rejection. The focused test passed 20 consecutive runs on both the #1211 slice and the final descendant stack; full checks pass. Re-review requested; no threads were resolved.

PR #1216, opened 2026-07-14
feat(network): add SSRF-safe bounded transport

devxoul:
## Background

Extracted from #1201 to make the work reviewable. Slice 5/11.

## Summary

Adds DNS-pinned public HTTP transport, rewires web fetch, and bounds multimodal response bodies.

## Review notes

The bounded response helper was reordered before its consumers so this slice builds independently.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -20,6 +20,7 @@
         "jsdom": "^29.0.2",
         "proper-lockfile": "^4.1.2",
         "qrcode": "^1.5.4",
+        "tough-cookie": "^6.0.2",
         "turndown": "^7.2.4",
         "zod": "^4.3.6",
       },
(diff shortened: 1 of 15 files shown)


devxoul:
Addressed all four blockers at `3a424a8a`: warmup cookies now use a standards-compliant URL-scoped `tough-cookie` jar on every redirect hop; IPv6 addresses are semantically normalized before mapped-IPv4 classification; multimodal workers share cancellation and await `allSettled`; and local files are opened/fstat’d first, read with a size+1 bound, then re-fstat’d. Added cross-origin Domain/Path/Secure/expiry, expanded/compressed/dotted mapped-loopback, sibling cancellation, and bounded-read regressions. Local lint, format check, typecheck, and full test suite pass (`11146 pass`, `20 skip`, `0 fail`).

PR #1225, opened 2026-07-15
Fix/container host ownership

gtg7784:
## Background

TypeClaw mounts the host-stage agent folder into the container at `/agent`. The container runtime previously executed as root, so container-stage writes could create root-owned files on the host.

Once files such as `typeclaw.json` became root-owned, host-stage commands failed with errors such as:

```
EACCES: permission denied, open './typeclaw.json'
```

This could affect any bind-mounted agent state written by the runtime, not only model configuration.

## Summary

Preserve host ownership for container-stage writes by passing the POSIX host UID/GID into the container and dropping privileges before starting the TypeClaw runtime.

The entrypoint still starts as root because network-policy bootstrap requires it. After bootstrap completes, it re-executes under the host identity using `setpriv`.

### Why this approach instead of `docker run --user`:
- iptables and other bootstrap operations still require root.
- Running the entire container as the host user would break those operations.
- Dropping privileges immediately before runtime startup preserves the existing bootstrap boundary while preventing root-owned bind-mount writes.

The runtime HOME is moved to `/agent/.typeclaw/home/runtime`. Ownership preparation is restricted to the TypeClaw-managed home subtree rather than recursively changing `/agent`.

For agent folders affected by older containers:
- `typeclaw start` stops early when an existing `typeclaw.json` is not writable.
- `typeclaw doctor` detects foreign-owned managed files and provides an explicit recovery command.
- Model configuration permission errors include ownership diagnostics and recovery guidance.

The model CLI also accepts a model reference as shorthand for updating the default profile:

```
typeclaw model set openai-codex/gpt-5.5
```

The existing `model set <profile> <ref>` syntax remains supported.

## Review-round fixes

The following issues were raised in review and addressed as separate commits (one concern each):

- **Xvfb double capability-drop (`container:` commit).** `start_xvfb` wrapped Xvfb in a nested `setpriv --bounding-set` to strip `CAP_NET_ADMIN`. But on the default POSIX path the shim first re-exec's into the non-root runtime phase via an outer `setpriv` that empties the bounding set and drops `CAP_SETPCAP` — so the nested bounding-set drop then requires a capability it no longer has and exits 127, which the liveness probe misreports as "Xvfb exited immediately". Since `docker.file.xvfb` defaults to true, ordinary POSIX container starts hit this. `start_xvfb` now branches on `TYPECLAW_ENTRYPOINT_RUNTIME`: in the runtime phase it launches Xvfb directly (caps are already gone — nothing left to strip), and retains the nested drop only for the root fallback paths where the shim is still PID 1 with the full capability set.

- **Privilege handoff wiped required runtime env (`container:` commit).** Both privilege-drop re-execs passed `setpriv --reset-env`, which clears every non-terminal variable; the trailing `env` then restored only `HOME` and `TYPECLAW_ENTRYPOINT_RUNTIME`. That lost the `--env-file` values (provider credentials, etc.) plus `TYPECLAW_SANDBOX_SYMLINKS`, `TYPECLAW_TUI_TOKEN`, `TYPECLAW_CONTAINER_NAME`, `TYPECLAW_HOSTD_URL`, `TYPECLAW_MODEL_CACHE`, and `TZ` — all injected by `start.ts` and read after the re-exec. `--reset-env` is removed from both handoffs; the re-exec already overrides exactly the two variables the runtime phase needs (`env HOME=… TYPECLAW_ENTRYPOINT_RUNTIME=1`) and forwards everything else unchanged.

- **Managed HOME hid image-baked CLI settings (`container:` commit).** The image writes Claude Code onboarding state and hooks (`~/.claude.json`, `~/.claude/settings.json`) and Codex hooks (`~/.codex/hooks.json`) against `/root` at build time. Under the new runtime `HOME`, those files became invisible, silently disabling Claude onboarding and both CLIs' completion hooks for `claudeCode`/`codexCli` users (`link_persistent_home_files` links only credentials). A new `seed_runtime_home` runs in the root setup phase (before the chown, so copies inherit host ownership) and `cp -n`s the baked files into the runtime HOME — filling only what's missing so an operator's persisted config is never clobbered.

- **Model shorthand suppressed the interactive thinking prompt (`model:` commit).** The default-profile shorthand changed the interactive guard to `args.profile === undefined`, which also suppressed the thinking-level prompt for `model set fast` — a named profile with no ref that still selects its ref interactively and, per the adjacent contract, should keep offering the prompt. The guard now keys on whether a ref actually came from the command line (explicit `<ref>` or shorthand), not on `args.profile`, so named-profile interactive flows still prompt while the shorthand and explicit-ref paths do not.

## What this does NOT do

- It does not recursively change ownership of the entire agent folder.
- It does not use permissive modes such as `chmod 777`.
- It does not invoke `sudo` or automatically repair existing files.
- It does not remove the root bootstrap phase.
- It does not apply POSIX ownership handling on native Windows.
- It does not change the existing explicit-profile model syntax.

## Review notes

The load-bearing entrypoint sequence is:
1. Start as root.
2. Prepare the TypeClaw-managed runtime HOME (including seeding image-baked Claude/Codex settings).
3. Install network-isolation rules when enabled.
4. Drop supplementary groups, capabilities, UID, and GID (without `--reset-env`, so injected runtime env survives).
5. Re-enter the runtime phase with the managed HOME.
6. Initialize persistent credential links, configured symlinks, and Xvfb (launched directly in the runtime phase).
7. Start TypeClaw.

Please pay particular attention to these invariants:
- Both network-policy paths reach the non-root runtime phase.
- Firewall rules are installed before privilege drop when internal-network blocking is enabled.
- Xvfb and the agent runtime do not retain `CAP_NET_ADMIN` — and Xvfb does not attempt a second capability drop once already in the non-root runtime phase.
- The privilege handoff preserves `--env-file` values and TypeClaw-injected runtime variables (no `--reset-env`).
- Image-baked Claude/Codex settings are seeded into the effective non-root runtime HOME without clobbering persisted files.
- Ownership changes remain restricted to `/agent/.typeclaw/home`.
- Startup remains functional when POSIX host identity variables are unavailable.
- The model shorthand does not suppress the interactive thinking prompt for named-profile flows.
- Native Windows behavior remains unchanged.


devxoul:
@gtg7784 could you resolve the conflicts?

gtg7784:
> @gtg7784 could you resolve the conflicts?

addressed conflicts

PR #1228, opened 2026-07-16
fix(agent): stop file-operand scanner rejecting prose in prompt/query/reason fields

devxoul:
## Background

A Discord agent running from `main` posted `file:///tmp/typeclaw-tool-input-XpwgiL/0` as a chat message (followed by an empty message) instead of a status update. That specific symptom was the `channel_reply` text leak fixed in #1227 — but it's one face of a broader regression in the generic file-operand scanner (`e478c37c`), and #1227 only closed the channel surface.

`enforceAndPinToolFiles` runs the undeclared-operand scan on every system tool except `mcp_call`. In `collectGenericFileTargets`, `isAmbiguousUndeclaredLocalOperand` treats any string that is path-shaped (contains a slash, starts with `./`, matches `name.ext`, or resolves to a real file) as a file operand — pinning it to a `/tmp/typeclaw-tool-input` copy, or throwing `ambiguous local file operand`. That is correct for `read`/`look_at`/attachments, but wrong for **prose** args.

Reproduced on `main` (post-#1227), still broken:

- `spawn_subagent({ prompt: "Look at src/router.ts and summarize." })` → **throws** `ambiguous local file operand at key "prompt"` — a subagent prompt that names a file is rejected, breaking the core delegation flow.
- `skip_response({ reason: "…notes.md" })`, `web_search({ query: "configure vite.config.ts" })`, `todo_write({ todos: [{ content: "edit src/router.ts" }] }])` all fail the same way whenever the value carries a path-shaped token.

## Summary

The root problem is that `tool-file-safety.ts` re-derives file-ness from a value's *spelling*, while the `private-surface-read` guard already classifies the same keys (`text`/`body`/`prompt`/`query`/`reason`/…) as prose. The two guards disagreed. I export that knowledge as `classifyFreeTextField` and consult it in `collectGenericFileTargets` before the value heuristic.

Precedence (per an Oracle design pass on the security intent):

1. declared `output`/`destructive` → ignore
2. declared `input` → pin (tool-author metadata always wins)
3. **`opaque` prose field** (`text`/`body`/`content`/`prompt`/`query`/`reason`/…) → skip, **even a whole-string `file:` URI** (it's a message payload, not a file to open)
4. **`file-uri` field** (`url`) → skip the path-shape heuristic but **still pin an explicit `file:` URI**
5. explicit `file:` URI on an unknown key → pin
6. unknown key, path-shaped value → still throw (fail-closed preserved)

**Why `url` is split out (not just lumped with prose):** an existing security test pins a `file://` URL passed to a reader tool so a concurrent symlink swap can't change the bytes read (TOCTOU defense). Blanket-skipping `file:` URIs on every free-text key would silently disable that. `url` is the one free-text key whose `file:` value the tool actually dereferences, so it keeps pinning; pure-prose keys never do.

## What this does NOT do

- Does not re-touch `channel_send`/`channel_reply`/`look_at` — #1227 already scopes those to `attachments[].path` and is the strongest contract for them. This generalizes the same idea to the rest of the prose surface.
- Does not weaken fail-closed behavior for undeclared operands on unknown keys (`value: "data.bin"` still throws).
- Does not add a new `fileOperands` opt-out API; it reuses the existing `private-surface-read` key knowledge so the two guards stop disagreeing.

## Review notes

Load-bearing change: the new `classifyFreeTextField(tool, key)` in `private-surface-read.ts` (returns `'opaque' | 'file-uri' | undefined`) and its two consult sites in `collectGenericFileTargets`. Verify the precedence: an `opaque` field short-circuits before the `file:` check; a `file-uri` field (`url`) does not.

Invariant to confirm still holds: the pre-existing "ambiguous system tools execute explicit file URLs against an immutable pinned copy" test still passes — a `file://` under `url` is pinned, so the symlink-swap defense is intact. New regression tests assert prose fields (`spawn_subagent.prompt`, `skip_response.reason`, `web_search.query`, `todo_write.todos[].content`) survive unchanged, `url`+`file:` still pins, and an unknown-key path-shaped operand still fails closed.

### DIFF
--- a/src/agent/plugin-tools.test.ts
+++ b/src/agent/plugin-tools.test.ts
@@ -1145,6 +1145,48 @@ describe('wrapSystemTool', () => {
     }
   })
 
+  test('free-text prose fields with path-shaped values are never pinned or rejected', async () => {
+    const agentDir = await mkdtemp(path.join(tmpdir(), 'typeclaw-freetext-'))
+    await mkdir(path.join(agentDir, 'src'), { recursive: true })
+    await writeFile(path.join(agentDir, 'src', 'router.ts'), 'x')
+    await writeFile(path.join(agentDir, 'status.txt'), 'done')
+    const bodyUrl = pathToFileURL(path.join(agentDir, 'status.txt')).href
+
+    const cases: Array<[string, Record<string, unknown>]> = [
+      ['spawn_subagent', { subagent_type: 'explore', prompt: 'Look at src/router.ts and summarize.' }],
+      ['spawn_subagent', { subagent_type: 'explore', prompt: bodyUrl }],
+      ['skip_response', { reason: 'nothing to do, see notes.md' }],
+      ['web_search', { query: 'how to configure vite.config.ts' }],
+      ['todo_write', { todos: [{ content: 'edit src/router.ts', status: 'pending', priority: 'high' }] }],
+    ]
+    for (const [tool, args] of cases) {
+      const before = JSON.stringify(args)
+      const pinned = await enforceAndPinToolFiles({ tool, args, agentDir, genericInputs: true })
+      await pinned.cleanup()
+      expect(JSON.stringify(args)).toBe(before)
+    }
+    await rm(agentDir, { recursive: true, force: true })
+  })
+
+  test('a url field still pins an explicit file: URI (symlink-swap defense preserved)', async () => {
+    const agentDir = await mkdtemp(path.join(tmpdir(), 'typeclaw-url-pin-'))
+    await writeFile(path.join(agentDir, 'src.txt'), 'real')
+    const args: Record<string, unknown> = { url: pathToFileURL(path.join(agentDir, 'src.txt')).href }
+    const pinned = await enforceAndPinToolFiles({ tool: 'custom_reader', args, agentDir, genericInputs: true })
+    await pinned.cleanup()
+    expect(String(args.url)).toContain('typeclaw-tool-input')
+    await rm(agentDir, { recursive: true, force: true })
+  })
+
+  test('undeclared path-shaped operand on an unknown key still fails closed', async () => {
+    const agentDir = await mkdtemp(path.join(tmpdir(), 'typeclaw-failclosed-'))
+    await writeFile(path.join(agentDir, 'data.bin'), 'd')
+    await expect(
+      enforceAndPinToolFiles({ tool: 'some_plugin_tool', args: { value: 'data.bin' }, agentDir, genericInputs: true }),
+    ).rejects.toThrow(/ambiguous local file operand/)
+    await rm(agentDir, { recursive: true, force: true })
+  })
+
   test('parallel read, look_at, and channel upload calls consume pinned bytes across symlink swaps', async () => {
     const agentDir = await mkdtemp(path.join(tmpdir(), 'typeclaw-pinned-read-'))
     const safe = path.join(agentDir, 'safe.txt')
--- a/src/agent/plugin-tools.test.ts
+++ b/src/agent/plugin-tools.test.ts
@@ -1145,25 +1145,57 @@ describe('wrapSystemTool', () => {
     }
   })
 
-  test('free-text prose fields with path-shaped values are never pinned or rejected', async () => {
-    const agentDir = await mkdtemp(path.join(tmpdir(), 'typeclaw-freetext-'))
+  test('first-party prose operands with path-shaped values are never pinned or rejected', async () => {
+    const agentDir = await mkdtemp(path.join(tmpdir(), 'typeclaw-prose-'))
     await mkdir(path.join(agentDir, 'src'), { recursive: true })
     await writeFile(path.join(agentDir, 'src', 'router.ts'), 'x')
-    await writeFile(path.join(agentDir, 'status.txt'), 'done')
-    const bodyUrl = pathToFileURL(path.join(agentDir, 'status.txt')).href
 
     const cases: Array<[string, Record<string, unknown>]> = [
       ['spawn_subagent', { subagent_type: 'explore', prompt: 'Look at src/router.ts and summarize.' }],
-      ['spawn_subagent', { subagent_type: 'explore', prompt: bodyUrl }],
+      ['spawn_subagent', { subagent_type: 'explore', description: 'check src/app.ts' }],
       ['skip_response', { reason: 'nothing to do, see notes.md' }],
       ['web_search', { query: 'how to configure vite.config.ts' }],
+      ['web_fetch', { url: 'https://example.com/a', query: '.data[0].path', selector: 'div.a/b', pattern: 'a/b\\d' }],
       ['todo_write', { todos: [{ content: 'edit src/router.ts', status: 'pending', priority: 'high' }] }],
+      ['channel_edit', { workspace: 'W', chat: 'C', message_id: '1', text: 'fixed in src/router.ts (7/16)' }],
     ]
     for (const [tool, args] of cases) {
       const before = JSON.stringify(args)
       const pinned = await enforceAndPinToolFiles({ tool, args, agentDir, genericInputs: true })
       await pinned.cleanup()
       expect(JSON.stringify(args)).toBe(before)
+    }
+    await rm(agentDir, { recursive: true, force: true })
+  })
+
+  test('prose exemption is tool-scoped: an undeclared reader with a common key still fails closed', async () => {
+    const agentDir = await mkdtemp(path.join(tmpdir(), 'typeclaw-toolscoped-'))
+    await writeFile(path.join(agentDir, 'data.json'), 'SENSITIVE')
+    // `content`, `body`, `prompt`, `name` are opaque prose ONLY for their declared
+    // first-party tools. An undeclared plugin/MCP reader must not inherit that.
+    for (const key of ['content', 'body', 'prompt', 'name']) {
+      await expect(
+        enforceAndPinToolFiles({
+          tool: 'plugin_reader',
+          args: { [key]: `${agentDir}/data.json` },
+          agentDir,
+          genericInputs: true,
+        }),
+      ).rejects.toThrow(/ambiguous local file operand/)
+    }
+    await rm(agentDir, { recursive: true, force: true })
+  })
+
+  test('a whitespace- or case-varied file: URI is pinned, never passed through', async () => {
+    const agentDir = await mkdtemp(path.join(tmpdir(), 'typeclaw-ws-uri-'))
+    await writeFile(path.join(agentDir, 'data.json'), 'SENSITIVE')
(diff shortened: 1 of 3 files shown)


devxoul:
## Review (standing in for `typeey`, which is blocked by this very bug)

`typeey` couldn't post a verdict — the runtime rejected its own reviewer-delegation `prompt` with `ambiguous local file operand at key "prompt"`, which is exactly the regression this PR targets. I reproduced that failure, then ran the fix through typecheck + the full `plugin-tools` / `private-surface-read` suites (all green) and an adversarial probe pass against the PR HEAD (`cfcbda3`).

**Verdict: request changes — 2 confirmed merge-blocking holes.** The fix is correct for the four motivating first-party tools, but reusing the *global* `NON_PATH_KEYS` set to gate the pinning scanner punches a hole in its fail-closed contract for undeclared plugin/MCP readers, and there's a whitespace-trim `file:` bypass on `url`. Both are reproduced below against this branch.

### What's right
- Root cause is correctly identified: `tool-file-safety.ts` re-derived file-ness from a value's *spelling* while `private-surface-read` already classified the same keys as prose; the two disagreed.
- Precedence for first-party tools is sound: declared `input` still wins (`declaredInput ? undefined : classifyFreeTextField(...)`), `opaque` short-circuits before the `file:`/heuristic checks, `url` is split into `file-uri` so a real `file:` URL still pins (TOCTOU preserved).
- The `tool` arg is threaded through all three `collectGenericFileTargets` sites and — verified across every call site — is **never model-controlled** (always a static literal / `tool.name` / `opts.toolName`), so `classifyFreeTextField(tool, key)` can't be gamed by tool-name spoofing.
- Whole-string `file:` URIs to *canonical* secrets in a prose field are still denied independently by `enforceCanonicalSecretDenial` (runs before `collectGenericFileTargets`), so that specific path is not a regression.

### Hole 1 (merge-blocking): global `opaque` keys break the generic fail-closed contract
An undeclared plugin/MCP reader whose arg key happens to be in `NON_PATH_KEYS` (`content`, `body`, `prompt`, `name`, …) now bypasses **both** the ambiguous-operand throw and pinning for a plain path. Reproduced on `cfcbda3`:

```ts
const args = { content: `${agentDir}/secrets.json` }
await enforceAndPinToolFiles({ tool: 'plugin_content_reader', args, agentDir, genericInputs: true })
// → does not throw, does not pin. args.content unchanged →
//   a downstream reader does Bun.file(args.content).text() === "TOP_SECRET"
// {"hole":1,"threw":false,"wouldLeak":true}
```

Control (same value, unknown key `blob`) still fails closed → confirms the regression is *caused* by `content` being globally opaque:

```ts
enforceAndPinToolFiles({ tool: 'plugin_content_reader', args: { blob: `${agentDir}/secrets.json` }, ... })
// → rejects: "ambiguous local file operand"  ✅
```

The whole point of the generic scan is to fail closed for a reader that forgot to declare `fileOperands.input`. Inheriting a prose exemption purely from a common key name (and with no operand metadata available at all inside `mcp_call`) reopens exactly that gap.

### Hole 2 (merge-blocking): leading-whitespace `file:` bypass on `url`
Both guards test `startsWith('file:')` without trimming; because `url` is `file-uri`, the path-shape fallback is suppressed, so a whitespace-prefixed `file:` URI is neither pinned nor denied. A downstream fetcher that trims-before-parse dereferences the secret. Reproduced on `cfcbda3`:

```ts
const args = { url: `  file://${agentDir}/secrets.json` }
await enforceAndPinToolFiles({ tool: 'web_fetch', args, agentDir, genericInputs: true })
// {"hole":2,"threw":false,"pinned":false}  — arg passes through untouched
```

### Suggested direction
1. Don't reuse the *global* `NON_PATH_KEYS` for pinning exemptions. Gate the pinning scanner on a **tool-scoped, full-operand-path allowlist** — `spawn_subagent.prompt`, `skip_response.reason`, `web_search.query`, `todo_write.todos.content` — so unknown plugin/MCP tools do **not** inherit an exemption just for using a common key. `classifyFreeTextField` can take the resolved operand-path (not just the leaf key) and return `undefined` for unlisted tools.
2. Keep `fileOperands.input` as the highest-priority signal (unchanged); `output`/`destructive` keep winning on overlap.
3. Normalize `file:` detection on a **trimmed** value consistently across both guards and `normalizeFileReference`, while preserving the original string for result restoration.
4. Add regressions for both hostile cases: undeclared `content: <agentDir>/secrets.json` must deny-or-pin, and `url: "  file:///…"` must deny-or-pin. Keep the four first-party prose tests proving payloads pass unchanged.

Effort is small (~1–4h). The current tests pass but don't cover either hostile case. Happy to re-review once the exemption is tool-scoped and the trim is added.

<sub>Reviewed by Sisyphus on behalf of the blocked `typeey` reviewer. Findings cross-checked with an Oracle security pass and empirical probes against PR HEAD `cfcbda3`.</sub>

devxoul:
Addressed both merge-blocking holes in `5e83114`. Thanks for the precise reproductions — both were real and are now closed with hostile regressions.

## Hole 1 — global prose-key exemption → tool-scoped allowlist
Root problem was exactly as described: reusing the *global* `NON_PATH_KEYS` handed any tool an exemption purely from a key name, so an undeclared plugin/MCP reader with `{ content: "<dir>/data.json" }` neither threw nor pinned.

Fix: dropped the shared `classifyFreeTextField`/`NON_PATH_KEYS` reuse for the pinning scanner. `private-surface-read`'s global prose set stays private to *its* read-guard (where a guest's `channel_reply({ text: "the memory leak" })` must not be blocked — a separate concern). The pinning scanner now has its own **exact `tool + operandPath` allowlist**:

```ts
const PROSE_OPERANDS = {
  spawn_subagent: {prompt, description}, skip_response: {reason},
  web_search: {query}, web_fetch: {query, selector, pattern},
  todo_write: {'todos.content'}, channel_edit: {text},
}
```

Unknown tools inherit nothing. Verified: `{ content | body | prompt | name: "<dir>/data.json" }` on `plugin_reader` now all throw `ambiguous local file operand`; the four first-party tools (plus `channel_edit.text` and `web_fetch.{query,selector,pattern}`, which an Oracle pass flagged as also-regressing and I added) still pass their prose through unchanged.

`fileOperands.input` still wins first; `output`/`destructive` still win on overlap.

## Hole 2 — leading-whitespace `file:` bypass → trim-aware detection
Centralized a `trimmedFileUri`/`isFileUri` helper and applied trimming everywhere `file:` is detected: the pinning scanner's URI checks, `normalizeFileReference`, `isExplicitNonFileUrl` (so `"  https://…"` isn't falsely ambiguous either), **and** the read-guard's `walk()`/`normalizeCandidate` — because the same untrimmed check let `"  file:///agent/memory/…"` slip past the *hidden-dir* deny too. The original untrimmed string is preserved for result restoration; the trimmed URI is what gets normalized/pinned. Verified: `  file://`, `\tfile://`, `\nfile://`, and `FILE://` variants now pin (or, to a hidden surface, deny).

## On the `url`/`file-uri` split
Dropped it entirely — simpler and safe. `url` is absent from `PROSE_OPERANDS`, so it falls to the heuristic, which pins an explicit `file:` URI. The existing `custom_system_reader` symlink-swap test still passes (a `file://` under `url` still snapshots to the immutable copy).

Full suite green (`11418 pass / 0 fail`), lint/format clean, typecheck clean apart from the pre-existing `kakaotalk` upstream-type errors. New regressions cover both hostile cases plus the fail-closed control. Ready for re-review.

devxoul:
## Re-review — commit `5e83114b` resolves both holes ✅

The new commit *"scope prose exemption per operand-path and trim file: detection"* directly closes both merge-blockers from my previous pass. Re-verified against `5e83114b`: typecheck + lint (0 errors) + format clean, all 253 tests across the four security suites green, all CI checks (incl. Windows + flake-guard) now passing, plus fresh adversarial probes.

### Hole 1 — FIXED (tool-scoped, not key-scoped)
`classifyFreeTextField`/`NON_PATH_KEYS` reuse is gone; replaced with an exact `tool + operand-path` allowlist:

```ts
const PROSE_OPERANDS = {
  spawn_subagent: new Set(['prompt', 'description']),
  skip_response:  new Set(['reason']),
  web_search:     new Set(['query']),
  web_fetch:      new Set(['query', 'selector', 'pattern']),
  todo_write:     new Set(['todos.content']),
  channel_edit:   new Set(['text']),
}
```

An undeclared plugin/MCP reader using a common key now fails closed again — verified:
```ts
for (const key of ['content','body','prompt','name','description','reason','query','text'])
  enforceAndPinToolFiles({ tool: 'plugin_reader', args: { [key]: `${agentDir}/secrets.json` }, agentDir, genericInputs: true })
  // → all reject: "ambiguous local file operand"  ✅
```
`url` correctly stays OUT of the list, so a `file:` URL under `url` still pins (TOCTOU preserved).

### Hole 2 — FIXED (trim before `file:` detection)
New `isFileUri()` trims first; applied at `propertyTarget`/`arrayTarget` `uri:`, in `collectGenericFileTargets`, `isExplicitNonFileUrl`, `normalizeFileReference`, and mirrored in the private-surface-read guard via `trimmedFileUri()`. Original untrimmed string preserved for result restoration. Verified:
```ts
// canonical secret via `  file://…`, `\t file://…`, `FILE://…` → all DENIED
// non-secret via same variants → all PINNED to /tmp copy (never passed through)  ✅
```

### Completeness audit of `PROSE_OPERANDS`
Cross-checked every first-party tool schema that reaches the generic scan:
- **Complete** — no missing prose operand. Other string-arg tools are safe by construction: `channel_send`/`_reply`/`_fetch_attachment`/`post_github_review` via earlier returns; `channel_react.emoji`, `channel_read`/`_history` IDs/cursors, `subagent_*` task ids, `stream_snapshot.target_id` are never path-shaped.
- **Correct** — no entry is a real file operand; none are dereferenced.
- **Names match** the registered `defineTool` names exactly; `todos.content` matches the array-recursion operand path.
- Built-ins (`bash`/`read`/`write`/`edit`/`grep`/`find`/`ls`/`look_at`) are unaffected — `wrapBuiltinToolDefinition` omits `genericInputs`, so they never hit the generic scan.

### Non-blocking nit (pre-existing, not this PR)
`grant_role.permission` values like `"channel.respond"` match the `name.ext` heuristic in `isAmbiguousUndeclaredLocalOperand` and survive only because the `lstat` misses (no such file). If a file by that name ever existed in the agent dir it'd throw. Pre-existing, unrelated to prose-scoping — worth a follow-up but not a blocker here.

**Verdict: LGTM — ship it.** Both reported holes are closed with tests that lock in the hostile cases, and the exemption is now scoped tightly enough that it can't leak to unknown tools.

<sub>Re-reviewed by Sisyphus on behalf of the (still-blocked-on-`main`) `typeey` reviewer; verified against PR HEAD `5e83114b` with typecheck/lint/format, the full security suites, and adversarial probes.</sub>

PR #1233, opened 2026-07-16
fix(sandbox): stop blocking bash on a root ref left dangling by gc

devxoul:
## Background

On a live agent, the canonical-secret Git guard (`src/git/secret-history.ts`) disabled **all** model-driven bash — PR review, channel snapshots, everything — and kept doing so even after the operator ran the exact `git gc --prune=now` the error message tells them to run. The bot repeated the same refusal every turn ("the integrity scan blocks on unreachable Git object `909aefd…^{}`; purge history, expire reflogs, run gc, restart").

Root cause is a stale pseudoref, not a real secret. A reset/rebase writes the previous tip into `.git/ORIG_HEAD`; a later pruning `git gc` **deletes that object but never rewrites `ORIG_HEAD`**, leaving it dangling at an oid now absent from the object DB. `scanNonCommitRefRoots` enumerates that oid (via `for-each-ref --include-root-refs`, which on Git ≥ 2.45 lists `ORIG_HEAD`) and peels it with `git rev-parse <gone>^{}`. That exits non-zero, `runGit` throws `GitSecretHistoryError`, and the **entire** scan fails closed. `git gc` is what *creates* the dangling ref, so the suggested operator sequence can't fix it — only a full history rewrite of `ORIG_HEAD` would, which no operator does.

## Summary

Probe `git cat-file -e <oid>` before peeling each root ref, and skip a root whose object is **absent** instead of failing the whole scan closed on the peel error. This is the same benign-history-rewrite-debris class the unreachable tree/blob fixes (#1229, #1231) already carved out: a gone object has no content and no tree, so it cannot conceal a canonical secret path.

**Why only exit code 1 counts as "absent":** `cat-file -e` exits `0` for present, exactly `1` for a valid-but-missing oid, and `128` for operational errors (corruption, bad invocation). Collapsing every non-zero into "absent" would let a corrupt object read as benign — so any exit code other than `0`/`1` still throws and fails closed, like every other scan step.

**Why `GIT_NO_LAZY_FETCH=1`:** `cat-file -e` on a partial clone can trigger a promisor network fetch. The scan is a point-in-time local check; it must never reach out to a remote. The env is added to the shared `spawnGit` so the whole scan is fetch-free.

## What this does NOT do

- **Does not weaken the non-commit invariant.** A root ref/HEAD/pseudoref whose object still **exists** and is a non-commit (tree/blob, or a tag peeling to one) still fails closed — the existence probe returns `true` and control falls through to the unchanged `peelTag` check. Only genuinely-missing objects are skipped.
- **Does not touch the reachable / reflog / unreachable-commit path scans** that attribute real canonical secrets by their root tree.
- **Does not add per-agent state or clear the in-memory `contaminatedCache`.** The cache still fail-closes for the process lifetime; this fixes the *scan* so a clean repo is no longer misclassified in the first place.

## Review notes

- Load-bearing change is the single `if (!(await objectExists(...))) continue` guard in `scanNonCommitRefRoots`, plus the exit-code-1-only semantics of `objectExists`. Verify: an **existing** non-commit root still returns `{ ok: false, paths: ['unattributable dangling Git objects'] }` (existing tests at `secret-history.test.ts:241-374` cover live refs, FETCH_HEAD, linked worktrees, and CUSTOM_HEAD).
- New regression test reproduces the production failure with a real `git reset --hard HEAD^` → `reflog expire` → `git gc --prune=now`, asserting `ORIG_HEAD` now dangles (`cat-file -t` rejects) and the scan passes.
- The all-zero oid is filtered before the probe, so the second test uses a nonzero absent oid (`'1'.repeat(40)`) to actually exercise `objectExists`.
- Full suite green: 11438 pass / 0 fail; typecheck, lint (0 errors), format:check clean.

### DIFF
--- a/src/git/secret-history.test.ts
+++ b/src/git/secret-history.test.ts
@@ -1,7 +1,9 @@
 import { afterEach, beforeEach, describe, expect, test } from 'bun:test'
-import { mkdir, mkdtemp, rm, writeFile } from 'node:fs/promises'
+import { appendFile, chmod, mkdir, mkdtemp, rm, writeFile } from 'node:fs/promises'
 import { tmpdir } from 'node:os'
-import { dirname, isAbsolute, join } from 'node:path'
+import { delimiter, dirname, isAbsolute, join } from 'node:path'
+
+import { isWindows } from '@/shared'
 
 import {
   GitSecretHistoryError,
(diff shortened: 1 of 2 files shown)


devxoul:
Addressed in the latest push (folded into the original commit).

You're right that `GIT_NO_LAZY_FETCH` alone doesn't hold on unpatched Git < 2.45 — it was introduced in 2.45 (and backported to the 2.39.4/2.40.2/2.41.1/2.42.2/2.43.4/2.44.1 maintenance releases), but an unpatched 2.44.0 silently ignores it and proceeds to the internal `git fetch`.

Rather than gate on a Git version (distros backport independently, so a version check is unreliable), `spawnGit` now also sets an **empty `GIT_ALLOW_PROTOCOL`**. That's a since-2.20 transport whitelist: an empty value denies every transport (`file`, ssh, http(s), git, ext, custom helpers), overriding any `protocol.<name>.allow` repo config. The internal lazy fetch is an ordinary child `git fetch` that inherits the env, so it fails closed at the transport layer before contacting a promisor remote — on every Git generation that has the partial-clone lazy-fetch path. Both controls are kept: `GIT_NO_LAZY_FETCH` gives cleaner missing-object behavior on patched Git, `GIT_ALLOW_PROTOCOL=''` is the all-version backstop.

Regression test added: it emulates unpatched Git < 2.45 with a `git` PATH-shim that strips `GIT_NO_LAZY_FETCH`, points a promisor remote at a local bare repo, references a missing object from `ORIG_HEAD`, and asserts via `GIT_TRACE` that the lazy `fetch origin` is attempted but **`git upload-pack` never runs** (the network is never touched). Removing `GIT_ALLOW_PROTOCOL=''` makes that test fail with `upload-pack` present in the trace, confirming it guards the fix.

I did not add partial-clone detection or a Git ≥ 2.45 requirement — the empty-whitelist fence covers the compatibility path without either.

PR #1235, opened 2026-07-16
init: add Slack confirmation-code login for 2FA workspaces

gtg7784:
## Background

Slack user-account sign-in uses the "Sign in on mobile" QR: the CLI takes a QR data URL, follows the redirect chain over HTTP, and captures the `d=xoxd-` session cookie. That works for open workspaces, but on a workspace that enforces a confirmation code the QR flow dead-ends: the redirect chain reaches an "Enter your authentication code" page and no session cookie is ever issued.
The upstream `loginWithQr` classifies this as `awaiting_device_confirmation` and throws with "use `agent-slack auth extract` on an already-authenticated device."

The catch: the code *is* delivered to the user's phone. Opening the QR sign-in URL already sends it.
So the sign-in is completable over HTTP — the missing step is submitting that code. The previous flow just never asked for it.

Root cause, confirmed by reverse-engineering a real login (HAR capture + live probe): the "Enter your authentication code" page is a React app with no `<form>`.
It embeds sign-in state as an HTML-escaped JSON blob in `<div id="props_node" data-props="{...}">`, most importantly a second z-app secret `magicLogin`.
Completing sign-in is a POST of that secret plus the user's code back to the workspace-host z-app path, which returns 302 with the `d=xoxd-` cookie.

## Summary

Add `loginWithConfirmationCode` (`src/init/slack-confirmation-login.ts`): open the QR sign-in URL, follow redirects onto the workspace host, parse `magicLogin` + `emailAddress` from the page's `data-props`, then POST the user-entered code to obtain the session cookie and reuse the existing `/ssb/redirect` scrape for the `xoxc-` token. `runSlackConfirmationBootstrap` wraps it to persist secrets, mirroring `runSlackBootstrap`. The `channel add` slack case now prompts for email then the code.

Key decisions:

- **Why read `magicLogin` from `data-props` and not hardcode a request shape.**
  The code-submit contract (query params, `2fa_*` body fields) is captured from a real login and is stable, but the `magicLogin` secret is per-attempt and server-issued.
  Reading it from the server-rendered page means only the user's email and code are ever injected — nothing about the secret is guessed.
- **Why a new typeclaw module instead of patching upstream.** 
  The login primitive lives in the `agent-messenger` dependency, which exposes only `loginWithQr`/`decodeSlackQr` and has no code-entry flow or patch mechanism.
  The small `/ssb/redirect` → `xoxc-` token step is reimplemented locally rather than reaching into a non-exported deep path.
- **Why the code prompt pauses the spinner.** 
  `runSlackAuth` runs inside the `channel add` progress spinner.
  The `requestCode` callback pauses it via the existing `LineAuthSpinnerHolder` (the same pattern LINE/Instagram use for mid-auth prompts), takes the code, then resumes.

## What this does NOT do

- Does not remove or change the existing QR-only path (`runSlackBootstrap`) or the bot-token path; both are untouched.
- Does not handle SSO/IdP workspaces — those still can't complete over HTTP and keep pointing at `agent-slack auth extract`.
- Does not add a "resend code" step or backup-code (`twoFactorBackupType`) entry; it consumes the code Slack sends when the sign-in page is opened.
- Does not rewire the `typeclaw init` wizard's `SlackAuthRunner` signature; only the `channel add` closure opts into the code flow.

## Review notes

- Load-bearing change is `parseSigninProps` + the code-submit POST in `loginWithConfirmationCode`. Invariant to verify: the POST targets the workspace-host z-app path with `?domain&domainLogin=1&email`, body carries the server-issued `2fa_magiclogin` (not the QR's own secret) plus `2fa_action=submit_primary`, and the flow only succeeds once a `d=xoxd-`
  cookie is present.
- `decodeHtml` intentionally decodes `&amp;` last so `&amp;quot;` doesn't collapse into a stray quote.
- Cookies are only ever sent to Slack hosts (`isSlackHost`); a redirect off Slack stops the flow without leaking the session cookie.
- Verified end-to-end against a real 2FA workspace: session cookie and `xoxc-` token both issued. Unit tests replay the redirect → data-props → code-submit chain with a mock `fetch`.

### DIFF
--- a/src/cli/channel.ts
+++ b/src/cli/channel.ts
@@ -41,7 +41,7 @@ import { runDiscordBootstrap } from '@/init/discord-auth'
 import { runInstagramBootstrap, type InstagramLoginCallbacks } from '@/init/instagram-auth'
 import { runKakaotalkBootstrap } from '@/init/kakaotalk-auth'
 import { runLineBootstrap } from '@/init/line-auth'
-import { runSlackBootstrap } from '@/init/slack-auth'
+import { runSlackConfirmationBootstrap } from '@/init/slack-auth'
 import { runTeamsBootstrap, type TeamsDeviceCodeCallbacks } from '@/init/teams-auth'
 import { runWebexBootstrap } from '@/init/webex-auth'
 import { SecretsKakaoCredentialStore } from '@/secrets/kakao-store'
@@ -1061,10 +1061,18 @@ async function collectCredentials(
     }
     case 'slack': {
       const qrDataUrl = await promptSlackQrDataUrl()
+      const email = await promptSlackEmail()
+      const spinnerControl = lineSpinnerHolder ? holderSpinnerControl(lineSpinnerHolder) : undefined
       return {
         channel,
         slackQrDataUrl: qrDataUrl,
-        runSlackAuth: ({ cwd: agentDir, qrDataUrl }) => runSlackBootstrap({ qrDataUrl, agentDir }),
+        runSlackAuth: ({ cwd: agentDir, qrDataUrl }) =>
+          runSlackConfirmationBootstrap({
+            qrDataUrl,
+            agentDir,
+            email,
+            requestCode: () => promptSlackConfirmationCode(spinnerControl),
+          }),
       }
     }
     case 'telegram-bot':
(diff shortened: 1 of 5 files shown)


devxoul:
@typeey can you also check if it's more related to agent-messenger layer?

devxoul:
@gtg7784 yo can you contribute to [agent-messenger](https://github.com/agent-messenger/agent-messenger) which powers TypeClaw?

gtg7784:
> @gtg7784 yo can you contribute to [agent-messenger](https://github.com/agent-messenger/agent-messenger) which powers TypeClaw?

oh yeah sure

PR #1261, opened 2026-07-20
fix(tests): make session /tmp redirect tests unique per process

devxoul:
## Background

CI's **flake guard (oversubscribed parallel)** job intermittently fails with a single test failure while the run reports `0 fail` overall — the "a worker died" signature. It failed on an unrelated PR ([run 29759779728](https://github.com/typeclaw/typeclaw/actions/runs/29759779728/job/88411210075)), confirming a flake rather than a regression:

```
(fail) wrapBuiltinToolDefinition /tmp path redirect (per-session scratch) >
  a sandboxed role (guest) has its /tmp write redirected to the session backing dir
```

**Root cause.** The flake guard deliberately runs **two full `bun test --parallel` processes at once** in the same container, both sharing the real `/tmp`. The session `/tmp`-redirect tests hardcoded their `sessionId` as module constants (`sid42-guest-write`, `sid42-guest-read`, `sid42-owner-write`, `sid42-owner-read`), which map to a **fixed absolute path** under `SESSION_TMP_ROOT` (`/tmp/typeclaw-session/<sessionId>`). Because the path is identical across both OS processes, they race on the same backing directory:

- Process A opens the anchored parent fd (`/proc/self/fd/<n>`) and is mid write/read.
- Process B's copy of the same test hits its `finally { rm -rf }` and unlinks the shared dir.
- Process A's `readFile` then hits ENOENT, or `realpath(anchored parent fd)` resolves to `".../<sessionId> (deleted)"` — the single observed failure.

Intra-process `--parallel` isolation doesn't help: the collision is **between two separate OS processes** sharing one real `/tmp`.

## Summary

Namespace each backing directory by `process.pid` + `randomUUID()` via a local `uniqueSessionId(label)` helper in `src/agent/plugin-tools.test.ts`, so the two concurrent suites never target the same path. This is the pattern `review-checkout.test.ts` already uses (`review-checkout-${process.pid}`) for the exact same shared-`/tmp` hazard; this PR brings the redirect tests in line and adds `randomUUID()` on top of `pid` to also cover PID reuse and crashed-run artifacts.

**Why per-test unique IDs and not an env override for `SESSION_TMP_ROOT`:** keeping the real production root means the redirect + inode-anchoring behavior stays fully exercised; an override would add production surface and mutable process-global test state without fixing anything better.

While here, the two "receipt" tests (`fakeWriteEchoingPath`) also got `try/finally` cleanup. Their fake tool never writes, but the wrapper still calls `ensureSessionTmpDir()`, so they were **leaking** a fixed `sid42` directory into the shared `/tmp` on every run.

## What this does NOT do

- No production code changes — `src/sandbox/session-tmp.ts` and the anchoring logic are untouched.
- Does not change the flake-guard workflow or the `SESSION_TMP_ROOT` contract.
- The `sid42` id in the "non-`/tmp` write" test is intentionally left as-is: it uses its own `mkdtemp` agentDir and never touches `SESSION_TMP_ROOT`.

## Review notes

- Load-bearing change: the hardcoded `sid42-*` sessionIds → `uniqueSessionId('...')`, and every assertion/cleanup now references a local `sessionDir` bound to that unique id.
- Invariant to verify: no test in this describe block writes to a `SESSION_TMP_ROOT` path that another concurrent process could also produce.
- The inode-anchoring tests are `skipIf(lacksInodeAnchoring)` (Linux-only), so they run only on the Linux CI; the receipt-test cleanup fix is verifiable on any platform.
- Verified locally on macOS: full suite green (`0 fail`), and 3 iterations of the oversubscription harness (two concurrent runs of this file) all passed with no leaked shared dirs.

### DIFF
--- a/src/agent/plugin-tools.test.ts
+++ b/src/agent/plugin-tools.test.ts
@@ -1,4 +1,5 @@
 import { afterEach, beforeEach, describe, expect, test } from 'bun:test'
+import { randomUUID } from 'node:crypto'
 import {
   link,
   mkdir,
@@ -3726,10 +3727,18 @@ describe('wrapBuiltinToolDefinition /tmp path redirect (per-session scratch)', (
   const tui: SessionOrigin = { kind: 'tui', sessionId: 's' }
   const guest: SessionOrigin = { kind: 'subagent', subagent: 'x', parentSessionId: 'p', spawnedByRole: 'guest' }
 
+  // Backing dirs live on the REAL process-shared container /tmp (SESSION_TMP_ROOT).
+  // The oversubscription flake guard runs two `bun test` processes at once; a fixed
+  // sessionId makes both race the same dir, so one's `rm -rf` unlinks it while the
+  // other still holds an anchored parent fd. pid separates the suites; randomUUID
+  // also guards PID reuse and crashed-run artifacts.
+  const uniqueSessionId = (label: string): string => `plugin-tools-${label}-${process.pid}-${randomUUID()}`
+
   test.skipIf(lacksInodeAnchoring)(
     'a sandboxed role (guest) has its /tmp write redirected to the session backing dir',
     async () => {
-      const sessionId = 'sid42-guest-write'
+      const sessionId = uniqueSessionId('guest-write')
+      const sessionDir = `${SESSION_TMP_ROOT}/${sessionId}`
       const record: { path?: string; resolvedParent?: string } = {}
       const wrapped = wrapBuiltinToolDefinition(fakeWrite(record), {
         agentDir: '/agent',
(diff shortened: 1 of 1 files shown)


devxoul:
@typeey Good catch — you're right, the original description was completely wrong: it described a nonexistent `tool-error-remediation.ts` feature that isn't in this diff at all. The actual change is only `src/agent/plugin-tools.test.ts` (+65/-41).

I've replaced the entire description with the accurate `/tmp` collision rationale: the flake-guard runs two `bun test --parallel` processes sharing the real `/tmp`, and the hardcoded `sid42-*` session IDs made both processes race the same `SESSION_TMP_ROOT` backing dir (one's `rm -rf` unlinking it under the other's anchored parent fd). Fix namespaces each dir by `process.pid` + `randomUUID()`, plus `try/finally` cleanup for the two receipt tests that were leaking a fixed `sid42` dir. Verification for the real change is in the updated body. Thanks for keeping the merge record honest.

devxoul:
@typeey so you're approving?

PR #1265, opened 2026-07-21
Write the git-askpass helper to a runtime-writable temp dir, not /usr/local/bin

devxoul:
## Problem

The `GIT_ASKPASS` helper (`src/bundled-plugins/github-cli-auth/git-askpass.ts`) wrote to a hardcoded `/usr/local/bin/typeclaw-git-askpass`. When the runtime runs as a **non-root user**, `/usr/local/bin` isn't writable, so `ensureGitAskPassHelper` failed with `EACCES` and the caller aborted.

Observed live: a code-review agent's scratch checkout failed with
```
Reviewer checkout failed: EACCES: permission denied, open '/usr/local/bin/.typeclaw-git-askpass.<hex>.tmp'
```
so the reviewer could not acquire the PR and (correctly) declined to post a groundless verdict. The block was **not** the git-secret-history gate, tool naming, or bash policy — it was purely this helper write failing.

## Why /usr/local/bin was wrong here

Its two callers on `main` — the **reviewer scratch checkout** (`review-checkout.ts`) and the **backup push** (`backup/git-auth.ts`) — both run `git` via Node `execFile`, **not** sandboxed bash. So the helper only needs to be **runtime-writable**; it does **not** need to sit under a sandbox-visible (`--ro-bind /usr`) mount. The old path optimized for a sandbox consumer that doesn't exist on `main` (authenticated git in model bash is currently blocked), at the cost of breaking the two real consumers whenever `/usr` is read-only for the runtime user.

## Fix

- **Default to a process-private temp dir** (`mkdtemp` under `os.tmpdir()`, mode `0700`) instead of `/usr/local/bin`. Always runtime-writable, and out of any **model-writable** tree (agentDir / session `/tmp`) — so a tool can't plant a script that captures a later runtime-issued token.
- **Keep `TYPECLAW_GIT_ASKPASS_PATH` as a verbatim override** with **no silent fallback**: an unusable configured path surfaces loudly rather than degrading to a path the caller never asked for. (A future sandbox consumer — e.g. #1263's authenticated-git-in-bash — sets this and binds an exact path in.)
- **Key the write-once cache by resolved path** (`Map` instead of one shared promise): the old shared promise returned the *first* caller's path even when a later caller explicitly requested a different one.

## Security invariants preserved

Unpredictable temp name, exclusive `wx` create (symlink-hijack safe), `chmod 0755`, atomic rename, github.com-only prompt guard, and token only ever in `TYPECLAW_GIT_TOKEN` — never in the script, argv, git config, or logs.

## Tests

- default path is outside `/usr` and is writable + executable (the regression this fixes);
- two distinct explicit paths are ensured independently (the cache-keying fix);
- all existing host-scoped prompt-guard tests unchanged.

## Verification

- `bun run typecheck` ✅
- `bun run lint` ✅ (0 errors)
- `bun run format:check` ✅
- `bun run test` ✅ (11703 pass, 0 fail)

### DIFF
--- a/src/bundled-plugins/github-cli-auth/git-askpass.test.ts
+++ b/src/bundled-plugins/github-cli-auth/git-askpass.test.ts
@@ -49,17 +49,39 @@ describe.skipIf(onWindows)('ensureGitAskPassHelper', () => {
     expect(mode).not.toBe(0)
   })
 
-  test('defaults to a runtime-writable path outside /usr so a read-only /usr does not EACCES', async () => {
-    const original = process.env.TYPECLAW_GIT_ASKPASS_PATH
+  test('defaults under the fixed real /tmp (writable, executable), not read-only /usr', async () => {
+    const originalOverride = process.env.TYPECLAW_GIT_ASKPASS_PATH
     delete process.env.TYPECLAW_GIT_ASKPASS_PATH
     try {
       const path = await ensureGitAskPassHelper()
       expect(path.startsWith('/usr')).toBe(false)
+      expect(path.startsWith('/tmp/typeclaw-git-askpass-')).toBe(true)
       expect((await stat(path)).isFile()).toBe(true)
       expect((await stat(path)).mode & 0o111).not.toBe(0)
     } finally {
-      if (original === undefined) delete process.env.TYPECLAW_GIT_ASKPASS_PATH
-      else process.env.TYPECLAW_GIT_ASKPASS_PATH = original
+      if (originalOverride === undefined) delete process.env.TYPECLAW_GIT_ASKPASS_PATH
+      else process.env.TYPECLAW_GIT_ASKPASS_PATH = originalOverride
+    }
+  })
+
+  test('ignores a model-writable TMPDIR: the default base stays on the fixed real /tmp', async () => {
+    // os.tmpdir() honors TMPDIR/TMP/TEMP; if the default were derived from it, a tool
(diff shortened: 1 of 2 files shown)


devxoul:
Addressed in d411a3da — thanks, this was a real gap.

**The fix:** the default base is now the **fixed literal `/tmp/typeclaw-git-askpass-`** (via `mkdtemp`), not `os.tmpdir()`. You're right that `os.tmpdir()` honors `TMPDIR`/`TMP`/`TEMP`, so a model-writable `TMPDIR` could have relocated the cached helper into a tree a sandboxed tool can write, then swapped it before an unsandboxed reviewer checkout / backup push executes it with `TYPECLAW_GIT_TOKEN`.

This matches the existing `SESSION_TMP_ROOT = '/tmp/typeclaw-session'` convention: bwrap binds a **fresh tmpfs over `/tmp` per session**, so the runtime's real `/tmp/typeclaw-git-askpass-*` is not visible to model bash and can't be pre-planted. The random `mkdtemp` subdir (0700) keeps the name unpredictable; `TYPECLAW_GIT_ASKPASS_PATH` stays the trusted verbatim override with no fallback.

**Regression test added** (as requested): points `TMPDIR` at a model-writable dir and asserts the default base stays on the fixed real `/tmp`, not `TMPDIR`.

All checks green (typecheck / lint / format:check / test 11704 pass). PTAL.

PR #1278, opened 2026-07-21
Stop brokering App tokens to model-driven Git

devxoul:
## Summary

- Delete `handleGitCommand` and its `GIT_ASKPASS`/`GIT_CONFIG_*` token-overlay wiring from `src/bundled-plugins/github-cli-auth/index.ts` — model-driven bash no longer gets a per-repo App token injected into a bare `git` invocation.
- Delete `analyzeGitCommand` and the rest of the model-shaped git command analyzer in `git-command.ts` (~793 lines) and its tests (~661 lines) — dead code once the injection path it fed is gone.
- Add `containsGhCommandInvocation` (tokenize + scan for an actual `gh` invocation) in `gh-command.ts` and use it in place of the lexical `command.includes('gh')` check, so a command that merely contains the substring "gh" (e.g. in a path or argument) no longer falsely triggers `gh`-command handling.
- Retain the validated `gh` brokerage path unchanged: repo-scoped App-token minting for allowlisted `gh` commands, PAT pass-through, and the sandboxed-PAT-withheld guidance.
- Ordinary tokenless and operator-credential Git (no App token involved) is untouched — this only removes the model-shaped App-token-to-git bridge, not Git itself.

## Why

Unlike `gh`, `git` runs repo-local hooks and credential helpers, so a token handed to a bare `git` invocation is not confined the way a validated `gh` command is — the analyzer that gated this was a large, hard-to-audit surface (nearly 1,700 lines between implementation and tests) protecting a path whose risk profile never matched the narrower, already-validated `gh` brokerage. Removing the model-driven git injection removes that structurally unsafe path rather than trying to further constrain it, while the `gh` brokerage — which analyzes a real command shape before minting — stays exactly as validated as before. The `command.includes('gh')` fix is a correctness fix surfaced while auditing this path: a lexical substring match could mis-trigger on unrelated commands.

## Behavior change (policy reversal)

This is a security-policy reversal, not just a refactor, so it's worth stating plainly: **TypeClaw no longer supplies channel App authentication to model-driven plain `git clone`, `fetch`, `pull`, `push`, or `ls-remote`.** The prior design brokered a per-repo App token into these through `GIT_ASKPASS`, gated by an analyzer that tried to constrain the command shape (single-bare-command, github.com only, one narrow "clone then inspect" exception) to keep the token from leaking to a sibling process.

That analyzer was fighting the wrong battle. Unlike `gh`, a plain `git` invocation's actual process tree isn't fully described by its command-line shape: hooks (`pre-push`, `post-checkout`, …), clean/smudge filters, `core.fsmonitor`, aliases, credential helpers, `commit.template`, submodule URLs, `insteadOf`/`pushInsteadOf` URL rewrites, and `GIT_TRACE*` output can all execute or read arbitrary content from inside git's own process tree — none of it visible to a denylist-style parse of the invoking command string. Auditing and closing every one of those escapes individually is a losing, ever-growing list; removing the token from git's reach entirely closes the whole class at once instead of chasing the next escape.

**Supported alternatives going forward:**
- **Validated, repo-scoped `gh` calls** for supported GitHub API/issue/PR operations — this brokerage is unchanged by this PR and still mints a least-privilege, per-repo App token for allowlisted commands.
- **`github_prepare_review_checkout`** for runtime-owned checkout of a private PR under review: the runtime verifies one allowlisted `owner/repo` and a full commit SHA, exposes the token only to a host-locked fetch process, then performs the checkout in a separate token-free process — model bash never sees the token.
- **Host-stage Git, or credentials the operator independently and deliberately exposes** (e.g. a PAT/SSH key declared in `.env`, which the sandbox already inherits into bash for every role) for any other network Git need.

**Unaffected:** ordinary local, tokenless Git — `status`, `diff`, `log`, `add`, `commit`, `checkout` of already-fetched refs, `stash`, and any network Git already running with a credential the operator independently exposed — none of that goes through this brokerage and none of it changes here.

## Verification

Focused suite (`github-cli-auth` gh-command / index / git-command remnants): 66 tests passed on repeated local runs. Top-level full suite: 11,738 passed, 30 skipped.

## Stack

Second PR in a 3-PR stack. PR 1 (Keep GitHub App tokens out of the process environment) has merged, so this PR's base is now `main`. PR 3 (Remove obsolete model Git broker artifacts) remains based on this branch (`fix/model-git-token-broker`):
1. Keep GitHub App tokens out of the process environment — merged.
2. **This PR** — remove the model-driven git token broker.
3. Remove obsolete model Git broker artifacts (base: this branch).


devxoul:
Closing this one. Authenticated, App-token-brokered Git for model-driven `git clone`/`fetch`/`pull`/`push`/`ls-remote` against private repos is a capability we need to keep, so removing the broker isn't something we want to ship even with the wider process-tree concerns raised in the description.

The narrower correctness fix bundled here — `containsGhCommandInvocation` replacing the `command.includes('gh')` substring check — is good on its own merits; happy to see it land in a follow-up PR scoped to just that, decoupled from the broker removal.

No action needed from reviewers on this one — closing as a product/security-policy call, not a review outcome.

PR #1307, opened 2026-07-24
Warn on a low macOS open-files limit in doctor

devxoul:
## Background

A user reported OrbStack crashing with a **"Stopped unexpectedly: killed (SIGKILL)"** dialog while running typeclaw agents. The log excerpt showed, on repeat:

```
http: Accept error: accept unix ~/.orbstack/run/vmcontrol.sock: accept: too many open files; retrying in {5ms…1s}
vmgr | level=warning msg="DNS query failed" error="no memory" name=slack.com./chatgpt.com./api.github.com. type=A
```

Investigating the affected host directly established the root cause is **host-side OrbStack resource exhaustion, not a typeclaw FD leak inside the container**:

- `launchctl limit maxfiles` = **256** (soft) — the untouched macOS default. There is no `/Library/LaunchDaemons/limit.maxfiles.plist` raising it, and no shell rc lowering it. GUI apps launched by launchd (OrbStack, Docker Desktop) inherit this 256 soft limit unless they raise it themselves.
- The kernel ceiling is fine (`kern.maxfilesperproc` = 61440); only the per-process launchd soft limit is 256.
- OrbStack's `vmgr` helper was at **6.4–6.6 GB RSS** with the host under heavy memory pressure (~200 MB free). Under that load, `vmgr` accumulates FDs, crosses 256, and fails `accept()` on `vmcontrol.sock` with `EMFILE` → the backoff storm → SIGKILL. The DNS `error="no memory"` is OrbStack's gVisor netstack returning `ENOMEM` under the same host memory pressure.
- The typeclaw containers themselves were healthy internally (~28 FDs each), and the code is already FD-hardened (Teams adapter avoids reconnect storms; `tool-file-safety.ts` destroys ReadStreams; `session-list.ts` bounds concurrency).

## Summary

typeclaw can't raise the host's launchd limit or fix OrbStack itself, so this surfaces the risk where operators already look: `doctor`.

A macOS-only check (`hostd.macos-maxfiles`) parses `launchctl limit maxfiles` and **warns when the soft limit is below 1024**, following the existing host-environment-warning pattern (`wslDriveMount`, `windowsSecretPerms`, `windowsBindMount`).

Key decisions:
- **Framed as a risk indicator, not proof.** The check can't read OrbStack's live `RLIMIT_NOFILE`, so the wording says the session default is inherited-and-suspicious, not that the runtime definitely hit it. The kernel ceiling and per-container limits are called out as separate.
- **Non-privileged remediation only.** The fix guidance is "update + relaunch the runtime; if it recurs, raise the login-session limit via your org's supported method." It never edits launchd files or runs privileged commands.
- **Fails safe.** Unreadable/unparseable `launchctl` output degrades to `info`, never `error` — so a non-macOS host or a locked-down box never turns this into a red check.
- The `launchctl` runner is injected as a dependency so tests never shell out; the output parser is a pure exported function with its own table of cases.

## What this does NOT do

- Does **not** add a per-container `--memory` cap or `--ulimit nofile`. Per the investigation, a container `nofile` bump is a non-fix (the container side is already fine; it doesn't touch the host `vmgr` limit), and a default memory cap risks OOM-killing Chrome/embeddings. Those belong in a separate resource-governance proposal, not this bug fix.
- Does **not** attempt to change any host limit automatically.
- Does **not** claim a specific required threshold as fact — 1024 is a conservative "well above the 256 default" line, not a measured OrbStack requirement.

## Review notes

- Load-bearing file: `src/doctor/checks.ts` — `macosMaxFiles()` + `parseLaunchctlMaxFilesSoft()`. The parser is the piece to scrutinize: it targets the exact `\tmaxfiles    <soft>    <hard>\n` shape, maps `unlimited` → `Infinity`, and returns `null` on anything unrecognized (→ `info`).
- Registered in `buildStaticChecks()` alongside the other `hostd` host-environment checks.
- Tests cover: non-macOS (ok), the 256 default (warning), a high limit (ok), `unlimited` (ok), unreadable (info), unparseable (info), registration, plus four parser cases including a non-numeric soft field.
- Docs: a "macOS open-files limit" subsection under the doctor CLI reference documents the check and the `vmcontrol.sock` / `ENOMEM` symptom + what to capture for an OrbStack report.

### DIFF
--- a/docs/content/docs/reference/cli.mdx
+++ b/docs/content/docs/reference/cli.mdx
@@ -152,6 +152,12 @@ typeclaw doctor --only docker,config               # narrow to categories
 
 Categories: `docker`, `config`, `secrets`, `network`, `git`, plus per-plugin checks.
 
+### macOS open-files limit
+
+On macOS, `doctor` warns (category `hostd`, check `hostd.macos-maxfiles`) when the login session's open-files **soft** limit is low. macOS ships a per-process default of 256 (`launchctl limit maxfiles`), and GUI apps launched by launchd — including **OrbStack** and **Docker Desktop** — inherit it. Under load the runtime can exhaust that limit and fail with `too many open files` on its control socket (e.g. `accept unix …/vmcontrol.sock: too many open files`), which surfaces as the runtime being killed (SIGKILL), often alongside DNS lookups failing with `no memory`.
+
+This is a **risk indicator**, not proof the running runtime inherited that exact limit — the kernel ceiling (`kern.maxfiles`, `kern.maxfilesperproc`) and per-container limits are separate. typeclaw does **not** change this limit for you. To clear the warning: update your container runtime to the latest version and fully quit + relaunch it (a well-behaved runtime raises its own soft limit on start); if it recurs, raise the login-session limit using your organization's supported macOS administration method, then relaunch the runtime. If you collect diagnostics for an OrbStack/Docker report, capture `launchctl limit maxfiles`, the runtime's log around the `vmcontrol.sock` errors, and `~/.orbstack/log/` (OrbStack).
+
 ## Update
 
 ```sh
--- a/src/doctor/checks.test.ts
+++ b/src/doctor/checks.test.ts
@@ -11,9 +11,12 @@ import { buildGitignore, GITIGNORE_FILE } from '@/init/gitignore'
 import {
   buildStaticChecks,
   detectWindowsBindMountIssues,
+  macosMaxFiles,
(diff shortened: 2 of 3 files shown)


devxoul:
Closing: warning on the macOS default (256 maxfiles) is bad UX — every macOS user would be warned on a universal factory default. 256 is only the ceiling; the real problem is whatever consumes FDs up to it. Pivoting to find and fix the actual FD consumer instead of warning about the limit.

PR #1323, opened 2026-08-02
Honor MCP tool-declared fileOperands in mcp_call

### LINKED ISSUE #1322: mcp_call ignores target tool's fileOperands declaration — free-text args rejected as ambiguous file operands
## Summary

`mcp_call` (the MCP dispatcher tool) rejects free-text string arguments that look path-shaped — e.g. a search query containing a slash like `"SPPN 3/2017"` — with:

```
ambiguous local file operand at key "query"; the tool author must declare fileOperands.input or the caller must use a file: URI
```

The error message tells MCP tool authors to declare `fileOperands`, and the docs state "Plugin/MCP tools declare their own via `fileOperands.nonFile`", but the `mcp_call` dispatcher never reads the target tool's declaration, so MCP tool authors have **no way** to exempt free-text parameters.

## Root cause

In `src/mcp/tools.ts`, `createCallTool` calls `enforceAndPinToolFiles` with `tool: 'mcp_call'` and **no `fileOperands`**:

```ts
const pinned = await enforceAndPinToolFiles({
  tool: 'mcp_call',
  args: toolArgs,
  agentDir: ctx.agentDir,
  genericInputs: true,
  logger: ctx.logger,
  signal: ctx.signal,
})
```

`mcp_call` is not in the static `NON_FILE_OPERANDS` first-party table (`src/agent/tool-file-safety.ts`), so every string arg is scanned by `isAmbiguousUndeclaredLocalOperand`. Values with slashes, leading `/`, `./`/`../`, or extension-like shapes are rejected before the request reaches the MCP server — even though the target tool's schema (`McpToolInfo` in `src/mcp/client.ts`) could declare them as `nonFile` operands.

This makes it impossible for any MCP server to expose a free-text parameter that legitimately contains slashes (search queries, document numbers, regexes, paths-as-data) without the agent-side call failing.

## Repro

1. Register an MCP server exposing `search_web(query: string)`.
2. From the agent, `mcp_call` with `args: { query: "SPPN 3/2017" }`.
3. The call throws the "ambiguous local file operand" error before reaching the server.

Verified against typeclaw 0.47.2: `enforceAndPinToolFiles({ tool: 'mcp_call', args: { query: 'SPPN 3/2017' }, agentDir, genericInputs: true })` reproduces the error; passing `fileOperands: { nonFile: ['query'] }` makes it pass — the scanner itself already honors declarations, the dispatcher just never supplies them.

## Suggested fix

Plumb the target tool's declared operands through the dispatcher:

1. `src/mcp/client.ts`: carry `fileOperands` on `McpToolInfo`, read from the tool's `_meta['x-file-operands']` (survives the SDK server → wire → client round-trip; `_meta` is an arbitrary record in the MCP spec).
2. `src/mcp/tools.ts`: in `createCallTool`, resolve the target tool from `connection.listTools()` and pass its `fileOperands` into `enforceAndPinToolFiles`.

The fail-closed boundary is preserved: tools that declare nothing are scanned exactly as today.

Iman-Sharif:
Fixes #1322

## Problem

`mcp_call` scans every string argument as a potential local file operand. Free-text values that look path-shaped — e.g. a search query containing a slash, `"SPPN 3/2017"` — are rejected before the request reaches the MCP server:

```
ambiguous local file operand at key "query"; the tool author must declare fileOperands.input or the caller must use a file: URI
```

The error message tells MCP tool authors to declare `fileOperands`, and the docs in `src/skills/typeclaw-plugins/SKILL.md` state "Plugin/MCP tools declare their own via `fileOperands.nonFile`" — but the `mcp_call` dispatcher never reads the target tool's declaration, so MCP tool authors have no way to exempt free-text parameters (search queries, document numbers, regexes, paths-as-data).

## Fix

Two changes:

1. **`src/mcp/client.ts`** — `McpToolInfo` now carries `fileOperands`, read from the tool's `_meta['x-file-operands']`. `_meta` is an arbitrary record in the MCP spec and survives the server → wire → client round-trip through the official SDK (verified against `@modelcontextprotocol/sdk@1.29`).
2. **`src/mcp/tools.ts`** — `mcp_call` resolves the target tool from the connection's (cached) tool list and passes its `fileOperands` into `enforceAndPinToolFiles`, so declared `nonFile` operand paths are skipped by the scanner.

The fail-closed boundary is unchanged: tools that declare nothing are scanned exactly as before — undeclared path-shaped args are still rejected, and explicit `file:` URIs still get pinned/snapshotted.

## Test

Added `mcp_call honors the target tool's declared fileOperands.nonFile operands`: a tool declaring `fileOperands: { nonFile: ['query'] }` receives `"SPPN 3/2017"` unchanged; the existing reject tests for undeclared path-like operands still pass.

## Verification

- `bun test src/mcp/tools.test.ts src/mcp/client.test.ts` — 39 pass, 0 fail
- Reproduced the original error with the scanner directly; confirmed the declared path passes and undeclared paths still throw

### DIFF
--- a/src/mcp/client.ts
+++ b/src/mcp/client.ts
@@ -7,12 +7,17 @@ import type { Transport } from '@modelcontextprotocol/sdk/shared/transport.js'
 import { CallToolResultSchema, type CallToolResult, type ListToolsRequest } from '@modelcontextprotocol/sdk/types.js'
 
 import type { McpServer } from '@/config/config'
+import type { ToolFileOperands } from '@/plugin'
 import { resolveSecret } from '@/secrets/resolve'
 
 export type McpToolInfo = {
   name: string
   description: string
   inputSchema: unknown
+  // Declared by MCP servers via the tool's `_meta['x-file-operands']`; consumed
+  // by the mcp_call dispatcher so the file-operand scanner honors tool-authored
+  // nonFile operand paths instead of rejecting free-text args that look path-shaped.
+  fileOperands?: ToolFileOperands
 }
 
 // The SDK defaults each request to 60s; typeclaw boot should fail fast enough
@@ -28,12 +33,19 @@ export type McpConnection = {
   close(): Promise<void>
 }
 
+function readDeclaredFileOperands(meta: Record<string, unknown> | undefined): ToolFileOperands | undefined {
+  if (typeof meta !== 'object' || meta === null) return undefined
+  const declared = meta['x-file-operands']
+  if (typeof declared !== 'object' || declared === null) return undefined
+  return declared as ToolFileOperands
+}
+
 export type McpSdkClient = {
   listTools(
     params?: ListToolsRequest['params'],
     options?: RequestOptions,
   ): Promise<{
-    tools: { name: string; description?: string; inputSchema: unknown }[]
+    tools: { name: string; description?: string; inputSchema: unknown; _meta?: Record<string, unknown> }[]
     nextCursor?: string
   }>
   callTool(
@@ -112,6 +124,7 @@ export function createMcpConnection(
           name: tool.name,
           description: tool.description ?? '',
           inputSchema: tool.inputSchema,
+          fileOperands: readDeclaredFileOperands(tool._meta),
         })),
       )
       cursor = result.nextCursor
--- a/src/mcp/tools.test.ts
+++ b/src/mcp/tools.test.ts
@@ -234,6 +234,28 @@ describe('createMcpDispatcherTools', () => {
     expect(received).toEqual(args)
   })
 
+  test('mcp_call honors the target tool\'s declared fileOperands.nonFile operands', async () => {
+    let received: Record<string, unknown> | undefined
+    const connection = fakeConnection('search', [
+      {
+        name: 'search_web',
+        description: 'Web search',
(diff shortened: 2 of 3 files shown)


devxoul:
Thanks for this @Iman-Sharif — the diagnosis was right and the fix is the right shape. I've picked your commit up unchanged in #1349 and stacked the CI fix and the review follow-ups on top of it, so your authorship is preserved in the history.

What's added there:

- **Blocker** — target metadata now reaches the pre-execution guard. `wrapSystemTool` resolves the target's `nonFile` paths (rebased under `args.`) before `runToolBefore`, via a cache-only lookup so nothing connects to an MCP server ahead of the security hooks. Canonical credential denial is left genuinely unconditional.
- **Validation** — `_meta['x-file-operands']` is parsed rather than cast; malformed recognized categories discard the whole declaration fail-closed.
- **Docs** — new `reference/mcp.mdx` documenting the key, categories, dotted-path semantics, and malformed-declaration behavior.
- **CI** — `format:check` was failing the job on both runners.

The regression test now runs through the real production path (`buildMcpDispatcherToolDefinitions` + `wrapSystemTool` + the actual security plugin hook under a restricted role) and fails if the fix is removed.

Closing this in favor of #1349.

PR #1326, opened 2026-08-03
Apply renewed channel credentials without restarting the container

devxoul:
## What's wrong

`hostd` recreates the **entire agent container** every time a channel credential is renewed, purely so the in-memory adapter client picks up the new token. Every in-flight session, cron job and subagent on *every other channel* dies with it.

Captured on a live agent — a Teams token expiring destroyed an in-progress GitHub PR review:

```
03:40:49 [hostd] teams renewal tick started for paul
03:40:50 [hostd] teams renewal OK for paul (new token expires 09:40:50)
03:40:50 [hostd] teams renewal scheduled container restart for paul
03:40:51 [hostd] registered paul                       <- whole container recreated
03:40:57 [channels] pr:2321:: restart-resume ... delivering interrupted-subagent notice
03:41:03 -> bot posts "review lost due to container restart"
```

`docker inspect` confirms a full recreate (`Created == StartedAt`, `RestartCount=0`, `ExitCode=0`), not a crash.

## Why it got this bad

The mechanism was introduced for **KakaoTalk**, where it's defensible: the renewal threshold is 5 days and a live LOCO session genuinely can't rebind a credential. Webex reused it (*"same lifecycle as the KakaoTalk cron"*), then Teams (*"alongside the existing kakao/webex managers"*). Nobody re-derived the cost — and cadence is a property of the token, not the mechanism:

| adapter | renewal cadence | container restarts |
|---|---|---|
| kakaotalk | 5-day threshold | ~0.2/day |
| webex | 8h window | ~1-3/day |
| **teams** | **6h TTL / 15min window** | **~4.2/day** |

Confirmed in `hostd.log`: 15 forced restarts in 4 days, every ~5h45m, like clockwork.

## The fix

Both halves already existed and had never been connected:

- `ChannelManager.restartAdapter()` — adapter-scoped restart, already used in production by `tunnel-bridge` when a tunnel URL changes.
- The reload transport (`{type:'reload', scope}` + `docker exec` fallback) — already backs `typeclaw reload`, the agent reload tool and `/reload`.

`reload()` already *detected* the rotation via a credential signature; for personal-account adapters it just pushed a string into `restartRequired` that renders as a bare count. So hostd reached for the sledgehammer.

Now: hostd sends a `channels`-scoped reload carrying `cause: { kind: 'credential-rotation', adapter }`, which bounces **only that adapter**.

### Keeping `typeclaw reload` non-destructive

Applying a rotation drops the channel turns that adapter is serving, so it is **opt-in and singular** — the caller must name the exact adapter. A reload with no cause behaves precisely as before. This is enforced by pre-existing tests, not just new ones: mutating the guard to always apply breaks three tests that predate this PR.

### Fallback policy

| outcome | action |
|---|---|
| reload applied | done, no restart |
| both transports fail / scope errored / adapter couldn't be stopped | full container restart |
| adapter down under supervision (`recovery-pending`) | **no restart** |

That last row is deliberate: recreating the container can't fix a credential the platform rejected, and would destroy unrelated work for nothing.

Waiting for the next tick is *not* a safe fallback — the token is already persisted, so the next tick reports `fresh_enough` and never calls back, leaving the adapter on a token with ≤15 min left.

## Also fixed

- **Signature TOCTOU** (`manager.ts`) — the credential signature was read *before* `adapter.start()`, but the adapter reads the file itself *inside* `start()`. A write landing in between recorded the pre-start signature while the adapter held the newer one, so the entry looked current and the next reload saw nothing to apply. Harmless while rotation only produced a report; not harmless now that it can be applied. Own commit, own test.
- **Stale `AGENTS.md`** — claimed the Teams realtime listener is dead in containers and must run REST-only. Untrue since `ContainerTeamsClient` began minting the realtime bearer from `aad_refresh_token` instead of inheriting the SDK's desktop cookie scrape. The listener runs; prod logs show `[teams] realtime listener connected`.

## Preserved invariants

The renewal managers' guards are untouched and still asserted: stale-registration (`cwd` match), single-flight + one coalesced rerun, `stop()` must not await the in-flight tick (self-deadlock), `drain()` must await before clearing, and no restart on `fresh_enough` / transient / `reauth_required`. The self-deadlock invariant still matters because the **fallback** path still deregisters.

## Testing

- 12241 pass, 0 fail. `lint` 0 errors, `format:check` clean, `typecheck` clean.
- 30 new tests across `manager.test.ts`, `reloadable.test.ts` (new), `credential-rotation-apply.test.ts` (new).
- Guards verified by mutation: removing the rotation guard fails 2 new tests; inverting it fails 5 (3 pre-existing); removing the TOCTOU recheck fails its test.

## Review order

9 atomic commits, each independently green. Read them in order — the first three are pure plumbing, the behavior change starts at `Apply a named credential rotation to just that adapter`.


devxoul:
Good catch — this was the common case, not an edge case. An expired token is precisely what knocks an adapter into the `failed` map, so the renewal that follows would almost always have hit the fallback it was written to avoid.

Auditing the named-adapter paths turned up **three** that returned no `credentialApply`, all reading as "not applied" downstream:

| path | before | now |
|---|---|---|
| down adapter starts on the new credential | `'started'`, no outcome -> **container restart** | `restarted` |
| down adapter still won't start | `null` -> **container restart** | `recovery-pending` |
| retry already queued | `null` -> **container restart** | `recovery-pending` |
| live adapter already on the new credential | `null` -> **container restart** | `already-current` |

That last row was outside what you flagged but is the same defect — silence from a named adapter reads as failure — so I fixed it in the same pass rather than leaving a second round. It needed a new outcome value: `already-current` isn't a restart outcome, so `CredentialApplyOutcome` is now `AdapterRestartOutcome | 'already-current'`.

The fallback rule is now stated positively in `describeUnapplied()`: **only `stop-failed` falls back**, because that is the only outcome where the adapter is still live on the superseded credential. Everything else means the rotation landed or is landing.

Ordinary reload behavior is unchanged — every new outcome is gated on the adapter being the named one. Covered by `an unnamed reload still reports nothing for a down adapter it revived`.

Added the end-to-end test you asked for: `does not restart the container when renewal revives an adapter that was down` starts with the adapter failed in the map, runs the **real** manager and reloadable, and asserts `restartContainer` is never called. Verified by mutation — dropping the outcome fails that test plus the manager-level one.

Also corrected the commit count in the description: 9, not 8. Miscounted after splitting the plumbing.

Folded into `Apply a named credential rotation to just that adapter` and `Add a host-side credential rotation applier`. 12241 pass, lint/format/typecheck clean. Leaving the thread for you to resolve.

PR #1328, opened 2026-08-03
Add mediated reads to Webex channel adapters

devxoul:
## Background

The sandbox refuses model-driven `agent-webex`/`agent-webexbot` CLI use outright — those bins sit in `AGENT_MESSENGER_BINS` (`src/sandbox/privileged-runtime.ts`) precisely because the credentials belong to hostd, not the model. That refusal was correct, but it left a dead end: the mediated alternative, `channel_read`, only registered a history callback on both Webex adapters. A model that needed to discover a space id or fetch one message by id had no path to either — the CLI it would reach for is refused, and the tool that's supposed to replace it didn't cover the same ground.

This gap is sharper for cron sessions specifically. `buildChannelTools` (`src/agent/index.ts:698-706`) always exposes `channel_send` and `channel_read` because they take explicit addressing, but `channel_history` is origin-bound and only appears when the session origin is a channel — a cron session never gets it. So for cron, `channel_read` isn't one option among several; it's the only door into Webex history at all, and it needs to carry the same coverage the origin-bound tool has.

## Summary

Register `list` and `message-get` callbacks on both Webex adapters (`webex.ts`, `webex-bot.ts`), backed by the upstream client's `listSpaces`/`getMessage`, so `channel_read` becomes the full mediated replacement for model-driven `agent-webex`/`agent-webexbot` CLI use rather than a history-only subset of it.

Both adapters follow the same shape already used for `history`:
- `createWebexListCallback` maps `listSpaces({ max: clampLimit(args.limit, 100) })` results to `ChannelListEntry` (space id as `chat`, title-or-id as `name`, `direct` → `dm` else `group`, `isMember: true`).
- `createWebexMessageGetCallback` wraps `getMessage(messageId)` through the same `mapWebexHistoryMessage` mapper history already uses, so a single-message fetch renders identically to a history entry.
- Failures are logged and rethrown unchanged (`describeError`), preserving the existing adapter-error semantics instead of masking upstream errors as "unsupported."
- Registration/unregistration is added symmetrically at `start()`, the failed-startup cleanup path, and `stop()` — the same three sites the `history` callback is already wired into — so there's no lifecycle asymmetry between the two new callbacks and the existing one.

`ADAPTER_READ_CAPABILITIES` in `src/channels/schema.ts` now lists `webex`/`webex-bot` as `['history', 'message-get', 'list']`, matching `slack-bot`/`discord-bot`. Webex's `listSpaces` has no pagination cursor upstream, so the list callback is capped at a single 100-entry page — documented rather than silently truncated.

## What this does NOT do

- Does not add pagination for space listing; the upstream Spaces API wrapper has no cursor, so a single capped page is the ceiling for both adapters.
- Does not touch the sandbox refusal itself — `agent-webex`/`agent-webexbot` remain blocked from model-driven bash; this PR only closes the mediated-path gap that made the refusal a dead end.
- Does not change `channel_history` or cron's origin-bound tool set; `channel_read` was already the only cron-reachable read path and stays that way here.

## Review notes

- `src/channels/adapters/webex.ts` / `webex-bot.ts`: the load-bearing invariant is the three-site registration symmetry (`start`, failed-startup cleanup, `stop`) — worth checking that `listCallback`/`messageGetCallback` appear in all three, matching `historyCallback`.
- `mapWebexSpace` reuses the same `dm`/`group` classification logic as the rest of the adapter; no new space-kind heuristic was introduced.
- Docs updates (`webex.mdx`, `webex-bot.mdx`, `channel-adapters.mdx`, `slack-bot.mdx`, `discord-bot.mdx`) correct the capability table's "only two adapters back all three modes" claim, which stopped being true, and note the single-page space-list limit so operators don't expect cursor-based pagination.

## Testing

- `bun test --parallel src/channels` — 2933 pass, 0 fail, 5852 expects
- `bun run typecheck` — passed
- `bun run lint` — passed, 0 errors, 69 pre-existing warnings
- `bun run format` — passed, 1404 files processed
- `bun run test` (full suite) — 12248 pass, 31 skip, 0 fail, 37153 expects


devxoul:
Closing: this is deployment-specific, not systemic.

Webex `list`/`message-get` is exactly what one reporting agent needed to resolve spaces and fetch message bodies. Adding it here implements a single deployment's unblock inside typeclaw and presents it as a class fix.

The 0.48.0 agent-messenger CLI refusal affects all 18 bundled platforms uniformly. Per-adapter capability work closes it one platform at a time and leaves the mechanism untouched. The systemic defects are tracked separately: the credential mask hole, and the refusal advertising a mediated path that does not exist for 10 of 12 adapters.

Branch `feat/webex-mediated-reads` is retained if the capability is wanted on its own merits.

devxoul:
Reopening. The stated reason for closing was wrong — retracted.

devxoul:
Closing for the correct reason this time.

The reported failure is in the agent-messenger CLI path (`agent-webex`, credentials at `workspace/.config/agent-messenger/`). The affected agent has no Webex channel adapter configured — `discord-bot` is its only entry in both `typeclaw.json#channels` and `secrets.json#channels`. This PR changes a subsystem with no connection to the failure.

The `webex-bot: list-not-supported` router logs that partly motivated it were the model guessing at an unconfigured adapter after the 0.48.0 refusal told it to "use channel_*" — not evidence of a configured adapter.

More generally, channel adapters are not a migration target for CLI users: the two use disjoint credential stores, and an operator using the CLI has by definition not configured the adapter. Capability coverage was never the blocker.

PR #1338, opened 2026-08-05
Render the live runtime paths in the system prompt

devxoul:
## Background

An agent in production memorized the absolute path `/root/.config/agent-messenger/MEMORY.md`. That was **correct** when it learned it. Releases 0.46.2/0.47.0 moved the runtime `HOME` to `/home/agent` and the agent-messenger config to `workspace/.config/agent-messenger`, and the memorized value silently rotted.

The memory shard was last reinforced five days *before* that move, with 17 citations. It was never re-observed afterwards, because the job that would have exercised it early-exits on most runs. So a belief that was true 17 times decayed into a high-confidence falsehood — and the strength model reads those citations as authority, not staleness.

The consequence was not academic. The agent produced a wrong self-diagnosis from that shard (blaming credential masking in a subprocess that is not masked) and posted a degraded message to a user-facing channel.

The structural gap: **the model has no authoritative, always-fresh source for these paths**, so it memorizes resolved absolute values, which rot on every infra change. The system prompt is recomposed every turn, so putting the live values there deterministically outranks a stale shard.

## Summary

One new `## Runtime paths` block, ~64 tokens, three values:

```
## Runtime paths

Current for this container boot; prefer these over remembered paths:
- Agent folder: `/agent`
- Runtime process `HOME`: `/home/agent` (tool sandboxes may differ)
- `AGENT_MESSENGER_CONFIG_DIR`: `/agent/workspace/.config/agent-messenger`
```

**Why live values and not the container constants.** A prompt that states a path with authority has to be right. An operator can override `AGENT_MESSENGER_CONFIG_DIR`, and startup wiring can regress. Rendering `CONTAINER_RUNTIME_HOME` would mean confidently printing the *intended* path rather than the real one — which is precisely the failure mode this PR exists to prevent. So the block reports `os.homedir()` and the live env var, with relative values normalized against cwd and a documented fallback.

**Why branding-independent.** `renderRuntimeBlock` is gated on `branding: true`. Gating paths the same way would deny them to exactly the agents that still need them for real work. A filesystem path is an operational fact, not agent identity, and the existing nondisclosure rule already tells the model that config filenames and CLI names are implementation details to use but not surface as identity. So this is a separate block, emitted in both modes.

**Why slim/cron and subagent too.** The confirmed failure happened on a **cron** turn, which takes the slim path. Excluding slim to save ~64 tokens would preserve the exact bug class this fixes. Cron stays at ~771 tokens, subagent ~340 — both well inside the existing guards.

**Placement.** These values are boot-stable, so the block sits in the cacheable prefix — after the branding-dependent runtime/disclosure block, before session origin. `AGENTS.md` and `scripts/dump-system-prompt.test.ts` are updated together, since the contract only holds if both move.

## What this does NOT do

- Does not add per-tool config dirs (OPENSOMA, GWS, …), credential file paths, host paths, or an env dump. The bar for a permanent per-turn slot: runtime-owned, boot-stable, routinely needed, and not reliably inferable from inside a tool sandbox. Exactly three values clear it. Tool-specific paths belong in that tool's skill.
- Does not restate the `workspace/` / `public/` write-zone guidance — the prompt already says it.
- Does not touch the trailing `## Now` pinning or reorder any other section.
- Does not attempt to invalidate stale memory shards. That is a much larger problem (classifying arbitrary prose as environment-dependent) and this block sidesteps it by making the authoritative value always present.

## Review notes

The load-bearing decision is **live resolution over constants** (`resolveRuntimePaths` in `src/agent/index.ts`). If a future change swaps in `CONTAINER_RUNTIME_HOME` "for consistency", the block starts asserting a path that may not exist in the running process, and a wrong authoritative path is worse than none.

Second: the block is branding-independent *on purpose*. It is not an oversight that it sits outside the `branding` conditional.

Worth verifying: tests cover the env override, the empty-env fallback, relative-path normalization, both branding modes, and presence in cron + subagent prompts — so it proves "live effective value" rather than just today's canonical deployment.

Design reviewed against the alternatives (a resolver tool the model calls on demand; memory-shard invalidation; fixing one agent's `AGENTS.md`). The prompt block is the only one that is global, deterministic, and does not depend on the model choosing to distrust its own memory first.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -30,7 +30,7 @@ bun run debug:prompt --origin cron         # just one
 bun run debug:prompt --origin channel --no-git-nudge
 ```
 
-`composeSystemPrompt` (`src/agent/index.ts`) is the right entry point if you're adding a new section. The cache-suffix contract (least-volatile first → identity → runtime → origin+role → git → memory → now) is enforced by both the helper and `scripts/dump-system-prompt.test.ts`. Reorder one without the other and CI fails. The trailing `## Now` block is pinned last to keep the cache prefix stable across sessions — don't move it.
+`composeSystemPrompt` (`src/agent/index.ts`) is the right entry point if you're adding a new section. The cache-suffix contract (least-volatile first → identity → runtime → runtime paths → origin+role → git → memory → now) is enforced by both the helper and `scripts/dump-system-prompt.test.ts`. Reorder one without the other and CI fails. The trailing `## Now` block is pinned last to keep the cache prefix stable across sessions — don't move it.
 
 Memory is injected per turn into the user prompt; the system prompt never contains long-term memory; vector memory is always on. The memory plugin's `session.turn.start` hook renders de-duplicated direct shards (under budget) or top-K hybrid-search results (over budget) into `event.retrievalContext.results`, which the four turn-drivers (server TUI, channel router, cron consumer, subagent runner) append to the user text.
 
--- a/scripts/dump-system-prompt.test.ts
+++ b/scripts/dump-system-prompt.test.ts
@@ -19,6 +19,7 @@ describe('dumpSystemPrompt', () => {
       expect(out).toContain('# Identity')
       expect(out).toContain('## Runtime')
       expect(out).toContain('TypeClaw runtime version: 1.2.3-debug.')
+      expect(out).toContain('## Runtime paths')
       expect(out).toContain('## Session origin')
       expect(out).toContain('## Your role in this session')
       expect(out).not.toContain('# Memory')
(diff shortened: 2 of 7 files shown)


devxoul:
Closing — wrong layer.

The system prompt carries identity, role, and invariants: things the model cannot derive and must never get wrong. A filesystem path is none of those. It is discoverable from the environment, it is needed by a small fraction of turns, and putting it here charges every turn of every session for it. The slim base, the token guards, and the cache-suffix contract all exist because this space is scarce, and a list of paths does not clear that bar.

The evidence I used to justify it argues for a different fix: the agent burned memory shards rediscovering a path because the `agent-webex` skill hardcodes `~/.config/agent-messenger/MEMORY.md` instead of referencing `AGENT_MESSENGER_CONFIG_DIR`. That belongs in the skill, which loads lazily and costs nothing on unrelated turns.

The mask-is-not-breakage message I wanted to add already exists in the guard's denial text. The reason it was never seen is that EACCES short-circuited into a generic internal-guard error — fixed in #1335. That delivers the right information at the moment it is needed, which is strictly better than saying it earlier and permanently.

PR #1361, opened 2026-08-12
Reduce reviewer subagent latency

devxoul:
## Background

The reviewer subagent flow had three sources of avoidable latency. The `typeclaw-channel-github` skill read the full PR diff and file list in the parent turn before ever spawning the reviewer, duplicating work the reviewer subagent was about to redo itself. Scratch checkouts for broad-context review had no canonical acquisition path, so skills mixed ad-hoc `git clone`/`fetch` with the existing `github_prepare_review_checkout` tool and re-derived `/tmp` paths instead of trusting the returned receipt. Re-review passes re-explored the whole PR from scratch instead of reusing the prior review's findings and narrowing to the intervening diff. Each of these adds a serial round trip before the formal review lands.

## Summary

Three independent latency cuts, one per commit:

- **Delegate GitHub reviews before parent analysis.** `typeclaw-channel-github`'s SKILL.md now does minimal intake only — repo, PR number, head SHA, authorship, prior blocking context — then spawns the reviewer immediately, with no `gh pr diff` or file reads first. After the reviewer returns, parent verification is scoped to the cited `path:line` anchors instead of a second full pass over the diff.
- **Reuse reviewer checkout and rereview context.** `github_prepare_review_checkout` is now the single acquisition path across `code-review`, `doc-review`, and `plan-review`: call it exactly once, then use the returned receipt path exactly — never guess or reconstruct a `/tmp` path, never call it twice. Re-review instructions in `code-review.ts` now diff the prior reviewed head against the current head and reuse prior findings for cumulative correctness, rather than re-exploring the full PR.
- **Batch reviewer exploration and completion waits.** The reviewer's `bash` tool is restricted to one read-only invocation at a time (no loops, pipelines, redirects, `&&`, `;`), independent reads are batched into a single turn, and both the reviewer and the channel skill now wait for the background completion `<system-reminder>` instead of polling `subagent_output`.

## What this does NOT do

- Does not change the read-only contract for the reviewer subagent — it still cannot edit files, install dependencies, or mutate remote state, and the scratch checkout remains a private read cache, never the reviewed artifact.
- Does not remove or weaken `github_prepare_review_checkout`'s allowlisted-repo/SHA verification — this only tightens callers to invoke it exactly once per review pass.
- Does not change verdict logic, severity calibration, or the formal-review-vs-comment decision rules.

## Review notes

Load-bearing: the "exactly once" / "use the returned receipt path exactly" language added to `reviewer.ts`, `code-review.ts`, `doc-review.ts`, and `plan-review.ts` — a second checkout call or a guessed `/tmp` path is precisely the regression this closes. On the channel side, step 1 (minimal intake) and step 4 (proportional anchor validation) in `typeclaw-channel-github/SKILL.md` are the two ends of the same cut: less work before delegating, bounded work after.

## Verification

- `bun run lint` — clean
- `bun typecheck` — clean
- `bun run test` — 12,644 pass, 0 fail


devxoul:
@typeey The branch and canonical checkout are still available at the same head. Please retry the review; the prior run reported checkout infrastructure failure and produced no code findings.

devxoul:
@typeey Please make one final bounded retry against the unchanged current head. CI is green and the earlier retries failed at the upstream provider before any findings or verdict; no code feedback remains to address.

devxoul:
Addressed the checkout failure in the updated branch. Reviewer skills now permit one bounded reacquisition only when a read-only operation against the exact runtime receipt fails with `ENOENT` / `No such file or directory`; recovery must use the same repo slug and full head SHA, total checkout calls are capped at two, and generic retries/path reconstruction/bash acquisition remain forbidden. Prompt-contract tests and the full suite pass (12,684 tests, 0 failures on the combined stack). @typeey please re-review the new head.

devxoul:
@typeey review plz

PR #1395, opened 2026-08-27
Fix no-reply reaction cleanup

devxoul:
## Background

A production incident showed a successful `skip_response` leaving a persistent :eyes: reaction on the triggering message — a "seen, intentionally not replying" marker on a turn that, from the operator's perspective, had nothing to acknowledge. The reaction only clears when the same conversation later gets a genuine reply, so on a conversation that never replies again, the mark is permanent.

The marker was armed by `armSilentTurnAck` on three "deliberate silence" reasons: `skip_response`, an explicit `NO_REPLY` (and its leaked-reasoning variant), and a leaked `skip_response(...)` plain-text tool call. All three are turns where the agent chose not to respond — they are not turns that produced output somewhere else. Reusing the same persistent-ack mechanism for "chose silence" and "output landed elsewhere" conflated two different situations under one reaction.

## Summary

`skip_response`, both `NO_REPLY` variants, and a leaked `skip_response(...)` text call no longer arm the persistent :eyes:. Their three `armSilentTurnAck(...)` call sites are removed; each path still logs its reason and drops continuation reactions exactly as before, it just no longer plants a mark.

`SilentAckReason` is narrowed from five values to two: `github_review_output` and `awaiting_background_child`. Those are the only remaining callers of `armSilentTurnAck`, and both are cases where the turn genuinely produced output — a GitHub review, or work handed to a background child — so a persistent "I did something, watch this thread" marker is still the right signal. The eligibility comment and the removal/cleanup comments on `LiveSession` are updated to describe "output landed elsewhere" rather than "deliberate silence" as the thing being acknowledged.

The ordering guard is unchanged: on a typing-less adapter that collapses same-actor same-emoji reactions into one toggle, the transient engage :eyes: removal is still awaited before the persistent add, so a still-in-flight engage-drop can never strip the ack.

## What this does NOT do

- Does not change the GitHub-review or background-child acknowledgement behavior — both still arm and clear the persistent :eyes: exactly as before.
- Does not change the transient engage :eyes: (added on route, dropped at turn end) for any turn kind.
- Does not touch `dropSilentAckReactions` semantics — a genuine reply still retires every outstanding persistent ack in the conversation, now scoped to just the two remaining reasons.

## Review notes

- `armSilentTurnAck`'s eligibility check is unchanged in shape (`github_review_output` always qualifies on the `github` adapter; `awaiting_background_child` requires the triggering message to have explicitly addressed the bot) — only the set of callers shrank.
- Regression coverage: typing-capable and typing-less adapters, the slow-add-races-a-later-reply case, multi-ack cleanup on a single reply, GitHub review output, and the background-child await path all have updated or new assertions in `src/channels/router.test.ts`.

## Verification

- `bun run lint` — 0 errors (pre-existing warnings only)
- `bun run typecheck` — clean
- `bun run test` — 13,276 pass / 31 skip / 0 fail

### DIFF
--- a/src/channels/router.test.ts
+++ b/src/channels/router.test.ts
@@ -3142,11 +3142,10 @@ describe('disengageReactionEmojiFor', () => {
   })
 })
 
-describe('ChannelRouter silent-ack :eyes: on deliberate silence', () => {
+describe('ChannelRouter reaction cleanup on deliberate silence', () => {
   const REACTION_REF: ReactionRef = { adapter: 'discord-bot', value: 'msg-ref' }
 
-  // discord-bot is typing-capable, so route() adds no eager engage :eyes: — every
-  // captured reaction is the silent-ack, keeping these assertions unambiguous.
+  // discord-bot is typing-capable, so route() adds no eager engage :eyes:.
   const setupSilentTurn = async (
     dir: string,
   ): Promise<{
@@ -3166,7 +3165,7 @@ describe('ChannelRouter silent-ack :eyes: on deliberate silence', () => {
     return { router, sessions, captured }
   }
 
-  test('skip_response leaves an :eyes: on the triggering message', async () => {
+  test('skip_response leaves no :eyes: on the triggering message', async () => {
     const dir = await tempDir()
     const { router, sessions, captured } = await setupSilentTurn(dir)
     sessions[0]!.onPrompt = () => {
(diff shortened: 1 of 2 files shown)


devxoul:
Addressed the retained-output ordering issue in the folded commit. `github_review_output` and `awaiting_background_child` are now evaluated before a clean `skip_response` return, with regressions combining `markTurnSkipped` with both paths; ordinary skips remain reaction-free. Full suite: 13,276 passed.

PR #1421, opened 2026-09-02
runtime: close launch, backup, and rehydrate races

geonu:
## Background

Three independent lifecycle races could turn expected transient states into stale control data or failed recovery:

- Container start registered with hostd before a cold image build and model provisioning. Hostd garbage collection could remove that registration while no container existed, leaving the later launch plan with expired control tokens.
- Backup converted porcelain status to a path-only list. An ordinary untracked file disappearing between `git status` and explicit `git add` aborted the whole backup; a blanket existence filter would incorrectly drop tracked deletions.
- A persisted channel mapping can outlive its JSONL transcript. The pre-open cap pass reported missing files as path-bearing warnings even though session open/fallback already self-heals and the router persists the replacement session.

These fixes are not the only conceivable designs. Deferring all hostd registration or adding leases would require a wider lifecycle/protocol change; broad Git staging would admit concurrent work; filesystem preflight checks would retain the same TOCTOU window. This PR uses the narrowest changes that preserve the existing cleanup, explicit-snapshot, and best-effort-rehydrate contracts.

## Summary

- Refresh hostd registration and rebuild the launch plan after build/model work, immediately before `docker run`. Carry source-appropriate deregistration so daemon-owned restarts can roll back in-process without a socket self-RPC. If the refresh is unavailable, preserve the existing degraded launch behavior but remove hostd control from the final argv.
- Parse NUL-delimited porcelain status into tracked/untracked entries, preserve tracked deletions and rename endpoints, and stage only the original snapshot. If the first ordinary add fails and a previously observed untracked path is now proven absent, retry once without that path; never admit paths discovered by the reread.
- Classify nested `ENOENT` causes during channel rehydrate. Missing-file cap races stay silent, other cap/open failures use fixed privacy-safe warnings, and fresh-session creation errors still propagate.
- Add anonymous regressions for build/model completion ordering, token replacement, in-process rollback, managed-path deletion, no-scope-widening backup retry, and rehydrate warning/fallback branches.

## What this does NOT do

- Does not change hostd GC timing or add a lease/heartbeat protocol.
- Does not switch backup to `git add -A`, `git add -u`, or any repository-wide staging command.
- Does not delete channel mappings on a preflight existence check or change poisoned-session self-healing.
- Does not address Docker log archive limits or unrelated runtime warnings.

## Review notes

- Hostd: the final plan contains only replacement tokens; deferred tests exercise both build-first and model-first completion orders. In-process registration and deregistration share the daemon's per-container serialization.
- Backup: reread status is evidence for removing vanished original untracked paths only. Tracked deletions remain stageable even when absent; present `sessions/` and `todo/` paths retain force-add behavior.
- Channels: `SessionManager.open` may itself create a fresh transcript for an absent file; the explicit thrown-`ENOENT` branch is also covered. The router's existing write-after-factory behavior remains responsible for replacing the durable mapping.

### DIFF
--- a/src/bundled-plugins/backup/README.md
+++ b/src/bundled-plugins/backup/README.md
@@ -38,10 +38,11 @@ If a new prompt arrives while the runner is in flight, the runner finishes its c
 
 ## What it commits
 
-The runner stages two categories of dirty paths:
+The runner stages an explicit snapshot from NUL-delimited porcelain status:
 
-- **Tracked or untracked agent paths** (anything `git status --porcelain=v1 --untracked-files=all` reports), **except** paths under `memory/` — those are owned by the memory plugin's dreaming subagent.
-- **Force-added `sessions/`** — gitignored, but force-added so transcripts survive across restarts.
+- **Tracked changes**, including deletions and both rename/copy endpoints, are staged even when an endpoint is no longer present on disk.
+- **Ordinary untracked paths** are staged only while still present. If one disappears between snapshot and `git add`, the runner re-reads status and retries that same snapshot once without the vanished path; paths discovered during that re-read are never added.
+- **`memory/`** remains excluded. Present **`sessions/` and `todo/`** paths remain force-added; tracked deletions under those prefixes stay in the ordinary snapshot. The post-message pass re-stages only `sessions/` paths that appeared while the message was selected.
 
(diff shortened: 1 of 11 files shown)


geonu:
Superseded by #1424, #1425, and #1426 so each independent root cause and CI surface can be reviewed separately. The hostd CI race from this combined PR is fixed in #1426 with deterministic in-process rollback coverage.

PR #1423, opened 2026-09-02
Give each bwrap file mask its own fd

devxoul:
## Background

The builder in `src/sandbox/build.ts` rendered `--ro-bind-data 3 <file>` for every masked file, reusing a single file descriptor. bwrap copies a `--ro-bind-data` operand out of its fd and then closes it (`copy_file_data(op->fd, tmp_fd)` followed by `close(op->fd); op->fd = -1;`), so the second mask read an already-closed fd and the whole invocation aborted:

    bwrap: Can't write data to file /agent/secrets.json: Bad file descriptor

`resolveHiddenPaths` masks four files unconditionally (`.env`, `secrets.json`, `auth.json`, `.typeclaw/incidents.json`), so this wasn't a failure to mask one secret — it took down the entire model bash surface. Every sandboxed bash call failed, down to `echo`. This surfaced in production on two independent agents immediately after they upgraded to 0.48.8, and both posted public "my execution environment is broken" reports instead of getting work done — the exact failure mode AGENTS.md's "Security serves the product" section warns about: a guard that disables a whole tool surface.

## The fd was never actually working

Only the FIRST mask ever consumed the intended fd. Feeding non-empty content into fd 3 across three masks makes this visible:

    # bubblewrap 0.11.0
    $ bwrap ... --ro-bind-data 3 /tmp/a --ro-bind-data 3 /tmp/b --ro-bind-data 3 /tmp/c ... 3</tmp/src
    a=[PAYLOAD] b=[] c=[]

Masks 2 and 3 did not read `/tmp/src`. In bubblewrap <= 0.11.0 the temp file descriptor is block-local and the bind is performed by PATH, so `close(op->fd)` releases fd 3 and the next iteration's `mkstemp` immediately reclaims it. `copy_file_data` then reads the freshly created, still-empty temp file into itself, hits EOF, and reports success. The mask came out empty by coincidence, not by design.

bubblewrap 0.12.0 reworked destination resolution to be fd-based (`setup_op_bind_mount_fd(..., tmp_fd, "tmpfile", dest_fd, op->dest)`), holding `dest_fd` open across the operation. `mkstemp` no longer lands on the freed slot, fd 3 stays closed, and the read fails `EBADF`.

Note this means the masking was never leaky on the old behaviour — the bind target is always a freshly created temp file, so the outcome was only ever "empty file" or "hard failure", never the original bytes.

## Why it broke at 0.48.8

Not a base-OS change. The published base images show the boundary is the sandboxer itself:

| typeclaw | base image OS | bubblewrap | three masks sharing fd 3 |
|---|---|---|---|
| 0.48.6 | Debian 13 (trixie) | 0.11.0 | exits 0 |
| 0.48.7 | Debian 13 (trixie) | 0.11.0 | exits 0 |
| 0.48.8 | Debian 13 (trixie) | 0.12.0 | `Bad file descriptor`, exits 1 |

0.48.7 was already on trixie, so the Bun/Debian base pin is not implicated. The only delta is that `bubblewrap` is installed unpinned in the base image, and the distro package resolved to 0.12.0 when the 0.48.8 base was rebuilt. No typeclaw code change was involved in the trigger — a rebuild alone was sufficient.

## Summary

Allocate one fd per masked file, counting up from `FIRST_MASK_DATA_FD` (3), and emit a matching `/dev/null` redirect for each, so the rendered command ends with e.g. `3</dev/null 4</dev/null 5</dev/null 6</dev/null`. Every mask now reads the descriptor it was actually given.

Upstream documents no reuse semantics for the operand fd (bwrap.xml: "Copy from the file descriptor FD to a file which is bind-mounted read-only on DEST"), and bubblewrap's own demo script plus flatpak's `flatpak_bwrap_add_args_data_fd_dup()` both allocate a distinct fd per op.

This is deliberately not switched to `--ro-bind /dev/null`, which would make the masked path a character device instead of a regular file and change what `stat`/`test -f` reports on it.

The pre-existing mask coverage was argv-shape only, which structurally cannot observe a file descriptor being closed — that's why a total bash outage shipped green. Added:

- two argv assertions pinning distinct fds and one redirect per masked file (verified to fail against the unfixed builder and pass after the fix)
- a bwrap-gated test that actually executes the rendered command with four masked files and asserts it exits 0, emits no `Bad file descriptor`, and leaks none of the masked content — using the existing `skipIf(!bwrapPresent)` pattern from `availability.test.ts` (bwrap ships in the container, not on the macOS dev host)

Also documents the per-mask fd requirement in `docs/content/docs/internals/sandbox.mdx`, which previously described `--ro-bind-data` as the mask primitive without saying where its content comes from or that the operand fd is single-use.

## What this does NOT do

- Does not change masking semantics — every target still resolves to an empty read-only regular file.
- Does not switch to `--ro-bind /dev/null` or any other primitive.
- Does not change which files or directories are masked, or when.
- Does not pin `bubblewrap` in the base image. Worth considering separately: an unpinned sandbox-critical package can change bwrap's behaviour between two typeclaw patch releases with no code change, which is exactly what happened here.

## Review notes

Verify `maskDataFd(index)` in `src/sandbox/build.ts` and the corresponding `maskRedirects` list stay in lockstep — the fd assigned to the Nth `--ro-bind-data` op must exactly match the Nth `<fd></dev/null` redirect appended to the command string, since a mismatch reproduces the same closed-fd failure at a different index.

Worth considering for a patch release soon: every agent on 0.48.8 currently has no working bash.

## Verification

- New argv tests fail against the unfixed builder and pass against the fixed one (red/green confirmed).
- `bun run test` — 13,364 pass, 32 skip, 0 fail.
- `bun run typecheck`, `bun run lint` (0 errors), `bun run format:check` — all clean.
- Version boundary established by running the same three-mask invocation against the published `typeclaw-base` images for 0.48.6, 0.48.7 and 0.48.8 (table above).
- End-to-end inside the real 0.48.8 agent image (bubblewrap 0.12.0): the fixed builder produces `--ro-bind-data 3/4/5/6` with matching redirects, exits 0 with empty stderr, and all four planted canary files read back empty. Replaying the old shared-fd argv in the same container reproduces `Can't write data to file .../secrets.json: Bad file descriptor` exactly.


geonu:
@devxoul please release...

devxoul:
@typeey try review again

PR #1440, opened 2026-09-03
Restore production parity in the mask verifier

devxoul:
## Background

This corrects a mistake in already-merged work, not a new hardening pass. #1437, #1438, and #1439 (all merged) progressively loosened the release mask-verifier's container while chasing a green run: first `--security-opt apparmor=unconfined`, then `--cap-add NET_ADMIN`, then `--privileged`. Review on #1437 said exactly this:

> The mask assertions themselves are unchanged, but this makes the release verifier less representative of the runtime it is meant to certify.

That review was correct, requested changes, and was overridden instead of addressed — #1437 merged with `reviewDecision: CHANGES_REQUESTED` still standing, and #1439 merged with a failing Windows check.

It mattered more than "less representative": agent containers get `--security-opt seccomp=unconfined` and nothing else (`runArgs` in `src/container/start.ts`). A verifier running MORE privileged than production can pass on a bwrap invocation that would fail inside a real agent container — the exact failure mode this gate exists to catch. So the end state didn't just drift from parity, it silently weakened what the gate certifies while its own inline comments still claimed otherwise.

## Summary

**Container (`scripts/verify-mask-fd-sandbox.sh`):** restore `run_args=(--rm --pull=always --security-opt seccomp=unconfined)`. This is byte-identical to the line as it stood before #1437 (`git show 1fe0ac89:scripts/verify-mask-fd-sandbox.sh` — the commit before the loosening chain started). The comment now states plainly that these flags are production parity and must not be widened to make a runner go green.

**Runner (`.github/workflows/release.yml`):** the three-PR escalation was chasing symptoms of one root cause. Ubuntu 24.04 sets `kernel.apparmor_restrict_unprivileged_userns=1`, so bwrap cannot open the unprivileged user namespace its sandbox is built from — and without that namespace it holds none of the capabilities setup needs. That surfaces first as the opening `mount(NULL, "/", MS_SLAVE|MS_REC)` being denied ("Failed to make / slave: Permission denied"), then, once AppArmor was waived, as loopback bring-up being denied ("RTM_NEWADDR: Operation not permitted") — which is also why `--cap-add NET_ADMIN` never helped: the restriction is on the namespace that would confer the capability, not on the capability itself. A new step clears that sysctl on the runner before the verify step runs, so the container under test stays identical to production instead of the check adapting to it.

Verified locally against the published `ghcr.io/typeclaw/typeclaw-base:0.48.10` with the production-parity flags restored: `MASK_CONTRACT_OK: 4 masks over 4 distinct fds, all empty`. The shipped image's mask contract was fine the whole time — only the verifier's own environment was ever wrong.

## What this does NOT do

- No product or runtime behavior change. 0.48.10 is already published and unaffected — the loosened flags only ever applied to the CI verifier's own throwaway container, never to shipped artifacts.
- Does not touch the mask/fd assertions themselves (`MASK_BWRAP_FAILED`, `MASK_LEAKED`, `MASK_CONTRACT_OK`) or `buildArgv()`/`appendMasks()`.

## Review notes

The sysctl step in `release.yml` is **unverified on a real runner** — it can't be exercised from a dev machine, and the release workflow is the only place it runs. If it doesn't fully clear the restriction, the verify step will still fail, just at the same loopback point as before. Watch this on the next release run.

## Verification

- `bun run test` — 13468 pass / 0 fail
- `bun run typecheck` — clean
- `bun run lint` — 0 errors
- `bun run format` — clean

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -325,18 +325,18 @@ jobs:
           docker pull --platform linux/arm64 "$IMAGE"
           docker buildx imagetools inspect "${REGISTRY_IMAGE}:latest"
 
-      # Ubuntu 24.04 ships kernel.apparmor_restrict_unprivileged_userns=1, so
-      # bwrap cannot open the user namespace it builds the sandbox from and ends
-      # up without CAP_NET_ADMIN for the loopback bring-up in its new netns
-      # ("RTM_NEWADDR: Operation not permitted"). Clearing the sysctl is what
-      # fixes that, and it is a runner setting rather than a change to the
-      # container under test.
+      # Only the DIAGNOSTIC lane needs this. Ubuntu 24.04 ships
+      # kernel.apparmor_restrict_unprivileged_userns=1, and a process that has
+      # been made AppArmor-unconfined is subject to that restriction, so the
+      # diagnostic container cannot open the user namespace bwrap builds the
+      # sandbox from ("setting up uid map: Permission denied"). Measured on this
+      # runner; it is unrelated to --share-net, which the verifier now also sets.
       #
-      # It is only half the story: measured on this runner, the opening mount is
-      # denied by docker-default AppArmor regardless of this sysctl, which is
-      # why the verifier also passes apparmor=unconfined. See the flag comment
-      # in scripts/verify-mask-fd-sandbox.sh for that divergence and its cost.
-      - name: Allow unprivileged user namespaces for bwrap
+      # It does nothing for the production lane: under docker-default AppArmor
+      # the userns is permitted and the opening mount is denied instead, so that
+      # lane fails at mount regardless of this sysctl. Do not read a green
+      # diagnostic lane as evidence the production lane can pass.
+      - name: Allow unprivileged user namespaces for the diagnostic lane
         run: sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0
 
       - name: Verify the sandbox file-mask contract against the shipped bwrap
@@ -358,10 +358,27 @@ jobs:
         # pulled last (arm64). Running that unqualified here would exec a
         # foreign-arch binary with no QEMU registered and fail the release on
         # exec format instead of on bwrap.
+        #
+        # Two lanes, both required, because either one alone misleads. The
+        # diagnostic lane lifts AppArmor so the mask assertions stay reachable
+        # even where host policy forbids bwrap; it answers "did the mask argv
+        # regress?". The production lane uses the flags an agent container
+        # actually gets and answers "can bwrap run at all here?". Running the
+        # diagnostic first means a production-lane failure is already
+        # attributable: masks OK + production dead == host policy, not the argv.
         env:
           VERSION: ${{ needs.checks.outputs.version }}
           PLATFORM: linux/${{ runner.arch == 'ARM64' && 'arm64' || 'amd64' }}
-        run: scripts/verify-mask-fd-sandbox.sh "${REGISTRY_IMAGE}:${VERSION}" "${PLATFORM}"
+        run: scripts/verify-mask-fd-sandbox.sh "${REGISTRY_IMAGE}:${VERSION}" "${PLATFORM}" diagnostic
+
+      # Authoritative. A failure here is NOT a CI misconfiguration to be flagged
+      # away — it means model bash is dead under the flags agent containers run
+      # with, which is the outage this whole gate exists to prevent shipping.
+      - name: Verify bwrap runs under production container flags
+        env:
+          VERSION: ${{ needs.checks.outputs.version }}
+          PLATFORM: linux/${{ runner.arch == 'ARM64' && 'arm64' || 'amd64' }}
+        run: scripts/verify-mask-fd-sandbox.sh "${REGISTRY_IMAGE}:${VERSION}" "${PLATFORM}" production
 
   # ─────────────────────────── release ───────────────────────────
   # The irreversible publish path, gated on a verified multi-arch base image.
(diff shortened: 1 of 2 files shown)


devxoul:
You were right, and the sysctl alone is not sufficient. I measured it on an `ubuntu-24.04` runner against the published `0.48.10` base image rather than reasoning about it, and pushed a follow-up commit.

| # | Container flags | Runner sysctl | Result |
|---|---|---|---|
| A | `seccomp=unconfined` | untouched (`=1`) | ❌ `Failed to make / slave: Permission denied` |
| B | `seccomp=unconfined` | `=0` | ❌ **same error** — the fix as originally proposed here |
| C | `seccomp=unconfined` + `apparmor=unconfined` | `=0` | ✅ `bwrap --unshare-all` completes |
| D | `seccomp=unconfined` (control) | `=0` | ❌ same error |

`docker info` reports `["name=apparmor","name=seccomp,profile=builtin","name=cgroupns"]`, so docker-default AppArmor is active on the runner.

So the two original failures had **two** causes, not one:

- the opening `mount(NULL, "/", MS_SLAVE|MS_REC)` is denied by docker-default AppArmor, and no runner-side sysctl affects it;
- the loopback `RTM_NEWADDR` is denied because `kernel.apparmor_restrict_unprivileged_userns=1` blocks the user namespace that would carry `CAP_NET_ADMIN`.

Both are needed. Note C passes *without* `--cap-add NET_ADMIN`, so that was never the right lever, and neither was `--privileged`.

`apparmor=unconfined` is now restored as a documented, minimal divergence — and I've stated its cost at the code site rather than papering over it: production does **not** pass `apparmor=unconfined` (`runArgs`, `src/container/start.ts`), so on an AppArmor-enabled Docker host the per-tool sandbox plausibly hits this same mount denial, and this gate can no longer catch that. That is a question about `src/container/start.ts` and wants its own change rather than a flag pick here.

devxoul:
Follow-up measurement. My previous table was itself not production-parity — it omitted `--share-net` (production sets `network: 'inherit'`, so `src/sandbox/build.ts:79,90-93` emits `--unshare-all --share-net`), ran as root, and kept full capabilities. Re-measured with the real production argv and posture on `ubuntu-24.04`, sysctl left at its **default** (`=1`):

| # | argv | posture | flags | Result |
|---|---|---|---|---|
| E | prod (`--share-net`) | non-root, `cap-drop ALL` | `seccomp=unconfined` | ❌ `Failed to make / slave: Permission denied` |
| F | prod (`--share-net`) | non-root, `cap-drop ALL` | `+ apparmor=unconfined` | ❌ `setting up uid map: Permission denied` |
| G | gate (no `--share-net`) | non-root, `cap-drop ALL` | `+ apparmor=unconfined` | ❌ `loopback: Failed RTM_NEWADDR` |
| H | prod (`--share-net`) | root, full caps | `+ apparmor=unconfined` | ❌ `setting up uid map: Permission denied` |

Only `sysctl=0` **and** `apparmor=unconfined` together produce a pass.

Two conclusions:

1. **`--share-net` does not remove the sysctl requirement.** Case F has it and still fails, at the user-namespace `uid_map` write rather than at loopback. So these are genuinely two independent blockers: docker-default AppArmor denies the opening mount, and `kernel.apparmor_restrict_unprivileged_userns=1` denies the user namespace itself.

2. **This is a product problem, not a CI one.** Stock Ubuntu 24.04 has both conditions by default, and case E is the exact production shape. Nothing in the runtime degrades around it — `applyBashSandbox` only checks `bwrap --version` (`src/sandbox/availability.ts:16-24`), the `proc-bind` probe treats a bwrap startup failure as `inconclusive` (`availability.ts:369-380`), and the `tmpfs` fallback still opens with `bwrap --unshare-all` (`build.ts:77-80`). On such a host every model bash call fails, down to `echo`.

The second blocker is a **host** sysctl, which the runtime cannot set from inside its own container. So "make the verifier match production" cannot be satisfied on GitHub's Ubuntu runners until the product decides what production does on an AppArmor-enabled host.

Holding this PR pending that decision rather than picking flags to force it green.

devxoul:
Addressed. You were right twice, and the second time pointed at something bigger than this PR.

**What changed:** the gate no longer asks one lane two questions. It now runs both.

- **diagnostic lane** — keeps `apparmor=unconfined`, answers *"did the `--ro-bind-data` mask/fd argv regress?"* (the 0.48.8 class of bug this check was built for). Stays reachable on a host whose policy forbids bwrap outright. **Cannot authorize a release on its own.**
- **production lane** — uses the flags agent containers actually get (`src/container/start.ts:888-899`). Answers *"can bwrap run at all here?"*. **Authoritative.**

Running the diagnostic first makes a production-lane failure attributable: masks OK + production dead means host policy, not the argv.

**The production lane fails on `ubuntu-24.04` today, and that is the correct outcome.** It is reporting a real defect, now filed as #1441: on a stock Ubuntu Docker host the per-tool sandbox is dead and nothing detects it — `availability.ts:16-24` only checks `bwrap --version`, the proc probe downgrades a startup failure to `inconclusive` (`availability.ts:369-380`), and the `tmpfs` fallback still opens with `bwrap --unshare-all` (`build.ts:77-80`). So this PR does not make the release green. It makes the gate honest, and the fix belongs in `src/container/start.ts`.

**Also corrected — the argv was never the production shape.** Production sets `network: 'inherit'`, so `build.ts` emits `--share-net`; the verifier omitted it, made bwrap bring up a loopback production never asks for, and failed `RTM_NEWADDR` in a way production cannot reach. Added.

**Sysctl rescoped to the diagnostic lane**, which is the only lane it helps. The measurement is counterintuitive and worth stating: docker-default permits the userns and denies the mount, whereas an AppArmor-*unconfined* process is subject to Ubuntu's userns restriction and is denied the namespace itself. So `apparmor=unconfined` alone trades one failure for an earlier one.

**Still unmatched and deliberately not faked:** production drops to the host UID and clears capabilities in the entrypoint shim before the runtime starts. The verifier does not, and the code says so rather than claiming full parity.

